PR #233, opened 2019-10-02
get jest working in client sub-project

AlexCLeduc:
I thought I'd get the ball rolling on this 

I'm far from done, I want to get a few different examples going so we have a proof-of-concept/template for different use cases.

- [ ] write at least one test with DOM interactions 
- [ ] write at least one react "shallow render" snapshot test 
- [ ] write at least one test that has "mock subjects", or populates the subject and table stores with dummy data
- [ ] write at least one test that tests a panel
- [ ] write at least one test that mocks out the graphQL API
- [x] find out how to test with real text coming from our templated yaml (tests won't have our custom `yaml-lang-loader`)
- [ ] figure out if there's a way we can avoid using a `.babelrc` file for jest, because it's confusing that webpack won't use it. I remember at some point webpack would even behave differently based on whether `.babelrc` was present

Also, I think we have to keep using babel in order to use ES modules (despite me testing w/ node 12 which supports modules) AND we have to go back to using `require()` because we're not using imports the way we should be. Apparently import statements are supposed to get hoisted and run before any other statements.

In other words, this snippet works in "webpack ES modules" but not spec-compliant node 12. But it works fine if you replace `import` with `require`

```javascript
import _ from 'lodash'
window._ = _ 
import 'script_that_uses_lodash_using_global_reference' //this script gets run before the above assignment  
```


AlexCLeduc:
to be continued via #936

Stephen-ONeil:
@AlexCLeduc with #936 open, is this branch good to delete or are you still referring to it?

PR #253, opened 2019-10-11
6 new financial models/schemas in db

ktw1016:


server/src/models/load_utils.js
```diff
@@ -42,6 +42,8 @@ const bilingual_remap = (obj, dest_key, source_key) => ({
   [`${dest_key}_fr`]: obj ? obj[`${source_key}_fr`] : null,
 });
 
+const check_ngtive_nan = (value) => _.isNaN(+value) ? null : +value;
```
Stephen-ONeil: Can you explain this one to me? I don't think what it's doing lines up with its name, but I'm also curious why it's needed for `orgTransferPayments` and `orgVoteStatPa` and not for any other tables. Is there something weird with the data in those two csvs?

server/src/models/finances/models.js
```diff
@@ -2,13 +2,88 @@ import _ from 'lodash';
...
+  const ProgramSobjsSchema = mongoose.Schema({
+    program_id: parent_fkey_type(),
+    so_num: number_type,
+    pa_last_year_3: number_type,
+    pa_last_year_2: number_type,
+    pa_last_year: number_type,
+  });
+  const ProgramVoteStatSchema = mongoose.Schema({
+    program_id: parent_fkey_type(),
+    pa_last_year_3: number_type,
```
Stephen-ONeil: This model should have a `vs_type` field too, please add it here/in the schema/in the snapshot query.

server/src/models/finances/schema.js
```diff
@@ -1,9 +1,96 @@
...
+  
   extend type Program{
+    program_sobjs: [ProgramSobjs]
+    program_vote_stat: [ProgramVoteStat]
     program_spending: [ProgramSpending]
     program_fte: [ProgramFte]
   }
+
+  type OrgVoteStatPa{
+    dept_code: String
```
Stephen-ONeil: Minor nitpick, but `dept_code` here's redundant since the `OrgVoteStatPa` type can only be reached through an `Org`, which already knows its own `dept_code`. Same with the `program_id` on `ProgramVoteStat` for example.

server/src/models/load_utils.js
ktw1016: From what I recall, this was because of negative values and NaNs. populate db won't cast value to number because of '-'.. Couldn't replicate the NaN issue though

ValidationError: OrgTransferPayments validation failed: pa_last_year_5_auth: Cast to Number failed for value "-" at path "pa_last_year_5_auth"

server/src/models/finances/models.js
ktw1016: Oops.. nice catch..!! thank you

server/src/models/load_utils.js
Stephen-ONeil: Ah, ok. It's a bit confusing because the function name here is the "why", when we usually name things after the "what".

Looking around, something like this is done in other populate scripts that are handling numbers too (well, at least line 54 of `result/populate.js`). We probably want something like this in `load_utils.js`
```
// Populate time validation will choke on either the string representation of a negative number or a value of NaN
// Need to parse strings to numbers ourselves to handle negatives, but take care to preserve nulls which we assume are csv blanks (trying to convert a null gives a NaN)
// Preserving nulls instead of defaulting to nulls because we still want validation errors to surface when any other data issue produces a NaN!
const null_preserving_to_number = (value) => _.isNull(value) ? _.toNumber(value) : null;
```

Could you use that and do a quick look in the other populate scripts for things following a similar pattern? Thanks!

ktw1016: So I think the problems are
1. if value isn't `null` (meaning value has actual numbers), this will always be null
2. It has to use _isNaN because some values are "-". isNull and isNaN will have different output for that.

So I rewrote it using _.toNumber but still not sure if this is sufficient. Please have a look..!!

server/src/models/results/populate.js
```diff
@@ -48,10 +50,10 @@ export default async function({models}){
       doc,
       stable_id,
     } = indicator;
-    
+
     indicator.indicator_id = indicator.id;
     indicator.id = null;
-    indicator.target_year = _.isNaN(parseInt(target_year)) ? null : parseInt(target_year);
+    indicator.target_year = null_preserving_to_number( _.parseInt(target_year) );
     indicator.target_month = _.isEmpty(target_month) ? null : +target_month;
```
ktw1016: `_.isEmpty` and `null_preserving_to_number` is different (I think mainly for `0` values) so I left it as is

server/src/models/load_utils.js
Stephen-ONeil: My bad on two counts
  1. was missing a `!` before the `_.isNull` in the example code above, just a typo
  2. ahh, wait, I thought you were saying that any string containing a '-' was failing to validate as a number with mongo/mongoose, e.g. negatives (maybe I was confused by the original function name mentioning negatives). I hadn't realized that there are some cells in the csvs that contain only a '-'. 

Those cells that contain only a '-' look to me like data errors, in which case my version (without the typo) does what I wanted by surfacing errors instead of suppressing them! We want to know when a non-number is in a numeric column of a csv because that's a sign something's broken up stream.

I'll talk to the pipeline folks about those -'s. The proper fix here is probably to get cleaner data, not to code around the issues.

Stephen-ONeil: Yup, those '-' cells should probably just be empties, Ross will confirm and fix them on the pipeline side.

ktw1016: Okay! so then I'll wait until it's fixed on pipeline side and see if I can remove this entirely, or go with your approach?

Stephen-ONeil: I think my approach, corrected, and replace all the "-" cells in org_transfer_payments.csv with 0's on this branch for now so you can continue testing. We'll just wait for the pipeline's final fix before merging this.

ktw1016: Ok!

Stephen-ONeil: Data updated on master, -'s should all be zeros now.

server/src/models/results/populate.js
Stephen-ONeil: I think they give the same value? What were you seeing that was different?

At this point, `target_month` should be a string or null, so `_.isEmpty` will only be true if the value is null or an empty string `""`.

server/src/models/load_utils.js
```diff
@@ -42,6 +42,11 @@ const bilingual_remap = (obj, dest_key, source_key) => ({
   [`${dest_key}_fr`]: obj ? obj[`${source_key}_fr`] : null,
 });
 
+// Populate time validation will choke on either the string representation of a negative number or a value of NaN
+// Need to parse strings to numbers ourselves to handle negatives, but take care to preserve nulls which we assume are csv blanks (trying to convert a null gives a NaN)
+// Preserving nulls instead of defaulting to nulls because we still want validation errors to surface when any other data issue produces a NaN!
+const null_preserving_to_number = (value) => _.isNaN( _.toNumber(value) ) ? null : _.toNumber(value);
```
Stephen-ONeil: This is the only thing holding me up still.

The comment above it is actually wrong. That was my bad, but I suggested it when I was misunderstanding what was going on with the `-` cells in the csvs and thought the error message you'd quoted was about all negative values. You can delete that whole comment.

`null_preserving_to_number` doesn't preserve nulls right now. `_.toNumber(null) === 0`, for a reason that is not completely clear to me (maybe so that `1 + null === 1`, although I'm not certain that's really desirable). Either way. `null_preserving_to_number(null) === 0`, not `null`.

And it looks like you made a commit removing it from use while I was typing it, so that works too. I was going to suggest either dropping it all together or changing it to
```
// toNumber but leaves nulls null, instead of casting them to 0
const null_preserving_to_number = (value) => _.isNull(value) ? null : _.toNumber(value);
```

Either way you'll have to reupdate the snapshots because a bunch of 0's will become nulls. Just skim the snapshots manually to make sure nothing is wrong with them before you commit the updated ones!

Need something like 
```
const null_preserving_to_number = (value) => ( _.isNull(value) || _.isNaN( _.toNumber(value) ) ) ? null : _.toNumber(value)
```

ktw1016: I personally think it should just be dropped as it's now only being used in results/populate..!

server/src/models/results/populate.js
ktw1016: So as of current `null_preserving_to_number`, it will output `0`,  where as isEmpty gives `null`. which now that I look back, is actually error from `null_preserving_to_number` as it should be `null`. I'll just be removing this function so there's no confusion..!

server/src/models/load_utils.js
Stephen-ONeil: Ah, that sounds good then, go for it!

server/src/models/results/populate.js
```diff
@@ -30,7 +27,7 @@ export default async function({models}){
       "planned_spend_pa_last_year",
       "planned_fte_pa_last_year",
     ], key => {
-      sub_program[key] = null_preserving_to_number(sub_program[key]);
+      sub_program[key] = _.isNaN(sub_program[key]) ? null : +sub_program[key];
```
Stephen-ONeil: Oh, wait, there's a missing conversion to number here, should be `_.isNaN(+sub_program[key])`.

ktw1016: Oh hmm this was the original code but yeah I think you're right

Issue #254, opened 2019-10-11
Dev performance issues with the client build script


Stephen-ONeil:
Recently, lots of MacBook Airs have been chugging and spinning their fans like mad when running any (?) flavour of the client webpack build script.

I've also had a couple near lock-ups on my arch linux Dell, but far less often and potentially for different reasons.

Things that happened recently and might be the culprit, from oldest to most recent:
  - I wrapped a bunch of the dev commands in nodemon to watch files and automatically re-run. Didn't seem to be TOO big a performance hit at the time
  - I enabled the webpack `sideEffect` optimization. I thought it was only on in prod builds, but I should confirm that. This was expected to increase webpack's work load so is a likely culprit
  - I asked everyone to upgrade to node 12. It was only after this that I started hearing about webpak builds causing fans to kick in, but this wasn't too long after the `sideEffect` optimization was enabled so it isn't necessarily node 12's fault

Ran `top` alongside an `npm run IB_q_both` build on both my arch linux and a MacBook air. Both hit around 130-170% CPU usage for most of the build process. On arch, the process quickly went to sleep after completing (staying alive in watch mode). On the MacBook though, the CPU usage actually SPIKED after the process finished, hitting ~200% CPU usage and the STICKING there. Ouch!

Is it something about the MacOS filesystem/the watcher that webpack is using? The nodemon file watchers all idle at ~0% CPU on both arch and MacOS.

PR #255, opened 2019-10-11
CPU-friendlier webpack watch builds

AlexCLeduc:
Makes client watch-build (`npm run IB_`) consume less CPU, see #254.

branch name is misleading, I haven't actually tested whether anything is faster

1. I made webpack's file-watcher ignore `node_modules/`  
    * this resulted in a moderate decrease in CPU usage
2. I added `fsevents` as a dependency because it helped get the CPU usage down. 
    * this resulted in much, much less CPU usage
    * If you do a little googling you'll see this weird trick has fixed many CPU usage issues with webpack's watcher in the past. I'm surprised it hasn't been properly integrated within webpack itself.

Note that I have not tested this with a machine that doesn't have watchman installed, I'm not even sure if watchman is being used under the hood ¯\_(ツ)_/¯ Someone should test a different machine before we merge

### DIFF
--- a/client/build_code/webpack_starter.js
+++ b/client/build_code/webpack_starter.js
@@ -88,6 +88,7 @@ gitsha(function(err, commit_sha){
         //uncomment these lines if watch isn't working properly
         //aggregateTimeout:300, 
         //poll:true
+        ignored: /node_modules/,
       },function(err, stats){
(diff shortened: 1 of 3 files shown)


Stephen-ONeil:
This fixes it on my Mac! I don't have Watchman installed on it for the record.

PR #262, opened 2019-10-15
GitHub link

Stephen-ONeil:
Thinking of adding a GitHub link to either the runtime footer links or to the site version number. If it's the version number, can (should?) I make it link to the deployed commit/deployed range of commits?


Stephen-ONeil:
:clap:

PR #270, opened 2019-10-16
Add lab landing page and text diff tool to route load test config

Stephen-ONeil:
Never updated the route load test config with these pages, would have surfaced that error I introduced last Thursday if we had. Ouch.

### DIFF
--- a/client/browser-tests/route-load-tests-config.js
+++ b/client/browser-tests/route-load-tests-config.js
@@ -190,4 +190,14 @@ module.exports.route_load_tests_config = [
     route: "resource-explorer",
     test_on: ["eng", "basic-eng"],
   },
+  {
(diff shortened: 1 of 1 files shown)


client/browser-tests/route-load-tests-config.js
```diff
@@ -190,4 +190,14 @@ module.exports.route_load_tests_config = [
     route: "resource-explorer",
     test_on: ["eng", "basic-eng"],
   },
+  {
+    name: "Data Lab landing page",
```
cookkate: It's not called the data lab, it's the InfoLab!!!!!11

Stephen-ONeil: `<jk>` Too late, you approved it so I'm merging it as is `</jk>`

PR #274, opened 2019-10-17
Some drr18 prep

Stephen-ONeil:
Marking any drr17 specific code that can be safely deleted come drr18, cutting down on hard coded references to result docs.

Ideally, standard dp and drr updates should only require updating:
  1) the data
  2) `results_docs` in `client/src/models/results.js`
  3) the doc keys in `server/src/models/results/results_common.js`

Could pull back the source of truth fully in to data, but that seems like overkill at the moment.


client/src/models/results.js
```diff
@@ -508,6 +511,10 @@ const get_result_doc_keys = (doc) => _.chain(result_docs)
   })
   .value();
 
+const current_drr_key = _.chain( get_result_doc_keys('drr') ).sortBy().last().value();
+const current_dp_key = _.chain( get_result_doc_keys('dp') ).sortBy().last().value();
```
cookkate: I think it makes more sense for the keys to come from `get_result_doc_keys` sorted.

Stephen-ONeil: Good idea.

Jokes on you though, I can't re-request review because I hadn't even requested one yet! Still a good bit more to tackle on this branch I think lots of annoying hard coded doc names in the results count code (hard coded by me...).

cookkate: wait you didn't? shoot, I got too eager

Stephen-ONeil: Aha

client/src/diff/TextDiff.js
```diff
@@ -100,9 +98,13 @@ const process_indicators = (matched_indicators, indicator_status) => {
...
-      const status = indicator.doc === "dp18" ?
-        "indicator_removed":
-          indicator.doc === "dp19" ? "indicator_added": indicator.doc;
+      const status = indicator.doc === previous_dp_key ?
+        "indicator_removed" :
+        (
+          indicator.doc === current_dp_key ? 
+            "indicator_added" : 
+            indicator.doc
+        );
```
Stephen-ONeil: Hmm, this is a semi common pattern for us, but I don't love it. Better than a `let` where the assignment happens in a following set of if statements though.

I've occasionally used an immediately invoked function to get the same single expression declaration. Does anyone else agree that the following is better than the nested ternary?
```
const status = (() => {
  switch(indicator.doc){
    case previous_dp_key:
      return 'indicator_removed'
    case current_dp_key:
      return 'indicator_added'
    default:
      return indicator.doc;
  }
})();
```

cookkate: From my perspective this results from trying to do things on one line. I'm not sure a switch construct is any better than an if/else.

client/src/diff/TextDiff.js
```diff
@@ -114,31 +116,36 @@ const process_indicators = (matched_indicators, indicator_status) => {
...
     })
     // target_changed is subset of indicator_desc_changed so it must be included
-    .filter(row => indicator_status["indicator_desc_changed"].active && row.status === "target_changed" ?
-            true : indicator_status[row.status].active)
+    .filter(
+      row => indicator_status["indicator_desc_changed"].active && 
+        row.status === "target_changed" ?
+          true :
+          indicator_status[row.status].active
+    )
```
cookkate: This redesign makes the ternary seem separate from the `indicator_status["indicator_desc_changed"].active`, but the test for the ternary is the full `indicator_status["indicator_desc_changed"].active && row.status === "target_changed"`

client/src/gen_expl/result_hierarchies.js
```diff
@@ -154,26 +157,22 @@ export function create_full_results_hierarchy({subject_guid, doc, allow_no_resul
...
+            id: crso.guid,
+            isExpanded: false,
+            data: {
+              subject: crso,
+              type: 'cr',
+              name: crso.name,
+              resources: get_resources(crso),
+            }, 
+          }))
+          .value();
```
cookkate: Does this work the same? Are first wave CRs really CRs or are they SOs?

client/src/panels/result_graphs/result_components.js
```diff
@@ -39,15 +39,18 @@ import { TM, text_maker } from './result_text_provider.js';
...
+    return is_previous ? <span>{indicator_previous_actual_text(indicator)}</span> : <span>{indicator_actual_text(indicator)}</span>;
   }
+  // vv delete on drr17 exit
   if(!is_drr17){
-    return is_previous ? <span>{indicator_previous_target_text(indicator)}</span> : <span>{indicator_target_text(indicator)}</span> ;
+    return is_previous ? <span>{indicator_previous_target_text(indicator)}</span> : <span>{indicator_target_text(indicator)}</span>;
   }
-  return <span>{drr17_indicator_target_text(indicator)}</span> ;
+  return <span>{drr17_indicator_target_text(indicator)}</span>;
+  // ^^ delete on drr17 exit
```
cookkate: `is_drr17` should be deleted but not just by deleting the text between the comments, which we're likely to do without thinking

client/src/diff/TextDiff.js
Stephen-ONeil: The desire is for `consts` over `lets`. `consts` give (some) assurance that the value isn't modified post-declaration. Doing things on one line/in a single expression is just a requirement of `consts`.

client/src/diff/TextDiff.js
Stephen-ONeil: Does
```
.filter(
  row => indicator_status["indicator_desc_changed"].active && 
    row.status === "target_changed" ? true : indicator_status[row.status].active
)
```
read more clearly?

When I want to wrap an expression to multiple lines, I try to break after an operator add a level of indentation after the first wrap. I also prefer to always break ternaries in to three lines, but I see how that scans oddly here.

client/src/gen_expl/result_hierarchies.js
Stephen-ONeil: First wave CRs are true CRs.

client/src/panels/result_graphs/result_components.js
Stephen-ONeil: When I started marking up drr17 specific code, I quickly realized it wouldn't all be a completely straight forward deletion. I kept at it mostly so I kept track of what `drr##` instances I'd dealt with and which I didn't need to bother with.

In all but one or two cases, deleting exactly what I've marked will surface anything complications when the linter complains about undefined/unused variables. I guess I am somewhat relying on me being the one to handle the cleanup as well though. 

Do you think it would be better to take out the deletion flags and leave it as a clean slate for the eventual clean up crew?

cookkate: I dunno. This code in particular won't start throwing errors or linter errors if you just delete that chunk though, it struck me as a particularly dangerous case. At this point it's *probably* stuck enough in both our minds not to be a problem.....hopefully....

Stephen-ONeil: I could change the comments to `delete, carefully, on drr17 exit` :laughing:

client/src/diff/TextDiff.js
Stephen-ONeil: Line seems to be part of a bigger bug, #280, #wontfix in this PR

PR #279, opened 2019-10-18
Fix infographic links in flat results tables

### LINKED ISSUE #275: Results table program links broken

cookkate:
Fixes #275


client/src/panels/result_graphs/result_flat_table.js
```diff
@@ -89,7 +90,7 @@ const subject_link = (subject) => {
   } else {
     return (
       <span>
-        <a href={infograph_href_template(subject,"results")}>
+        <a href={infograph_href_template(subject.data.subject,"results")}>
```
Stephen-ONeil: Oh, ok, so `ind.parent_subject` passed as the `subject` argument isn't a subject, it's a list (or flattened hierarchy?) node containing a subject. If I've got that right, some renaming could help make sense of things.

PR #282, opened 2019-10-18
Featured content refresh

Stephen-ONeil:
Haven't had a big data update in a while, but it was definitely time to remove the new badges from the people and lab links.

Might be a good time to break from "Featured Content" being primarily a list of recent updates. What should we specifically and intentionally feature?


Stephen-ONeil:
IGOC probably

PR #283, opened 2019-10-21
Welcome matt dead programs

### LINKED ISSUE #276: Welcome mat text has a bug for dead programs
See e.g. [#orgs/program/ND-AVI00/infograph/financial](http://localhost:8080/build/InfoBase/index-eng.html#orgs/program/ND-AVI00/infograph/financial)

Screen Shot 2019-10-18 at 9 10 02 AM

Years refers to "1 years ago" but the year is 2013-14. So it looks like two issues with the text itself:
- grammatical (1 years)
- 13-14 isn't 1 year ago

Should check the French as well

ktw1016:
fixes #276

### DIFF
--- a/client/src/core/format.js
+++ b/client/src/core/format.js
@@ -179,7 +179,6 @@ const types_to_format = {
     return`${year_int}-${ window.lang === "en" ? year_int-2000+1 : year_int+1}`;
   },
   "fiscal_year_to_year": (fiscal_year) => _.chain(fiscal_year).split("-").head().value(),
-  "year_plural_text": (year_num) => year_num > 1 ? "years" : "year",
 };
 
 
--- a/client/src/core/format.js
+++ b/client/src/core/format.js
@@ -179,6 +179,7 @@ const types_to_format = {
     return`${year_int}-${ window.lang === "en" ? year_int-2000+1 : year_int+1}`;
   },
   "fiscal_year_to_year": (fiscal_year) => _.chain(fiscal_year).split("-").head().value(),
+  "year_plural_text": (year_num) => year_num > 1 ? "years" : "year",
 };
 
 
--- a/client/src/panels/welcome_mat/welcome_mat.js
+++ b/client/src/panels/welcome_mat/welcome_mat.js
@@ -158,12 +158,18 @@ const WelcomeMat = (props) => {
     latest_hist_spend_data,
     oldest_hist_spend_data,
   } = calcs;
-  const current_year = _.parseInt( _.split( _.last(actual_history_years), '-') ) + 1;
-  const actual_hist_years_apart = current_year - _.parseInt( _.split(oldest_hist_spend_data.year, '-') );
+  const current_year = _.parseInt( run_template('{{pa_in_year_short_first}}') );
+  const oldest_hist_year = _.chain(oldest_hist_spend_data.year)
+    .split('-')
+    .head()
+    .parseInt()
+    .value();
+  const actual_hist_years_apart = current_year - oldest_hist_year;
   const latest_equals_oldest_hist = oldest_hist_spend_data.year === latest_hist_spend_data.year;
 
   //vars used multiple times accross multiple cases
   const years_ago = <TM k="years_ago" args={{
+    plural_years: actual_hist_years_apart > 1,
     actual_hist_years_apart: actual_hist_years_apart,
     oldest_hist_spend_year: oldest_hist_spend_data.year,
   }} />;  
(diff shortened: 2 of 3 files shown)


client/src/core/format.js
```diff
@@ -179,6 +179,7 @@ const types_to_format = {
     return`${year_int}-${ window.lang === "en" ? year_int-2000+1 : year_int+1}`;
   },
   "fiscal_year_to_year": (fiscal_year) => _.chain(fiscal_year).split("-").head().value(),
+  "year_plural_text": (year_num) => year_num > 1 ? "years" : "year",
```
Stephen-ONeil: This is something that would make more sense as a handlebar helper in `src/handlebars/helpers.side-effects.js`.

Defining it as a format is just indirectly making a handlebar helper out of it anyway, since all of the formats are registered as handlebars helpers for use in `text_maker` text. It isn't really a format though.

client/src/panels/welcome_mat/welcome_mat.yaml
```diff
@@ -8,9 +8,9 @@ no_wrap_year:
   handlebars_partial : true
   text: <span class='text-nowrap'>({{year}})</span>
 years_ago:
-  transform: [handlebars]
+  transform: [handlebars, markdown]
   en: |
-    {{actual_hist_years_apart}} years ago ({{oldest_hist_spend_year}})
+    {{actual_hist_years_apart}} {{fmt_year_plural_text actual_hist_years_apart}} ago ({{oldest_hist_spend_year}})
   fr: |
     Il y a {{actual_hist_years_apart}} ans ({{oldest_hist_spend_year}})
```
Stephen-ONeil: @cookkate or @AlexCLeduc, does the French also need a separate plural and singular case here?

client/src/core/format.js
Stephen-ONeil: It also doesn't need to be a format/handlebar helper unless its going to be used in multiple places. If we're dynamically pluralizing "year" a lot (we probably are), then if you add a utility for that then you should also try to update all the existing one off cases to use the utility.

client/src/panels/welcome_mat/welcome_mat.yaml
```diff
@@ -8,9 +8,9 @@ no_wrap_year:
   handlebars_partial : true
   text: <span class='text-nowrap'>({{year}})</span>
 years_ago:
-  transform: [handlebars]
+  transform: [handlebars, markdown]
   en: |
-    {{actual_hist_years_apart}} years ago ({{oldest_hist_spend_year}})
```
Stephen-ONeil: As an alternative to defining a new handlebars helper, you can make use of the handlebars `{{#if}}{{/if}}` block here. Something like:
```
  {{actual_hist_years_apart}} year{{#if plural_years}}s{{/if}} ago ({{oldest_hist_spend_year}})
```

Where `plural_years`, or whatever you name it, is a boolean you calculate in JS and pass in as a `text_maker` arg.

client/src/panels/welcome_mat/welcome_mat.js
```diff
@@ -158,8 +158,8 @@ const WelcomeMat = (props) => {
     latest_hist_spend_data,
     oldest_hist_spend_data,
   } = calcs;
-
-  const actual_hist_years_apart = _.parseInt( _.split(latest_hist_spend_data.year, '-') ) - _.parseInt( _.split(oldest_hist_spend_data.year, '-') ) + 1;
+  const current_year = _.parseInt( _.split( _.last(actual_history_years), '-') ) + 1;
```
Stephen-ONeil: Can get the value of `current_year` directly from `_.parseInt( run_template('{{pa_in_year_short_first}}') )`.

client/src/panels/welcome_mat/welcome_mat.js
```diff
@@ -158,8 +158,8 @@ const WelcomeMat = (props) => {
...
+  const actual_hist_years_apart = current_year - _.parseInt( _.split(oldest_hist_spend_data.year, '-') );
```
Stephen-ONeil: This is weird because `_.split(...)` returns an array of length 2, so it's a bit surprising that `_.parseInt` returns just the parsed value of the first element. I'd prefer the more explicit
```
_.chain(oldest_hist_spend_data.year).split('-').head().parseInt().value()
```

client/src/panels/welcome_mat/welcome_mat.yaml
cookkate: Yes, 'Il y a un an' or 'Il y a deux ans' would be correct. Note that the number is spelled out in words, see [here](http://dev.ea-ad.ca/welcome_matt_dead_programs/index-fra.html#orgs/program/TBC-BXA01/infograph/financial) (the "1 ans" vs "trois ans" in the headers)

ktw1016: So I did "Il y a 1 an". Is that incorrect?

cookkate:
Looks good! :+1:

PR #289, opened 2019-10-23
Use indicator seeking_to field

### LINKED ISSUE #192: Use indicator `seeking_to` field instead of inferring from target type/min/max data
The values of `target_min` and `target_max` coming out of the pipeline already depend on the value of `seeking_to` as it comes out of Titan, so re-inferring it's value from those fields in the InfoBase is a bit round-about. Using it will also simplify some code and set us up for if new more complicated cases are added in the future.

Stephen-ONeil:
Closes #192


client/src/panels/result_graphs/results_common.js
```diff
@@ -116,41 +116,36 @@ const type_by_data_type = {
...
-        } else if (!min && max){
-          return `${text_maker("result_upper_target_text")} ${formats[type_by_data_type[data_type]](+max)}` + (measure_display(meas) || "");
-        } else {
-          return target_unspecified_display; 
+        switch(seeking_to){
+          case 'target':
+            return `${text_maker("result_exact_text")} ${formats[type_by_data_type[target_type]](+target_min)}` + (measure_display(measure) || "");
+          case 'min':
+            return `${text_maker("result_upper_target_text")} ${formats[type_by_data_type[target_type]](+target_max)}` + (measure_display(measure) || "");
+          case 'max':
```
Stephen-ONeil: I'm thinking of asking the pipeline to use clearer values for `seeking_to`, unless these are what Titan uses under the hood.

Maybe:
```
  target -> exact
  min -> minimize
  max -> maximize
  range -> in_range // or just leave as range
```

Stephen-ONeil: Mainly it bugs me that the "min" case corresponds to use of the `target_max` value. Replacing minimize/maximize with more than/less than would probably be even better for clarity.

Stephen-ONeil: These values are set in Titan, not going to change for now. 

Note Titan actually has three target value columns, min, max, and point. In the "target" seeking_to case, the pipeline stores the point value from Titan in both the min and max columns to save a column in the output.

PR #291, opened 2019-10-24
New checkbox selector component

cookkate:
Looks a lot like GraphLegend but is (hopefully) more accessible and is more react-y

### DIFF
--- a/client/src/components/CheckboxSelector.js
+++ b/client/src/components/CheckboxSelector.js
@@ -1,4 +1,3 @@
-import { Fragment } from 'react';
 import './CheckboxSelector.scss';
 
 export class CheckboxSelector extends React.Component {
@@ -8,10 +7,10 @@ export class CheckboxSelector extends React.Component {
   componentDidUpdate(){
   }
   render(){
-    const { checkboxes, handleToggle } = this.props;
+    const { items, handleToggle } = this.props;
     return (
-      _.map(checkboxes, (checkbox, index) =>
-        <Checkbox key={index} id={index} toggleHandler={handleToggle} {...checkbox}/>
+      _.map(items, (item, key) =>
+        <Checkbox key={key} toggleHandler={handleToggle} {...item}/>
       )
     );
   }
@@ -21,7 +20,7 @@ export class Checkbox extends React.Component {
   constructor(props){
     super(props);
     this.state = {
-      checked: props.defaultChecked,
+      checked: props.active,
     };
   }
   render(){
@@ -32,19 +31,19 @@ export class Checkbox extends React.Component {
       this.props.toggleHandler(this.props.id,newChecked);
       this.setState({checked: newChecked});
     };
-    const borderStyle = {border: `2px solid ${color}`}
+    const borderStyle = {borderColor: color};
     return (
-      <div>
+      <div className="checkbox-selector">
         <button
-          className={checked ? "checkbox checkbox--is-checked" : "checkbox"}
+          className={"checkbox-selector--box"}
           style={checked ? { backgroundColor: color, ...borderStyle} : borderStyle }
-          tabIndex={1}
+          tabIndex={0}
           aria-pressed={checked}
-          aria-labelledby={"checkboxLabel"+id}
+          aria-labelledby={"checkboxLabel_"+id}
           onClick={onClickHandler}
         />
         <label
-          className="checkbox--label"
+          className="checkbox-selector--label"
           id={"checkboxLabel"+id}
           onClick={onClickHandler}
         >
--- a/client/src/components/CheckboxSelector.js
+++ b/client/src/components/CheckboxSelector.js
@@ -19,7 +19,7 @@ export class FancyCheckbox extends React.Component {
           className={"checkbox-selector--box"}
           style={active ? { backgroundColor: color, ...borderStyle} : borderStyle }
           tabIndex={0}
-          aria-role={"checkbox"}
+          role={"checkbox"}
           aria-checked={active}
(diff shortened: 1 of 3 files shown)


client/src/components/CheckboxSelector.js
```diff
@@ -0,0 +1,55 @@
...
+      this.setState({checked: newChecked});
+    };
+    const borderStyle = {borderColor: color};
+    return (
+      <div className="checkbox-selector">
+        <button
+          className={"checkbox-selector--box"}
+          style={checked ? { backgroundColor: color, ...borderStyle} : borderStyle }
+          tabIndex={0}
+          aria-pressed={checked}
```
Stephen-ONeil: Would be better to use `role="checkbox"` and `aria-checked={checked}` for checkbox behaviour a11y.

client/src/components/CheckboxSelector.js
```diff
@@ -0,0 +1,55 @@
...
+          onClick={onClickHandler}
+        />
+        <label
+          className="checkbox-selector--label"
+          id={"checkboxLabel"+id}
+          onClick={onClickHandler}
+        >
+          {label}
+        </label>
+      </div>
```
Stephen-ONeil: Not playing well with mac voiceover, labels and checked/unchecked state not being read. 

For labels, it's possible voiceover doesn't support `aria-labelledby`. Try giving the button an id and the `label` element a `for={id}` attribute.

For the state not being read, I'm not sure, but maybe my earlier suggestion about the role and `aria-checked` together will fix that, will need to test that though.

Stephen-ONeil: Oh, wait, on closer inspection, there's an underscore in the `aria-labelledby` id construction that's not in the label id. Should DRY that.

Nitpick, but while you're fixing that, out convention is for ids to be in kebab-case.

client/src/components/CheckboxSelector.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+  render(){
+    const { items, handleToggle } = this.props;
+    return (
+      _.map(items, (item, key) =>
+        <Checkbox key={key} toggleHandler={handleToggle} {...item}/>
+      )
+    );
+  }
+}
```
Stephen-ONeil: Small cleanup, guessing this originally needed to be a class component but I'd suggest rewriting it a function component now
```
export const CheckboxSelector = ({items, handleToggle}) => _.map(
 items,
 (item, key) => <Checkbox key={key} toggleHandler={handleToggle} {...item}/>
);
```

'nother nitpick, but `handleToggle` swapping to `toggleHandler` could confuse someone down the line.

Next level nitpick: I'd call this a callback, not a handler.

client/src/components/CheckboxSelector.js
```diff
@@ -0,0 +1,55 @@
...
+      checked: props.active,
+    };
+  }
+  render(){
+    const {id, label, color} = this.props;
+    const checked = this.state.checked;
+    const onClickHandler = () => {
+      const newChecked = !this.state.checked;
+      this.props.toggleHandler(this.props.id,newChecked);
+      this.setState({checked: newChecked});
```
Stephen-ONeil: Since this component has to use a callback to have any real behaviour, there's always going to be some parent component that's keeping track of its own copy of the checkbox items and their checked/unchecked state, right? I'd suggest making this component stateless and just using the `active` prop on every render.

cookkate: :+1:

client/src/components/CheckboxSelector.js
cookkate: :+1:

client/src/components/CheckboxSelector.js
cookkate: :+1:

client/src/components/CheckboxSelector.js
```diff
@@ -0,0 +1,37 @@
+import './CheckboxSelector.scss';
+
+export const FancyCheckboxSelector = ({items, callbackToggle}) =>
+  _.map(items,
+    (item, key) => <FancyCheckbox key={key} callbackToggle={callbackToggle} {...item}/>
+  );
```
Stephen-ONeil: I think it would make sense for this component to also take a `legend` prop, and to return something like
```
<fieldset>
  <legend>{legend}</legend>
  { _.map(
    items,
    (item, key) => <FancyCheckbox key={key} callbackToggle={callbackToggle} {...item}/>
  ) }
</fieldset>
```

Maybe with a `WellList` style border around it by default. If we ever want a non-default style somewhere we can import `FancyCheckbox` and build it as needed.

Stephen-ONeil: Hmm, bootstrap has some pretty heavy styling on `legend`'s. Might need to add a `legend-unstyled` utility class in `extended_bootstrap_css/bootstrap-wet-fixes-exstensions.scss` to reset all of that styling.

client/src/components/CheckboxSelector.js
```diff
@@ -0,0 +1,43 @@
+import './CheckboxSelector.scss';
+
+export const FancyCheckboxSelector = ({items, legend, callbackToggle}) =>
+  <div className="legend-container">
+    <fieldset>
+      <legend className="checkbox-selector__legend">{legend}</legend>
```
Stephen-ONeil: BEM nitpick, `checkbox-selector__legend` ends up as a sibling of a bunch of `checkbox-selector`s, not a child. Maybe make the containing div or fieldset `checkbox-selector`, and the current `checkbox-selector`s `checkbox-selector__option`s?

PR #292, opened 2019-10-24
Cleanup of general_utils.js

cookkate:
I found a "Select" object in there that looks super old, plus another function for escaping single quotes that isn't used anywhere, so I deleted them.


cookkate:
My bad, I somehow missed that bit in my commit.

PR #295, opened 2019-10-25
Actually fix the Chrome cookie warnings

Stephen-ONeil:
#286 didn't do it, my bad for testing in Chromium because it looks like those warnings are Chrome-only in the first place (at least right now).

There's another warning about a canada.ca cookie, haven't tracked the source of that one down yet, but might be google analytics/google tag manager related too.


Stephen-ONeil:
Wait, aha, this "fixed" the warning by breaking the cookies. `ga` doesn't seem to have good error logging, just silently fails.

PR #297, opened 2019-10-25
Add icon array shapes with css

### LINKED ISSUE #196: Waffle diagram should use different shapes
It's not currently colour blind friendly.

cookkate:
An attempt to fix #196.

Not sure I like it, honestly.

Screen Shot 2019-10-25 at 2 48 34 PM

### DIFF
--- a/client/src/common_css/charts.scss
+++ b/client/src/common_css/charts.scss
@@ -149,6 +149,4 @@ div.d3-list {
     width: 20px;
     height: 20px;
   }
-
-
 }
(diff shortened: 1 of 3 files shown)


Stephen-ONeil:
Could try leaving them squares and giving them textures instead? Maybe stripes of different orientations, `background: repeating-linear-gradient(...)` and the like

Stephen-ONeil:
I take that back, I forgot how small the squares get on the Gov level graphic, texturing those would get visually incomprehensible.

cookkate:
This isn't a good solution.

PR #314, opened 2019-10-30
Indicator tombstone panel

cookkate:
Now that indicators are subjects, we can generate a tombstone panel to display stuff for a single indicator.

Might delete the infographic stuff (unless it turns out to be necessary...) as I don't think we want more than a single panel route for a tombstone type panel.

- [x] query to get a single indicator by ID
- [x] figure out how/when do load that data (will the single panel route work?)
- [x] make the panel
- [x] turn the indicators from the flat table into links
- [x] don't use ensure_loaded: results, just have the panel make its own graphQL query, that way I can take out the hacky faux subject stuff


client/src/models/populate_results.js
```diff
@@ -238,6 +249,13 @@ function extract_flat_data_from_results_hierarchies(hierarchical_response_data){
     indicators = [],
     pi_dr_links = [];
 
+  const process_indicator = (indicator) => {
+    indicator.target_year = _.isEmpty(indicator.target_year) ? null : parseInt(indicator.target_year);
+    indicator.target_month = _.isEmpty(indicator.target_month) ? null : parseInt(indicator.target_month);
```
Stephen-ONeil: Nitpicky, and fact check me on this, but I think empty target year values come out of the API as a nulls already.

Stephen-ONeil: Less nitpicky reason to delete this, I think this is the only part of `crawl_hierachy_level` that mutates the data passed to it. Possibly irrelevant, because I'm not sure the original object is used again after this, but still something to generally avoid.

client/src/panels/result_graphs/indicator_display.js
```diff
@@ -0,0 +1,64 @@
...
+  componentDidMount(){
+    const {
+      match: {
+        params: { id },
+      },
+    } = this.props;
+    ensure_loaded({
+      subject: {id: id, level: "indicator"},
+      results: true,
+    })
```
Stephen-ONeil: Might be best to break away from the old `ensure_loaded` paradigm and have this panel just directly make its own GraphQL request.

client/src/models/results.js
```diff
@@ -311,10 +311,10 @@ class Indicator {
   static create_and_register(def){
     const { id, result_id } = def;
 
-    if(this.lookup(id)){
+    if(this.lookup(id).doc){ // hack to allow indicators as subjects before they're loaded
```
Stephen-ONeil: As discussed, these hacks won't be needed if the individual indicator route makes its own GraphWL queries.

client/src/models/populate_results.js
cookkate: wontfix in favour of #321 😝

client/src/panels/result_graphs/indicator_display.js
```diff
@@ -0,0 +1,141 @@
...
+    .then( (response) => {
+      Indicator.create_and_register(process_indicator(response.data.root.indicator));
+      return Promise.resolve();
+    })
+    .catch(function(error){
+      const resp_time = Date.now() - time_at_request;     
+      log_standard_event({
+        SUBAPP: window.location.hash.replace('#',''),
+        MISC1: "API_QUERY_FAILURE",
+        MISC2: `Results, took  ${resp_time} ms - ${error.toString()}`,
```
Stephen-ONeil: We'll be able to tell from the location I guess, but could you also change MISC2 to identify this as the indicator query?

cookkate: Oops, yeah I meant to do that. fixed.

PR #317, opened 2019-10-30
More server unit tests

Stephen-ONeil:
Just a start for now. Will be adding more in follow up PRs.


server/src/server_utils.js
```diff
@@ -0,0 +1,59 @@
+import { decompressFromBase64 } from 'lz-string';
+import md5 from 'md5';
+
+import _ from 'lodash';
+
+// Side effect alert: this function mutates the suplied request object so that the conversion persists to subsequent server midleware
+// ... bit of a hack, and I'm not just talking about a function having side effects
```
cookkate: yeah this is super weird....but whatever it's a testing script :P

Stephen-ONeil: Oh, no, that's prod code.



Too late, you approved!

PR #318, opened 2019-10-30
Tweak name of text based version to text-only version

Stephen-ONeil:
"text-only" seems to be the common phrase. Definitely a little more accurate than "text based".

Could someone take a second look at the french? There's also a short (partial sentence) untranslated, if anyone things they know what it should be. If not, I'll get it translated proper.

### DIFF
--- a/client/src/InfoBase/index_data.js
+++ b/client/src/InfoBase/index_data.js
@@ -136,15 +136,19 @@ const index_lang_lookups = {
     en: "index-basic-eng.html",
     fr: "index-basic-fra.html",
   },
+  a11y_version_title: {
+    en: "Text-only version",
+    fr: "Version texte seulement",
+  },
   a11y_version_text: {
-    en: "Text based version",
-    fr: "Version basé en texte",
+    en: "recommended for screen reader users",
+    fr: "TODO",
   },
   standard_version_url: {
     en: "index-eng.html",
     fr: "index-fra.html",
   },
-  standard_version_text: {
+  standard_version_titile: {
     en: "Graphical version",
     fr: "Version graphique",
   },
--- a/client/src/InfoBase/index_data.js
+++ b/client/src/InfoBase/index_data.js
@@ -142,7 +142,7 @@ const index_lang_lookups = {
   },
   a11y_version_text: {
     en: "recommended for screen reader users",
-    fr: "TODO",
+    fr: "recommandé pour les utilisateurs de lecteur d'écran",
   },
   standard_version_url: {
     en: "index-eng.html",
--- a/client/src/InfoBase/index_data.js
+++ b/client/src/InfoBase/index_data.js
@@ -138,11 +138,11 @@ const index_lang_lookups = {
   },
   a11y_version_title: {
     en: "Text-only version",
-    fr: "Version texte seulement",
+    fr: "Version texte",
   },
   a11y_version_text: {
     en: "recommended for screen reader users",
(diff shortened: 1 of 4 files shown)


client/src/home/a11y-home.yaml
```diff
@@ -29,14 +29,14 @@ home_a11y_subject_section_title:
 home_a11y_non_a11y_redirect_warning:
   transform: [markdown]
   en: |
-    You have been redirect to the text based InfoBase from a page with no text-only equivalent. 
+    You have been redirect to the text-only InfoBase from a page with no text-only equivalent. 
     Pages with no text-only equivalent are purely for visualization purposes and are not the sole presentation of any data in the InfoBase.
     Any information which was presented on the page you were redirected from can be found in the [report builder](#rpb).
   fr: |
-    Vous avez été redirigé vers une version d'InfoBase textuel à partir d'une page sans équivalent textuel.
+    Vous avez été redirigé vers une version d'InfoBase texte seulement à partir d'une page sans équivalent texte seulement.
```
cookkate: I don't think this change was necessary and "version d'InfoBase texte seulement" doesn't make sense

client/src/home/a11y-home.yaml
```diff
@@ -29,14 +29,14 @@ home_a11y_subject_section_title:
...
     Les pages sans équivalent textuel servent à visualiser des donées et ne constituent pas la seule présentation de ces données dans l'InfoBase.
     Toute information présentée sur la page à partir de laquelle vous avez été redirigé se trouve dans le [générateur de rapports](#rpb).
 
 home_a11y_desc:
-  en: Text based Infobase home for improved accessibility
-  fr: Version texte de l’Infobase pour meilleure accessibilité
+  en: Text-only Infobase home, for improved accessibility
+  fr: Version texte seulement de l’Infobase, pour meilleure accessibilité
```
cookkate: Ditto, the French version doesn't really distinguish between "text based" and "text only"

Stephen-ONeil: I got there from combing through https://www.linguee.com/english-french/translation/text+only.html, which I use when trying to find what the accepted french equivalent for specific common phrases are, but it was pretty inconsistent on this one. Thanks for checking it over!

Stephen-ONeil:
> Another change that we'll need to update index.html for.

Yup, I guess I should send out more emails over that again... probably should try and snare a new contact over there.

client/src/InfoBase/index_data.js
```diff
@@ -136,15 +136,19 @@ const index_lang_lookups = {
     en: "index-basic-eng.html",
     fr: "index-basic-fra.html",
   },
+  a11y_version_title: {
+    en: "Text-only version",
+    fr: "Version texte seulement",
```
cookkate: Actually I think this should be "version texte" as well, because otherwise the full phrase is "Version texte seulement, 
 recommandé pour les utilisateurs de lecteur d'écran" which looks like the comma is an error and you're saying that the text version is only recommended for screen reader users.

client/src/InfoBase/index_data.js
```diff
@@ -136,15 +136,19 @@ const index_lang_lookups = {
...
+  },
   a11y_version_text: {
-    en: "Text based version",
-    fr: "Version basé en texte",
+    en: "recommended for screen reader users",
+    fr: "recommandé pour les utilisateurs de lecteur d'écran",
```
cookkate: Also just realized that version is feminine, so should be "recommandée"

PR #322, opened 2019-11-01
Server test coverage status badge

Stephen-ONeil:
Updated by CI runs on `master`, displayed in the root and server/ README's.

Looks like Test coverage status badge


Stephen-ONeil:
Ah, hmm, the CI produced coverage stats are from tests run on the transpiled build, that's not ideal. I think that was because the text images used an older node, although I'v updated them somewhat recently.

On the other hand, the transpiled tests probably are the ones we want running in CI since the deployed version is also transpiled. At least for the snapshot (end to end) tests.

PR #326, opened 2019-11-04
Update text on auth/exp/planned spending data

### LINKED ISSUE #116: Adding more text for auth, exp, spending panel
Just a reminder so it doesn't slip but we should add additional text about lapse (the difference). Julie can potentially help with writing up the text

cookkate:
Fixes #116. Text update, so French needs to happen still. Also will have to check with someone on the definition for lapse (which I added to the glossary)

Soliciting opinions before I start doing that.

Screenshot of the text in action:

Screen Shot 2019-11-04 at 3 01 28 PM


Stephen-ONeil:
Seems good to me, no strong opinion. 

Reminder that glossary changes have to made on in the csv on GCDocs ultimately.

cookkate:
Yep, just made the change here to have it reflected in the dev link if someone wanted to look.

cookkate:
The term "lapse" is not quite correct here since many of the funds are carried over. Save lapsing -- for a new panel of its own (maybe a coop student project?)

PR #327, opened 2019-11-04
Update nivo

Stephen-ONeil:
Going to need a lot of eyes on this dev build (once it's ready) to look for chart errors. I'll do the IE testing.


Stephen-ONeil:
Hmm, package patches didn't get applied in CI?

Stephen-ONeil:
TODOs:
  - [ ] fix patch-package in CI
  - [ ] GOCO graph hover fading the wrong bars
  - [ ] replace legends with our own
  - [ ] sort items in line slice tooltips 
  - [ ] Spending by Standard Objects line graph colours

ktw1016:
- [ ] Color issue: http://dev.ea-ad.ca/update-nivo/index-eng.html#orgs/dept/1/infograph/financial/.-.-(panel_key.-.-'spend_by_so_hist)

- [x] No data points (I think I know this one, I can fix it): http://dev.ea-ad.ca/update-nivo/index-eng.html#orgs/dept/146/infograph/financial/.-.-(panel_key.-.-'auth_exp_prog_spending)

ktw1016:
So there's still an error in goco on spending area text opacity when you click on the graph.. It should be faded out but it doesn't. (It actually does fade out in `handleClick`, but I think something else gets triggered after to set unfade it. and I can't figure that out)

Stephen-ONeil:
Low priority, putting on-hold to focus on bugstravaganza.

Stephen-ONeil:
There are so many little unfixable (without going full hack, and I can't get in to the habit of hitting every nail with the `patch-package` hammer) headaches here, I'm going to drop this PR. 

New plan is to wait for nivo to _**actually stabilize**_ (or at least have a new feature worth the update pain) before updating it again. Not worth going through this more than once. Risk that it ends up meaning our current code is too far out of sync to simply update, but in that case we can just do a full re-write of everything using nivo and probably come out the better for it.

ktw1016:
rip nivo😱

Stephen-ONeil:
😞

PR #329, opened 2019-11-05
internal services panel

### LINKED ISSUE #83: internal service FTE (historical? spending?) panel
dept level infograph needs new panel showcasing proportion of internal services programs

* if historical trends are not completely static, it should use a stacked isc vs non-isc bar chart
* spending might be interesting

ktw1016:
fixes #83

### DIFF
--- a/client/src/panels/internal_services/isc.js
+++ b/client/src/panels/internal_services/isc.js
@@ -1,5 +1,4 @@
 import text from './isc.yaml';
-import { GlossaryEntry } from '../../models/glossary.js';
 import { 
   InfographicPanel, 
   Subject, 
@@ -14,23 +13,20 @@ import {
 const { Gov, Tag } = Subject;
 const { std_years } = years;
 const { GraphLegend } = declarative_charts;
-
 const { text_maker, TM } = create_text_maker_component(text);
-
 
 export const declare_internal_services_panel = () => declare_panel({
   panel_key: "internal_services",
   levels: ["dept"],
-  panel_config_func: (level, panel_key) => ({
-    depends_on: ['programFtes', "programSpending"],
+  panel_config_func: () => ({
+    depends_on: ['programFtes'],
     title: "internal_service_panel_title",
-    calculate(subject, info){
+    calculate(subject){
       const { programFtes } = this.tables;
   
       const isc_crsos = _.filter(subject.crsos, "is_internal_service");
       if(_.isEmpty(isc_crsos)){
-        //org has no isc programs 
-        return false;
+        return false; //org has no isc programs 
       }
   
       const isc_tag = Tag.lookup("GOC017");
@@ -77,9 +73,6 @@ export const declare_internal_services_panel = () => declare_panel({
         },
       } = calculations;
 
-      const more_footnotes = [{
-        text: GlossaryEntry.lookup("INT_SERVICES").definition,
-      }].concat(footnotes);
       const years = _.map(std_years, yr => run_template(yr));
       const label_keys = [text_maker("internal_services"), text_maker("other_programs")];
       const colors = infobase_colors();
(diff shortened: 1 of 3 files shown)


client/src/panels/internal_services/isc.yaml
```diff
@@ -7,7 +7,7 @@ internal_service_panel_text:
   en: |
     Last year ({{pa_last_year}}), **{{fmt_percentage1 isc_fte_pct}}** of {{gl_tt "employees" "FTE"}} in {{subj_name subject}} worked for internal service programs. In comparison, government as a whole uses **{{fmt_percentage1 gov_isc_fte_pct}}** of its workforce for internal services.
   fr: |
-    L'an dernier ({{pa_last_year}}), **{{fmt_percentage1 isc_fte_pct}}** des {{gl_tt "FTE" "employés"}} dans l'organisation {{subj_name subject}} travaillaient pour des programmes de services internes.
+    L'an dernier ({{pa_last_year}}), **{{fmt_percentage1 isc_fte_pct}}** des {{gl_tt "employés" "FTE"}} dans l'organisation {{subj_name subject}} travaillaient pour des programmes de services internes.
```
cookkate: Need to add the second sentence in French:

"En comparaison, **{{fmt_percentage1 gov_isc_fte_pct}}** de l’effectif de le gouvernement travaillaient pour des programmes de services internes."

client/src/panels/internal_services/isc.js
```diff
@@ -81,27 +69,26 @@ export const declare_internal_services_panel = () => declare_panel({
...
           isc_fte,
-  
           series,
         },
       } = calculations;
-  
+
       const more_footnotes = [{
         text: GlossaryEntry.lookup("INT_SERVICES").definition,
       }].concat(footnotes);
```
cookkate: I'm not sure if we want to keep this hardcoded footnote instead of using the standard glossary tooltip that we usually use for glossary definitions. @Stephen-ONeil ?

Stephen-ONeil: I agree, moving it to a glossary tooltip in the panel text would be better 👍

cookkate:
Sorry, couple more comments:

- The program spending data is listed as a source but isn't used in this panel
- I would bump it up a few places. Probably before the two panels that are just tables
- Suggest changing the panel title to "Full-time employees in Internal Services" / "L'effectif des services internes"

client/src/panels/internal_services/isc.yaml
ktw1016: Another thing to note here from Val: Wording of `Last year (2017-18)` should change to `In 2017-18`. Just note for myself to fix

client/src/panels/internal_services/isc.js
ktw1016: 👍👍 thanks for the good suggestions kate!

client/src/panels/internal_services/isc.js
```diff
@@ -28,44 +13,43 @@ import {
...
     title: "internal_service_panel_title",
-    calculate(subject,info){
+    calculate(subject){
       const { programFtes } = this.tables;
   
       const isc_crsos = _.filter(subject.crsos, "is_internal_service");
       if(_.isEmpty(isc_crsos)){
-        //org has no isc programs 
-        return false;
+        return false; //org has no isc programs
```
cookkate: Ok, the problem with FCAC is that it *does* technically have an internal services CR (as of 2018-19), but it doesn't actually have any internal service spending or FTEs (because we're not done 2018-19 yet). Partly that will fix itself after public accounts are out, but the 5-year graph will be a little weird still (because it will have 0s for the four previous years).

What you will need to do to calculate "first active year" and "last active year" the way the people tables do. You can see how it's done for employee age in lines 130-148 of `orgEmployeeAvgAge.js`.

PR #334, opened 2019-11-05
Docker best practices

### LINKED ISSUE #331: Container image hosting cost
With all the additional CI jobs I've set up, our GCloud image hosting costs have really gone up lately. Look in to if there are suitable images available on dockerhub, etc. If not, maybe try slimming ours down? Worst case, slam a bunch of the CI jobs back together (rather wouldn't) or make sure the budget covers our higher costs (we're probably fine, for the rest of the year at least).

Stephen-ONeil:
Use an unprivileged user account in docker images.

Also re-hosting on Docker Hub, to close #331

### DIFF
--- a/dockerfiles/test/Dockerfile
+++ b/dockerfiles/test/Dockerfile
@@ -3,16 +3,19 @@ FROM mhart/alpine-node:latest
 ENV SERVICE_USER test-image-user
 ENV SERVICE_HOME /home/${SERVICE_USER}
 
+# commom
 RUN adduser -h ${SERVICE_HOME} -s /sbin/nologin -u 1000 -D ${SERVICE_USER} && \
   apk update && apk upgrade && \
   apk add --no-cache \
-    udev \
-    xvfb \
-    chromium \
-    ca-certificates \ 
-    mongodb \
     git \
-    bash
+    bash \
+    ca-certificates \
+    mongodb
+# image specific
+RUN apk add --no-cache \
+  udev \
+  xvfb \
+  chromium
 
 RUN mkdir -p /data/db # create default mongodb data directory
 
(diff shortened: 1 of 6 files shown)


dockerfiles/test/Dockerfile
```diff
@@ -1,11 +1,29 @@
...
-    ca-certificates \ 
-    mongodb \
-    git
-RUN mkdir -p /data/db # create default mongodb data directory
-RUN apk add --no-cache bash
\ No newline at end of file
+# by :13.0.1 at least, alpine-node started including comming with old versions of chromium and mongodb installed
+# No luck upgrading or uninstalling them, and they were breaking testcafe and cleanup_dev_dbs.sh respectively
+# 12.9.0 was the previous image I built from/knew to be good, not necessarily the latest pre-funny buisness
+# ... can't stay locked to this forever though
```
cookkate: 13.0.1 is the just-released version of alpine-node, right? so this might be a temporary problem, or at least not something we need to fix immediately.

Stephen-ONeil: Yeah, 12.9.0 is from late August this year. Fingers crossed future alpine-node drops, or at least upgrades, its mongodb/chromium dependencies. We shouldn't get too out of date though, becomes a security hole if we aren't updating our images (limited somewhat because these are just for CI, not prod).

I might dig in to WHY it requires them some time, as doesn't seem like they should be in there, could be something they overlooked in the latest releases.

cookkate: Especially since they've self-branded as "minimal"

Stephen-ONeil: Exactly, chromium and mongodb are practically maximal

Issue #342, opened 2019-11-07
Panel module organization


Stephen-ONeil:
It's all over the place (dept profile panel is in `panels/igoc/igoc_panel.js` but the program equivalent is in `panels/intro_graphs/profile_panels.js`, etc.), it would make more sense to group by bubble menu item (dept, program, etc. profile panels should all be in `panels/about` or some such).

Issue #44, opened 2019-04-11
Sort and group footnotes in RPB


Stephen-ONeil:
They aren't displayed well, and I think it's often un clear what they link to (are they gov footnotes, specific dept footnotes, etc).

Stephen-ONeil:
If I do this on panels as well, will need to deal with the cases where panels throw in mocked footnote objects through their render function, I'm sure it's done more than once and I'm sure they don't bother mocking all of the footnote fields fully.

PR #343, opened 2019-11-07
Key concepts cleanup

### LINKED ISSUE #226: Review the "Some things to keep in mind..." content
A few of these are, or should be, glossary items and are, or should be, found in glossary tooltips throughout the page instead. Some should also be reorganized or clarified.

Stephen-ONeil:
Closes #226

Started to get carried away and do #342, but stopped myself.


client/src/panels/people/employee_type.js
```diff
@@ -22,6 +24,12 @@ const { tenure } = businessConstants;
 const { A11YTable } = declarative_charts;
 
 
+const employee_type_glossary_footnotes = _.map(
+  ['INDET_PEOPLE', 'TERM_PEOPLE', 'CASUAL_PEOPLE', 'STUD_PEOPLE'],
+  (glossary_key) => ({text: GlossaryEntry.lookup(glossary_key).definition})
+);
```
Stephen-ONeil: Just occurred to me that this kind of hack will interfere with (or at least be a special case in) #44, if it goes in the direction of being more than just an RPB feature.

cookkate: Can we generate footnotes from certain glossary items in `populate_footnotes.js`? That avoids the multiple sources of truth problem, and lets us display them like footnotes if we want.

Stephen-ONeil: Could do. I was also thinking that it might be good to tag panels with relevant glossary entries, make it standard. There's lots of panels that might have relevant definitions that don't show up exactly in text and, as a result, don't get direct glossary tooltips/links. Maybe put them in a details drawer like the footnotes (or rename the existing one to Footnotes & Definitions), or add a little dictionary icon to the utils bar to bring up a definitions modal.

Stephen-ONeil: I guess I should capture that thought in an issue, not doing it in this PR.

PR #348, opened 2019-11-08
Add line of panel glossary items

### LINKED ISSUE #345: Tag panels with relevant glossary entries
Thinking that it might be good to tag panels with relevant glossary entries, make it standard. There's lots of panels that might have relevant definitions that don't show up exactly in text and, as a result, don't get direct glossary tooltips/links. Maybe put them in a details drawer like the footnotes (or rename the existing one to Footnotes & Definitions), or add a little dictionary icon to the utils bar to bring up a definitions modal.

Right now, a couple panels include glossary entires with their footnotes, but not in a standard (or robust) way. Preferable (IMO) to having the same text in both the footnote and glossary datasets though, since they'd risk getting out of sync.

cookkate:
Frees us from having to include text for every glossary definition we want to show. Generate a line of glossary tooltip-generating items underneath the source links.

Closes #345

Still TODO:

- [x] update the glossary input file on GCdocs
- ~[ ] add INT_SERVICES to footnotes for the internal services FTE panel~ not really necessary, it's in the text and not shoehorned


client/src/core/PanelGraph.js
```diff
@@ -183,6 +182,14 @@ class PanelGraph {
     }
   }
 
+  get_glossary_keys(){
+    if (this.glossary_keys === false){
+      return [];
+    } else {
+      return this.glossary_keys;
+    }
+  }
```
ktw1016: Did the ternary operator not work for this one? like `this.glossary_keys ? this.glossary_keys : []`

client/src/components/Panel.js
```diff
@@ -42,7 +43,30 @@ const PanelSource = ({links}) => {
   );
 };
 
-export const Panel = ({allowOverflow, title, otherHeaderContent, children, sources, footnotes }) => (
+const PanelGlossary = ({keys}) => {
+  if(_.isEmpty(keys)){
+    return null;
+  }
```
ktw1016: I suppose this one can also be written inside `return` with ternary operator as well? Although this is still clean

client/src/components/glossary_components.js
```diff
@@ -0,0 +1,50 @@
+import { IconQuestion } from '../icons/icons.js';
+import { GlossaryEntry } from '../models/glossary.js';
+import { glossary_href } from '../link_utils.js';
+
+const GlossaryTooltipItem = ({id, arrow_selector, inner_selector, children}) => (
+  <a
+    href={( window.is_a11y_mode || window.feature_detection.is_mobile() ) ? glossary_href(id) : null}
```
Stephen-ONeil: Were the tooltips not working on mobile then?

client/src/core/PanelGraph.js
Stephen-ONeil: Or `this.glossary_keys || []`

client/src/components/glossary_components.js
Stephen-ONeil: Oops, should have done this as a review. This is my one question, approve the rest!

cookkate: They do, this was wrong!

client/src/components/Panel.js
cookkate: It could, but this code matches the other function (the source links) and I like it this way 😝

PR #351, opened 2019-11-12
Transfer payments by province

Stephen-ONeil:
I figured it'll be better to share legends and graphs altogether with people's Canada graph. Let me know if there's a problem!

Taewan (not Stephen lol)


ktw1016:
why is this one on hold?

Stephen-ONeil:
Just cause I wasn't planning to work on it right away, was just opening the PR to share the data. If you want to take over this branch you can go ahead and remove the label! (or maybe close this PR and open your own, so that I'll be able to review it eventually)

ktw1016:
just fyi I am currently working on this branch

ktw1016:
TODO:
- [ ] Report builder description text
- [ ] Figure out footnotes (both rpb, panel)
- [ ] Format legend
- [ ] Small graph adjustments (NCR arrow, center ON text to the province)

client/src/panels/people/employee_prov.js
```diff
@@ -293,16 +129,22 @@ const calculate_common = (data) => {
...
     const { orgEmployeeRegion } = this.tables;
-    return calculate_common( people_years.map( year => orgEmployeeRegion.prov_code(year, false) ) );
+    return calculate_common( context_years.map( year => orgEmployeeRegion.prov_code(year, false) ) );
   },
   dept: function(subject){
     const { orgEmployeeRegion } = this.tables;
-    return calculate_common( people_years.map( year => orgEmployeeRegion.prov_code(year, subject.unique_id) ) );
+    return calculate_common( context_years.map( year => orgEmployeeRegion.prov_code(year, subject.unique_id) ) );
   },
 };
```
Stephen-ONeil: Love how much you cleaned this whole file up!

client/src/panels/people/employee_prov.js
```diff
@@ -9,148 +9,18 @@ import {
...
-          prov_callback("Canada");
-        }
-      }, 200);
-      active_prov = false;
-    });
-
-    canada_graph.render();
-  }
-}
+const context_years = people_years;
```
Stephen-ONeil: `context_years` feels a lot less specific/informative than `people_years`. I'd prefer if `people_years` was used in this file (and only passed in to the chart config as `context_years`).

client/src/panels/transfer_payments/tp_by_region.js
```diff
@@ -0,0 +1,119 @@
...
+    .reduce( (sum, value) => sum + value, 0)
+    .value();
+  
+  return _.chain(provinces)
+    .map((province, prov_code) => {
+      if(prov_code!="onlessncr" && prov_code!="qclessncr" && prov_code!="ncr"){
+        const yearly_tp = _.map(years_by_province, (year_by_province) => year_by_province[prov_code]);
+        const formatted_yearly_tp = _.map(years_by_province, (year_by_province) => formats["compact2_written_raw"](year_by_province[prov_code]));
+        const five_year_avg_share = d3.sum(yearly_tp)/total_tp;
+        const formatted_avg_share = formats["percentage1_raw"](five_year_avg_share);
```
Stephen-ONeil: All the average share stuff was only ever done in the people tables. I figure it makes sense there because the demographics are fairly stable, and people are generally curious about how demographics are split up by percentage. I think with money, people prefer to get hard values (even if what they really want probably is the distribution in this case), and fluctuations tend to be more common/meaningful. Big dollar values are just exciting a lot more exciting than percents!

Just my opinion, nothing to back these claims up really.

client/src/charts/extended_canada.js
```diff
@@ -0,0 +1,217 @@
+
+import {
+  formats,
+  run_template,
+  businessConstants,
+  declarative_charts,
+  NivoResponsiveHBar,
+  hex_to_rgb,
+  Canada,
+} from "../panels/shared.js";
```
Stephen-ONeil: `panels/shared.js` is only the actual source for `hex_to_rgb`, everything else here is just being re-exported by `shared.js`. It'd be better to import everything directly from its actual source (and maybe even move `hex_to_rgb` from `shared.js` to `general_utils.js` or something). Less circular dependency risk, and less confusion about where to find the original definition of any given dependency.

Up until recently, the `shared.js` pattern was bad for bundle optimization too (not an issue currently, although things could be a bit better if we broke it up still). Might still want to do that some day.

client/src/charts/extended_canada.js
```diff
@@ -0,0 +1,217 @@
...
+const graph_color = window.infobase_color_constants.secondaryColor;
+
+const get_graph_color = (alpha) => {
+  const rgb = hex_to_rgb(graph_color);
+  return rgb && `rgba(${rgb.r}, ${rgb.g}, ${rgb.b}, ${alpha || 1})`;
+};
+
+const format_prov_data = (prov, years_by_province, context_years) => {
+  var prov_data;
+  if(prov==="Canada"){
```
Stephen-ONeil: A bit weird to me that "Canada" is a special `prov` case. I feel like it makes more sense to have `prov` be false/null when nothing's hovered, with everything set up so that the Canada case is the default/fallback. Then instead of checking `prov==="Canada"` in multiple places you just need to do a single check for if `prov` has a truthy of falsy value before either using defaults or applying the general province logic.

client/src/panels/people/employee_prov.js
Stephen-ONeil: That wasn't a comment to do with that specific line, I just left it somewhere near the bottom of the file,

client/src/panels/transfer_payments/tp_by_region.js
ktw1016: okay yeah that makes sense so are you suggesting to remove average share column? (As you know this is for A11y table column!) Or something like how much money increased/decreased over 5 years? Just asking to be clear!

Stephen-ONeil: I say drop it, yeah. Thanks!

client/src/panels/panel_declarations/finances/transfer_payments/tp_by_region.js
```diff
@@ -0,0 +1,108 @@
...
+  dept: function(subject){
+    const { orgTransferPaymentsRegion } = this.tables;
+    return calculate_common( std_years.map( year => orgTransferPaymentsRegion.prov_code(year, subject.unique_id) ) );
+  },
+};
+
+const prepare_data_for_a11y_table = (years_by_province) =>
+  _.chain(provinces)
+    .map((province, prov_code) => {
+      if(prov_code!="onlessncr" && prov_code!="qclessncr" && prov_code!="ncr"){
```
Stephen-ONeil: This pattern can be cleaned up, IMO, with `_.includes`, e.g. `!_.includes(["onlessncr", "qclessncr", "ncr"], prov_code)`.

Side note, should be `!==`'s. Not specifically relevant here, but generally better practice (same as using `===` instead of `==`, the `==/!=` type coercion is usually not desired and frequently not what you'd expect).

client/src/tables/orgTransferPaymentsRegion.js
```diff
@@ -0,0 +1,133 @@
...
+    this.add_col({
+      "type": "percentage1",
+      "nick": "five_year_percent",
+      "header": trivial_text_maker("five_year_percent_header"),
+      "description": {
+        "en": trivial_text_maker("five_year_percent_description"),
+        "fr": trivial_text_maker("five_year_percent_description"),
+      },
+      "formula": people_five_year_percentage_formula("region",std_years),
+    });*/
```
Stephen-ONeil: Could delete

client/src/charts/canada/canada_d3_component.js
```diff
@@ -36,7 +36,7 @@ const ordering = {
   pe: 16,
 };
 
-export class Canada {
+export class Canada_D3_Component {
```
Stephen-ONeil: Little nitpick (this is my fault, I suggested the wrong name for this file in me earlier feedback), by convention this (and the file name) would be CanadaD3Component. I forgot it was a class.

client/src/charts/canada/canada.js
```diff
@@ -0,0 +1,213 @@
+
+import {
+  formats,
+  run_template,
+  businessConstants,
+  declarative_charts,
+  NivoResponsiveHBar,
+} from "../../panels/panel_declarations/shared.js";
```
Stephen-ONeil: Please break these up and import each of them from their original source rather than from the `panel_declarations/shared.js` re-exporter. More transparent about where dependencies come from that way, and less likely to result in circular dependencies, etc.

`shared.js` should only be used locally to the panel declarations (and even then I'd like to eventually get rid of the pattern).

client/src/charts/canada/canada_d3_component.js
ktw1016: oh true yup yup

client/src/panels/panel_declarations/finances/transfer_payments/tp_by_region.js
ktw1016: ah okay!! didn't actually know `!==` was a thing until now

ktw1016:
I just noticed that I'm the reviewer for this PR lol

Stephen-ONeil:
Hahaha, yup! It's fine, I'll be closing it out while you're away so I'll go back to actually being the author eventually.

cookkate:
Comments from Ross:

Thanks for sharing. This will be a great addition to InfoBase. A couple of comments/suggestions:

 

·         Perhaps change the Not Available to “Unknown”?

o   Replace the N/A with a question mark? (N/A makes me think of Not Applicable)

·         For the title, maybe “Transfer payment by recipient region”?

o   We will probably want to add a footnote to add some more context (or disclaimers)

·         For the legend, rather than $0+, $10.00 B+ etc.., could we have ranges? I.e. $0-$10.0 B, $10.0 B - $20.00 B etc..

o   The last one should stay as $50.00+ B

·         I noticed that Taewan opened an issue on Github for the idea of presenting transfer payments on a per capita basis. I think that this a great idea and given the current focus on regional discrepancies, it might be important to provide this context.

o   Someone mentioned to me to the other day that InfoBase would be more useful if it provided more context.

Stephen-ONeil:
Need a more granular tag/footnote key than "GEO", now that it's shared between a people and a financial table.

rossbm:
A slight tweak to my suggestion for the English title: _Transfer payments by region of the recipient_ A bit long, but slightly more accurate (the money doesn't go to a region directly, but someone located in a specific region). The French would be _Paiements de transfert par région du bénéficiaire_

client/src/core/format.js
```diff
@@ -41,7 +41,8 @@ const compact = (precision, val, lang, options) => {
   let symbol;
   let new_val;
   if (val === 0){ 
-    return "0";
+    new_val = 0;
+    symbol = "";
```
ktw1016: Is this okay?

cookkate: Probably? It looks like that could generate a bug in cases where `raw` wasn't set.

ktw1016:
Any suggestions for text is welcome!

cookkate:
This should be added to footnotes.csv with the key 'TP_GEO'. I have added it to the gcdocs input spreadsheet.

Recipient location is not available for transfer payments under $100,000, for recoveries and for cases where the recipient’s identity is withheld from Public Accounts, volume III section 6 for security or privacy reasons. 

La location du récipiendaire n’est pas disponible pour les paiements de transferts de moins de 100 000 $, pour les recouvrements et lorsque l’identité du récipiendaire n’a pas été divulguée au Comptes publics, volume III section 6 pour des raisons de confidentialité ou sécurité.

cookkate:
Oh, also the footnote that's there is for Phoenix and isn't related to this data.

cookkate:
Here's my suggestion for text. The first instance of 'transfer payments' should have a glossary tooltip.

In *2018-19*, *${total sum}* in transfer payments above $100,000 was paid out. Recipients in *${province}* received the largest average share *(${percent_of_total})* over the past five fiscal years.

Hover over or tap a province or territory in the map below to see the total transfer payments over the last five years. Note that the scale may change to accommodate the data displayed.

Stephen-ONeil:
The new table has the `GEO` tag, which used to only be used by the people table (so the Phoenix footnote is tagged GEO).

Maybe the people GEO (in the people region table and the footnotes, both in the repo and GCDocs) should become PPL_GEO and the new table's GEO should be TP_GEO as suggested? Should they both still have the GEO tag?

cookkate:
Not unless we think we will have footnotes that apply to all geographic data? We don't currently.

client/src/charts/canada/canada.js
```diff
@@ -165,16 +165,19 @@ export class Canada extends React.Component{
...
         color: get_graph_color( color_scale(tick) ),
-      })
-    );
+      });
+      result.prevTick = tick;
+      return result;
+    }, {
+      prevTick: null,
+      items: [],
+    });
```
Stephen-ONeil: The standard forth argument to a lodash array/collection/object util is the iterated array/collection/object itself, so you don't need this pattern of storing the `prevTick` in the reduce accumulator, could write this as something like:
```
const legend_items = _.map(
  color_scale.ticks(5).reverse(),
  (tick, index, ticks) => ({
    label: index > 0  ? `${formatter(tick)} - ${formatter(ticks[index - 1])}` : `${formatter(tick)}+`,
    active: true,
    id: tick,
    color: get_graph_color( color_scale(tick) ),
  })
);
```

Would be a bit more involved if `color_scale.ticks(5).reverse()` returned an object because you couldn't just address `index - 1`. Multiple options to get an array from an object though (just `map` it, or use `toPairs` if the keys are important):
```
const legend_items = _.chain(color_scale.ticks(5).reverse())
  .toPairs()
  .map(
    ([key, tick], index, keys_and_ticks) => ({
      label: index > 0  ? `${formatter(tick)} - ${formatter(keys_and_ticks[index - 1][1])}` : `${formatter(tick)}+`,
      active: true,
      id: tick,
      color: get_graph_color( color_scale(tick) ),
    })
  )
  .value();
```

Stephen-ONeil: Why's this already marked `outdated`? Did I somehow review an old version?

Pretty sure it all still applies ¯\_(ツ)_/¯

Stephen-ONeil:
> Not unless we think we will have footnotes that apply to all geographic data? We don't currently.

We don't, and I don't think we will have any down the road, but it seems like a useful grouping in the RPB tag cloud? 

Ehh, maybe not. Ok, just change both tags to be more specific.

client/src/charts/canada/canada.js
ktw1016: oh shoot I didn't know all functions can return itself, that's super useful to know Thanks!

Stephen-ONeil:
Got my approval, just need that text stuff now.

ktw1016:
We just need:
French title of panel
French text of panel

cookkate:
Can you put the % share in brackets, e.g:

Recipients in British Columbia received the largest average share (45.4%) over the past five fiscal years.

cookkate:
I added links to the open datasets page for this data to the metadata page.

ktw1016:
Just curious, is the text development still ongoing @cookkate ?

Stephen-ONeil:
Oh yeah, can't even delete this branch as long as a child branch targets it. Well, don't want it deleted yet anyway, not until we're done with the dev link.

PR #358, opened 2019-11-15
Results intro panel

cookkate:
Make a new data-light results intro panel for the top of results infographics, offering definitions and an overview (maybe some PI data)


client/src/panels/panel_declarations/results/results_intro.js
```diff
@@ -0,0 +1,123 @@
...
+      const {
+        subject,
+        panel_args,
+      } = calculations;
+  
+      return (
+        <InfographicPanel
+          title={text_maker("results_intro_title")}
+          sources={sources}
+          allowOverflow
```
Stephen-ONeil: Just curious, what case is the overflow needed for?

client/src/panels/panel_declarations/results/results_intro.js
```diff
@@ -0,0 +1,123 @@
...
+        switch (level){
+          case 'dept':
+            return ResultCounts.get_dept_counts(subject.id);
+          case 'gov':
+            return ResultCounts.get_gov_counts();
+        }
+      })();
+
+      if(verbose_counts[`${latest_drr_doc_key}_total`] < 1){
+        return false;
```
Stephen-ONeil: Hmm, it would take complicating the text a bit (well, adding some conditionals to the yaml, which I know you don't like), but I think it would be good for this panel to display as long as the subject has EITHER the latest DRR or the latest DP.

I don't know any cases of it off the top of my head, but I think it would read right now in the case where a department has a DP but no DRR, as the panel will display but the DRR text will be full of 0's.

cookkate: Oh yeah good point. I think the DP and DRR text are separate keys so the conditional can be in the JS 😜

client/src/models/results.js
```diff
@@ -480,6 +480,7 @@ const drr_docs = {
     year_short: run_template('{{pa_last_year_short_second}}'),
     has_resources: true,
     could_have_previous: false,
+    doc_url: "https://www.canada.ca/en/treasury-board-secretariat/services/departmental-performance-reports/2017-18-departmental-results-reports.html",
```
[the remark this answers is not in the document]
cookkate: D'oh

client/src/panels/panel_declarations/results/results_intro.js
Stephen-ONeil: Aha, that works too!

Stephen-ONeil:
Here's my take on the government level text:

"Starting in April 2018, the public service migrated to a new results reporting structure based on the Policy on Results. This new reporting structure consists of a Departmental Results Framework and a Program Inventory for each organization.

Under this policy, organizations are made up of Core Responsibilities and their corresponding Programs. Core Responsibilities and Programs both seek to achieve certain results on a yearly or ongoing basis. Each result declares one or more indicator target by which its performance will be assessed.

Across all 2019-20 Departmental Plans, the Government is seeking to achieve 1,616 results, measured by 2,759 indicators. The assessed performance of these planned results for 0 Core Responsibilities and 0 Programs will be reported on in Fall 2020.

Across all 2017-18 Departmental Results Reports, the Government sought to achieve 1,934 results,  measured by 2,994 indicators. A results status was assigned to each indicator across the X Core Responsibilities and X Programs reported on."

Aside from the shared changes to the first two paragraphs, I think dept level text is good as-is!

client/src/panels/panel_declarations/results/results_intro.js
```diff
@@ -0,0 +1,143 @@
...
+        subject,
+        is_gov,
+        summary_result_counts,
+        doc_urls,
+      };
+    },
+    render({ calculations, sources}){
+      const {
+        subject,
+        panel_args,
```
Stephen-ONeil: This `subject` should be the same as the `subject` on `panel_args`, right? This and line `136` may be redundant. Minor, but gave me pause and I could see future developers also stopping to wonder wether or not these are supposed to be the same subjects too.

client/src/panels/panel_declarations/results/results_intro.js
```diff
@@ -0,0 +1,143 @@
...
+
+      const summary_result_counts = {
+        dp_results: verbose_counts[`${latest_dp_doc_key}_results`],
+        dp_indicators: verbose_counts[`${latest_dp_doc_key}_indicators`],
+        drr_results: verbose_counts[`${latest_drr_doc_key}_results`],
+        drr_indicators: verbose_counts[`${latest_drr_doc_key}_total`],
+        num_crs: is_gov ? 0 : verbose_counts.num_crs,
+        num_programs: is_gov ? 0 : verbose_counts.num_programs,
+        depts_with_dps: is_gov ? verbose_counts.depts_with_dps : 1,
+        depts_with_drrs: is_gov ? verbose_counts.depts_with_drrs : 1,
```
Stephen-ONeil: Could you break these cases down for me?  Is this right:
  - `depts_with_...` are 1 for non-gov because they self count? 
  - `num_crs/programs` is 0 for the government because it doesn't directly have any of either (and rolling up from departments isn't the right choice because...?)

I didn't look to heavily, but are the special cases of either necessary? If not, maybe just filter out the un-needed fields in the special case isntances.

client/src/panels/panel_declarations/results/results_intro_text.yaml
```diff
@@ -0,0 +1,77 @@
...
+
+
+reports_links_text:
+  transform: [handlebars,markdown]
+  en: |
+    The most recent {{ext_link "Departmental Plans" dp_url}}
+    and {{ext_link "Departmental Results Reports" drr_url}} can be found on canada.ca.
+  fr: |
+    Vous trouverez les plus récents {{ext_link "plans ministériels" dp_url}} et {{ext_link "rapports sur les résultats ministériels" drr_url}}
+    à canada.ca.
```
Stephen-ONeil: Maybe also clarify that the data can also be found here? Someone who doesn't know may take this as us telling them to stop wasting their time on InfoBase and follow that link.

PR #362, opened 2019-11-19
export button in rpb for simple view

### LINKED ISSUE #302: RPB should provide an export button on summary reports
Always a bit confusing that an export button is available for granular data but not summary data, especially since summary is the default (so users looking for a download might bounce without ever seeing the option).

ktw1016:
fixes #302


cookkate:
I'm not sure this is working in IE? You should rebase off of Stephen's branch to get the fixed clipboard functionality.

PR #373, opened 2019-11-21
Test service data

Stephen-ONeil:
TODO:
  - [x] some fields missing in this test data (service scope, etc.)
  - [x] snapshot tests
  - [x] before merging, disable this all behind a feature flag

This is largely a 1-to-1 representation of the denormalized Titan output. Hopefully we'll get the pipeline involved and clean things up some what before we start using the data in production (might actually want to end up somewhere in between the totally normalized and totally denormalized representations that Titan can produce).


server/src/models/services/models.js
```diff
@@ -0,0 +1,94 @@
...
+  model_singleton.define_model("ServiceStandard", ServiceStandardSchema);
+  model_singleton.define_model("Service", ServiceSchema);
+  
+  const {
+    Service,
+  } = model_singleton.models;
+
+  const loaders = {
+    services_by_org_id: create_resource_by_foreignkey_attr_dataloader(Service, 'org_id'),
+    services_by_program_id: create_resource_by_foreignkey_attr_dataloader(Service, 'program_ids'),
```
cookkate: Does this work for a list of IDs?

Stephen-ONeil: Yes (haha, took me a while to verify this because I forgot about the distinction between `.load` and `.loadMany`).

E.g. `services_by_program_id.loadMany(["TBC-BXA01", "TBC-BXA02"])` returns 
```
[
  [...services with TBC-BXA01 in their program_ids array],
  [...services with TBC-BXA02 in their program_ids array],
]
```

PR #387, opened 2019-11-28
Run client prod builds in parallel in build_all.sh

Stephen-ONeil:
TODO:
  - [x] make sure they still exit properly on error
  - [x] make sure standard error from  background processes gets through to the terminal
  - [x] maybe add a spinner
  - [x] could try taking this all the way and doing four parallel builds for the languages too. Webpack has some amount of overhead for each process and also some amount of optimization/reuse when doing multiple builds in one process. If it's faster or not is probably totally up to system resources (and it will probably put the MacBook Airs in helicopter mode). Worth a try though! 
    - four parallel builds is maybe thirty seconds slower locally but maybe a whole minute faster in CI. Gonna keep an overwhelming majority of this script's use is in CI!


Stephen-ONeil:
Only have one CI run worth of data on this, but it saved about one and a half minutes compared to the previous run! That ain't nothing.

PR #389, opened 2019-11-28
Fresh pull of pipeline data

Stephen-ONeil:
Should be good, just leaving some time to test before merging and releasing


data/glossary.csv
```diff
@@ -241,7 +218,30 @@ Chaque programme devrait avoir une méthode d’intervention principale. Toutefo
...
-"INDET_PEOPLE","Indeterminate Employee","Employé nommé pour une période indéterminée","Indeterminate / Permanent employees are persons appointed to the Public Service whose tenure in the position is of an unspecified duration.","Par « employé permanent ou nommé pour une période indéterminée », on entend toute personne nommée à un poste de la fonction publique pour une période indéfinie."
+"INDET_PEOPLE","Indeterminate Employee","Employé nommé pour une période indéterminée","""Employee Type refers to the nature of the appointment to the Public Service. There are four employee types: casual, student, term, indeterminate (including seasonal indeterminate).
```
cookkate: There's a triple quotation mark here. Is that intended/going to mess with things?

Stephen-ONeil: Oops, thought I'd fixed that in the source. I'll hotfix it before merging here and double check upstream when I'm in the office.

cookkate: :+1:

Stephen-ONeil: Oh, wait, I think pipeline didn't pull in the current glossary, this is a version before my latest corrections. I'll just drop the glossary changes from this PR and remind pipeline to make sure they grab a newer glossary source before the next run (thought maybe they always did or had automated that, but maybe not).

PR #391, opened 2019-11-29
Service Inventory prototype

cookkate:
Still a work in progress, but ready for a review at this point.

TODO:

intro:
- [x] add number of services for this dept

service fees donut:
- [x] fix $ sign in label
- [ ] table version

ID methods bar chart:
- [ ] is this the best way to display this? maybe paired bar would be better than stacked?
- [x] Legend text is based on data keys, need to process these through textmaker. Would be easier with an updated nivo 😔
- [x] y axis doesn't look good (too many numbers) in cases where there's a low number of services:
Screen Shot 2020-01-09 at 10 57 56 AM
- [ ] look at real data -- how is the data quality for this? The question may have been worded "are you collecting this, not are you using it"

communication channel bar:
- [x] rotate x axis labels
- [x] panel layout
- [ ] NAs exist
- [ ] only show columns with non-NA values
- [ ] "service transactions" or "Applications" (if it doesn't have enq & visit)

digital status heatmap:
- [x] redesign tooltip more better
- [x] panel layout
- [x] text maker for top axis labels
- [ ] fix text in table view (currently true/false/NaN, should be bilingual)
- [ ] some text: E-enablement, "not applicable" (instead of not available), look at real data, "end to end online service" something that is all true or NA, "client interaction points"
- [ ] fix order of rows between graph and table

related programs:
- [ ] make this. venn diagram? network?

client groups:
- [ ] make this

standards:
- [ ] look into standards

misc:
- [x] swap the icon being used for the services graphic for a ~headset~ id card
- [x] add a switch to show the table view
- [ ] add option to show full list of services by this metric
- [x] make sure all the text is medium size
- [ ] fix padding around the text
- [ ] right-align the swap-to-table button
- [ ] set state to table for the a11y view
- [ ] add services route to CI
- [ ] add X and Y axis labels to all graphs
- [ ] add links to data
- [ ] look at the policy on service and digital


client/src/panels/panel_declarations/services/services_channels.js
```diff
@@ -0,0 +1,164 @@
...
+    } = this.props;
+    const { selected_service_id } = this.state;
+    const colors = infobase_colors();
+
+    const data_keys = ["telephone_enquires",	"website_visits",	"online_applications",	"in_person_applications",	"mail_applications",	"other_channel_applications"];
+
+    const all_services = _.chain({})
+      .mergeWith(..._.map(panel_args.services, serv=>_.pick(serv,data_keys)),_.add)
+      .merge({name: text_maker("all"), id: "all"})
+      .value();
```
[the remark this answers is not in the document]
Stephen-ONeil: Haha, I'm basically lodash golfing now. Note: untested!
```
_.chain(panel_args.services)
  .map( service => _.pick(service, data_keys) )
  .flatMap( _.toPairs )
  .groupBy( _.first )
  .mapValues( (key_value_pairs) => _.reduce(key_value_pairs, (sum, [key, value]) => sum + value, 0) )
  .merge({name: text_maker("all"), id: "all"})
  .value();
```

client/src/panels/panel_declarations/services/services_channels.js
```diff
@@ -0,0 +1,164 @@
...
+    
+    const selected_title = selected_service.name;
+
+    const bar_data = _.chain(data_keys)
+      .map(key => ({
+        number: selected_service[key],
+        label: text_maker(`label_${key}`),
+      }) )
+      .each(obj => {obj[text_maker("number_of_enquiries")] = obj.number;})
+      .value();
```
Stephen-ONeil: Mutation,  😝. We've got a good syntax for declaring objects with variable keys these days, can write this as:
```
const bar_data = _.map(
  data_keys,
  key => ({
    number: selected_service[key],
    [text_maker("number_of_enquiries")]: selected_service[key],
    label: text_maker(`label_${key}`),
  })
);
```

client/src/panels/panel_declarations/services/services_channels.js
```diff
@@ -0,0 +1,164 @@
...
+
+    const options = _.chain([all_services])
+      .concat(panel_args.services)
+      .map(({name, id}) => ({
+        id,
+        display: name,
+      }) )
+      .value();
+    
+    const all_feedback_channels = ["Online", "Postal Mail", "Fax", "Email", "In-Person", "Telephone"];
```
Stephen-ONeil: Might be useful to put these, and the `data_keys` too, somewhere like business constants, on as properties of the Services subject? They feel like something useful to have a single source of truth for. Check me on that, I'm not as close to this stuff as you at the moment.

client/src/panels/panel_declarations/services/services_channels.js
```diff
@@ -0,0 +1,164 @@
...
+
+    const feedback_channel_statuses = _.reduce(all_feedback_channels, (result, channel) => {
+      result.push({
+        id: channel,
+        label: channel,
+        color: _.includes(selected_service.feedback_channels, channel) ? colors(0) : null,
+      });
+      return result;
+    }, []);
```
Stephen-ONeil: Unless I'm missing something, this could be a map instead?

client/src/panels/panel_declarations/services/services_digital_status.js
```diff
@@ -0,0 +1,151 @@
...
+
+const ServicesDigitalPanel = ({panel_args}) => {
+
+  const data_keys = ["account_reg_digital_status", "authentication_status", "application_digital_status", "decision_digital_status", "issuance_digital_status", "issue_res_digital_status"];
+
+  const heatmap_data = _.chain(panel_args.services)
+    .map(serv=>_.pick(serv,_.concat(["name"], data_keys)))
+    .map(serv=>_.each(serv, (value, key)=>{
+      if(value===null) serv[key] = NaN;
+      if(_.includes(data_keys, key)) serv[text_maker(key)] = serv[key];
```
Stephen-ONeil: What is this, the Witcher?

That was a mutations joke.

cookkate: Ugh yeah, this is hacky, I should remake those objects. Nivo tends to be pretty free about presenting keys which is particularly problematic for us because of the need to put all text through text_maker.

client/src/models/services.js
```diff
@@ -17,9 +17,14 @@ const Service = class Service extends static_subject_store(){
   static get type_name(){ return 'service'; }
   static get singular(){ return trivial_text_maker("service"); }
   static get plural(){ return trivial_text_maker("services"); }
+  static get_by_dept(org_id){
+    return _.filter(Service.get_all(), serv => serv.subject.id === org_id);
+  }
   get level(){ return "service"; }
   get guid(){ return `service_${this.id}`; }
 
+  get subject(){ return Dept.lookup(this.org_id); }
```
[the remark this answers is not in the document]
cookkate: That, plus consistency with the rest of the infobase that tends to use subject.

client/src/components/DisplayTable.js
```diff
@@ -114,7 +114,7 @@ export class DisplayTable extends React.Component {
                   column_keys, 
                   col => (
                     <td key={col}>
-                      {col_data[col]}
+                      {`${col_data[col]}`}
```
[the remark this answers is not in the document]
cookkate: I had some data being stored as true or false rather than strings and this was a hack to get it to show up (rather than just be blank). I'll add a TODO to change it since that will have to be changed either way to accommodate translation.

client/src/panels/panel_declarations/services/services_channels.js
```diff
@@ -0,0 +1,164 @@
...
+  }
+  
+  render(){
+    const {
+      panel_args,
+    } = this.props;
+    const { selected_service_id } = this.state;
+    const colors = infobase_colors();
+
+    const data_keys = ["telephone_enquires",	"website_visits",	"online_applications",	"in_person_applications",	"mail_applications",	"other_channel_applications"];
```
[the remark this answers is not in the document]
cookkate: fixed

client/src/panels/panel_declarations/services/services_channels.js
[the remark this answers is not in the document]
cookkate: Seems to work!

client/src/panels/panel_declarations/services/services_channels.js
cookkate: A lot of the services data has constants like these that are relevant for a single aspect of the services (in this case they're the possible values of the available feedback channels) but nowhere else. I think there's more of these than exist for the other data we've got.

Probably `client/src/models/services.js` is the right spot for them, but fair warning there's going to be a bunch...

ktw1016:
Digital Status panel:
- [ ] Minor but colour of non-digital option could be more distinct from digital option
- [ ] Solid checkboxes for legend
- [ ] Needs table view

Communication Channels panel:
- [ ] Can we get rid of dropdown, show stacked bar chart of toggle-able programs?
- [ ] Remove empty values

Identification methods panel:
- [ ] fix y-axis

ktw1016:
Closed in favor of #672

PR #395, opened 2019-12-02
Services client data

cookkate:
Client-side populate etc code for services data

### DIFF
--- a/client/src/models/populate_services.js
+++ b/client/src/models/populate_services.js
@@ -5,13 +5,10 @@ import {
   Service,
 } from './services.js';
 
-const { service_years } = Service;
 
-
-const dept_service_fragment = (years_to_load) => (
-  `${_.map(years_to_load,
-    year => `
-      services${year}: (year: ${_.toInteger(year)}){
+const dept_service_fragment = (
+  `
+      services$: services {
         service_id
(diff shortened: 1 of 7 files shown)


client/src/models/populate_services.js
```diff
@@ -0,0 +1,215 @@
...
+
+const get_all_services_query = gql`
+query($lang: String!) {
+  root(lang: $lang) {
+    orgs {
+      ${dept_service_fragment}
+    }
+  }
+}
+`;
```
Stephen-ONeil: Nitpick: these aren't get functions anymore, they directly contain the query values now.

client/src/panels/panel_declarations/finances/budget_measures/budget_measures_panel.js
```diff
@@ -33,6 +33,8 @@ const {
   CRSO,
 } = Subject;
 
+import {Service} from '../../../../models/services.js';
```
Stephen-ONeil: Drop this line before merging please.

client/src/models/populate_services.js
cookkate: yeah good point :+1:

PR #396, opened 2019-12-02
Shore up has_results logic

### LINKED ISSUE #393: Disappearing Results Panel for inactive programs
Issue: For some programs, after clicking on the Finances panel, when I click on the Results panel, it just dissapears and no content is loaded. If I just click on the Results panel without clicking on the Finances panel, the results load just fine.

This issue seems to affect a number of programs that have a last active year of 2018-19 in Titan. These  programs have `is_active = 0` in program.csv. See below for a sample of affected programs. 

Side note: These programs currently have no Finances data. However, once public accounts are tabled, we will have planned and actual spending for 2018-19.

| Deptcode| Activitycode |
| ------------- |:-------------:| 
| FEDSO | BNM00
 | INDSC | BWU00
 | NRC | BNQ00

Stephen-ONeil:
As a handy optimization, `has_results` gets prepopulated as a side-effect of queries for results counts and actual results bundles. The pre-population logic in results counts was valid (if over-stated), but the logic for the side-effect while loading a results bundle was faulty.

Basically, loading a result's bundle for a given doc_key and seeing that it has data tells you that its `has_results` should be true, but getting an empty response doesn't tell you anything (unless you're asking for every single doc_key, which we don't ever do so I won't account for that special case). The code had been incorrectly acting like an empty response confirmed that `has_data` should be false. Once `has_results` is set to false, it's never updated again.

So, any subject that didn't have all possible docs could potentially get marked as not having results erroneously (in practice, there weren't many cases where you could cause this in the app).

I've also set up `subject_has_results()` in `populate_results.js` to short circuit it's promise if it's being called for a subject that's already had it's `has_results` flag set, to stop any other ways it could be mucked with going forward (the first fix was still needed anyway, to stop it from being locked in with a potentially wrong value).

Closes #393

### DIFF
--- a/client/src/models/populate_results.js
+++ b/client/src/models/populate_results.js
@@ -25,67 +25,70 @@ query($lang: String!, $id: String) {
 `;
 const _subject_has_results = {}; // This is also populated as a side effect of api_load_results_bundle and api_load_results_counts calls
 export function subject_has_results(subject){
-  if ( !_.isNull( subject.has_data('results_data') ) ){
-    // short curicuit if already set, this isn't a status that changes durring a session
+  
+  try {
+    // throws an error if the has results_data status has yet to be loaded (in which case we load it in the catch block)
+    subject.has_data('results_data');
+    // if it didn't throw an error then we're done
     return Promise.resolve();
-  }
+  } catch(error){
+    const { id } = subject;
 
-  const { id } = subject;
+    const level = subject.level === "dept" ? "org" : subject.level;
 
-  const level = subject.level === "dept" ? "org" : subject.level;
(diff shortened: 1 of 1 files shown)


client/src/models/populate_results.js
```diff
@@ -25,6 +25,11 @@ query($lang: String!, $id: String) {
 `;
 const _subject_has_results = {}; // This is also populated as a side effect of api_load_results_bundle and api_load_results_counts calls
 export function subject_has_results(subject){
+  if ( !_.isNull( subject.has_data('results_data') ) ){
+    // short curicuit if already set, this isn't a status that changes durring a session
```
cookkate: spelling (x2)

PR #414, opened 2019-12-09
More compact mobile homepage

### LINKED ISSUE #380: Mobile: Size of bubbles on homepage
Bubbles on homepage occupy the entire screen on mobile. Therefore, the verticality and size of the options cause users to continuously scroll to view even a few options.

Stephen-ONeil:
Closes #380 

TODO:
  - [x] double check in IE, touched some CSS that was an "IE fix"
    - [x] oops, lots of IE issues. Mostly sizing, SVGs need explicit and equal width and height to display as properly-sized squares in IE, start there and see what's left to fix 
  - [x] some places where things could get even more compact if we really want
    - still true, but I think it's good right now. Can be revisited when we shuffle the search bar and such, as we expect to do soon

### DIFF
--- a/client/src/components/CardTopImage.js
+++ b/client/src/components/CardTopImage.js
@@ -12,7 +12,11 @@ const CardTopImage = ({
   text_args,
   tmf,
 }) => (
-  <a className="link-unstyled" href={link_href} target={link_open_in_new_tab ? "_blank" : "_self"} rel={link_open_in_new_tab ? "noopener noreferrer" : ""}>
+  <a 
+    className={'top-img-card-container link-unstyled'}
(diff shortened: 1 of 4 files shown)


Stephen-ONeil:
IE fixes needed, but going to leave it awaiting review for feedback on appearance in an actual browser.

cookkate:
Would suggest lowering the font size in things like the home page badges and the sub-app badges, and possibly rearranging them (ie rearranging the elements within the badges) to be more compact.

Stephen-ONeil:
Aha, argh, some of these IE issues actually exist on master. I was trying to narrow them down by combing over just this PRs changes, thinking I'd caused them here, but I hadn't.

PR #428, opened 2019-12-13
Periodic audit

Stephen-ONeil:
Nothing special here, just some standard `npm audit fix` runs.


Stephen-ONeil:
Self reviewed, self approved 👍

Issue #425, opened 2019-12-12
Post-PA 2019


Stephen-ONeil:
Some issues came up during the PA 2019 update, a lot of light was shone on some old rotten code. Had to monkey patch over some rough stuff to get it out the door, but need to deal with these larger problems ASAP.

  - [x] get the `drr_planned_actual` panel working again, strip all reference to/reliance on the DRR from it. It moves forward with public accounts, not with DRRs: #430
  - [x] sort out years for DRRs/DPs, it's a mess. ~This might turn in to re-thinking how years work, as the current relative year templates are starting to feel like a fragile kludge (in places)~ Going to try and limit the scope of that #435
  - [x] roll the tag explorer actual resources tab forward #446
  - [x] ~coordinate with the pipeline on dropping tags to old policy programs~ can sleep on this, dropping the old policy programs might be put off until after DPs

rossbm:
Not certain if this is the right place, but for all these programs were we have just one year of data, we are showing "0 year" ago on the welcome mat when it would be good to display "Last year" instead. I guess it is preferring to show whatever the latest historical is instead of the latest year.

Stephen-ONeil:
I'll look in to that too, thanks!

Stephen-ONeil:
Welcome Mat text issues uncovered by this update should be addressed by #431

PR #431, opened 2019-12-16
Welcome mat year checkup

Stephen-ONeil:
Ross noticed a weird text construction (see comment on #425) after the recent PA update. Fixing that and looking for any other flawed logic that stands out in here.


Stephen-ONeil:
Ok, bunch of stuff cleaned up and sorted out, I think. Still other bugs and problems with the Welcome Mat, but those are problems for another PR.

PR #433, opened 2019-12-17
Fix the glossary entry case of general_href_for_item

### LINKED ISSUE #432: Glossary links in home page search are busted
On the home page, if you search a term like "FTE" you get glossary results, but if you click (or hit enter) on the link it goes nowhere. Something's busted.

Stephen-ONeil:
Fixes #432

### DIFF
--- a/client/src/link_utils.js
+++ b/client/src/link_utils.js
@@ -3,11 +3,11 @@ import { rpb_link } from './rpb/rpb_link.js';
 import { Table } from './core/TableClass.js';
 import { GlossaryEntry } from './models/glossary.js';
 
-const glossary_href = (subject_or_id) => {
+const glossary_href = (subject_or_id, first_character = '#') => {
   const id = _.isString(subject_or_id) ? subject_or_id : subject_or_id.id;
   const is_valid_glossary_item = !_.isUndefined( GlossaryEntry.lookup(id) );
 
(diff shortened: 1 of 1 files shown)


client/src/link_utils.js
```diff
@@ -3,11 +3,11 @@ import { rpb_link } from './rpb/rpb_link.js';
 import { Table } from './core/TableClass.js';
 import { GlossaryEntry } from './models/glossary.js';
 
-const glossary_href = (subject_or_id) => {
+const glossary_href = (subject_or_id, first_character = '#') => {
   const id = _.isString(subject_or_id) ? subject_or_id : subject_or_id.id;
   const is_valid_glossary_item = !_.isUndefined( GlossaryEntry.lookup(id) );
 
-  return is_valid_glossary_item && `#glossary/${id}`;
+  return is_valid_glossary_item && `${first_character}glossary/${id}`;
```
cookkate: I think there might be some code duplication between link_utils.js and routes.js.

Also, it's GlossarySearch.js that originally does the switch from # to /, not sure if there's an opportunity to simplify any of this?

cookkate: Scratch that, GlossarySearch.js is downstream of this, but the format of the functions in link_utils.js and routes.js should at least be standardized. Will file an issue.

PR #438, opened 2019-12-23
Link utils standardization

### LINKED ISSUE #434: Standardize function definitions between link_utils.js and routes.js
And maybe combine them into one file?

Stephen-ONeil:
Closes #434


client/src/link_utils.js
```diff
@@ -11,19 +11,13 @@ const glossary_href = (subject_or_id, first_character = '#') => {
 };
 
 const general_href_for_item = item => {
```
cookkate: This is a super weird function. Looks like it's only used in EverythingSearch. Does that even work on tables?

Stephen-ONeil: Yup, you can search by tables from the home page. They show up in the results under the heading "Build your own report" as links to the report builder.

Stephen-ONeil: I've cleaned up the function, let me know if it feels less weird. I still think `link_utils.js` is the right place for it since it's just a smart switch wrapped around the other exports of `link_utils.js`.

cookkate: Cool 👍

Issue #437, opened 2019-12-23
contextualize footnotes with some kind of title based on topics and year


AlexCLeduc:
titan users sometimes don't understand how their explanations in titan will get used by infobase. 

Since the titan forms just say something along the lines of.. "Variance explanations for FTE and Spending" people might fill it out as "spending difference is due to factor Y".  When the infobase shows this, it's not always clear what years the footnote applies to.

Instead of a bullet-point `<ul>` maybe certain footnotes can have auto-generated headings. 

Here's a screenshot  (the text doesn't really sense) that visually shows what I had in mind:

![Screen Shot 2019-12-23 at 10 40 57 AM](https://user-images.githubusercontent.com/7769215/71366710-bd56e000-259a-11ea-9e54-4c1353d3f7a3.png)

Stephen-ONeil:
Related a bit to #44

I like that mock up's look a lot!

PR #439, opened 2019-12-23
FootnoteList rework

### LINKED ISSUE #436: footnotes should not use basic bullet-point ULs anymore
These have started breaking down since titan started collecting markdown input 

Screen Shot 2019-12-23 at 10 44 56 AM

As an easy step, I suggest we just make those `<ul>`s into `ul.fancy_ul`s

Stephen-ONeil:
Closes #436

TODO:
  - [x] don't love the margins on `<p>`s in inside the `FancyUL`,  might tweak that
  - ~should the footnote header/details behaviour be part of the standard `FootnoteList` component? I think I want it to be, for consistency between the panels and the RPB (and anywhere else we display footnotes, if I'm forgetting somewhere)~ Nah, that'd be a more brittle direction to take things for not much benefit
  - ~doesn't completely solve the problem where footnote items that were entered through Titan as single item lists look like they're broken (although it looks less wild than before). Should I try to deal with that more directly? Maybe ask the pipeline to handle them~ Looks like, at least the case that kicked this off, was not a markdown list. It was a footnote that actually contained a bullet character and a tab...

### DIFF
--- a/client/src/components/FootnoteList.js
+++ b/client/src/components/FootnoteList.js
@@ -1,0 +1,24 @@
+import './FootnoteList.scss';
+
+import { sanitized_dangerous_inner_html } from '../general_utils.js';
+
+import { FancyUL } from './misc_util_components.js';
+
+
+const FootnoteList = ({ footnotes }) => <div className={"footnote-list"}>
+  <FancyUL>
+    {_.chain(footnotes)
+      .uniqBy("text")
+      .map( 
+        ({text}, ix) => <div 
+          className={"footnote-list__note"} 
+          key={ix} 
+          dangerouslySetInnerHTML={sanitized_dangerous_inner_html(text)}
+        />
+      )
+      .value()
+    }
+  </FancyUL>
+</div>;
+
+export { FootnoteList };
--- a/client/src/components/FootnoteList.scss
+++ b/client/src/components/FootnoteList.scss
@@ -1,0 +1,11 @@
+.footnote-list {
+  margin: 0px 20px;
+}
+
+.footnote-list__note {
+  font-size: 14px;
+
+  & > p {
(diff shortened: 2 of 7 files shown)


Stephen-ONeil:
Not totally set on the style right now, particularly the panel footnote details. Feels a bit like the drawer is disconnected from the details summary when its open.

cookkate:
Yeah I'm not 100% sure about the border at all.

Just to double check, this doesn't actually touch #437 at all, right?

cookkate:
What about having the border around the entire details object? I'm not 100% sure about this, just trying to get at the "disconnected" issue.

I think the style is going to end up being more linked to the potential headers, since the issue is going to be making clear which footnotes are linked to which headers (or not). May want to hold off on tweaking the style too much right now if we're planning on doing that.

Stephen-ONeil:
> Just to double check, this doesn't actually touch #437 at all, right?

Yup, doesn't touch #437 

> What about having the border around the entire details object? I'm not 100% sure about this, just trying to get at the "disconnected" issue.

I'll give something like that a try, see how it looks.

> I think the style is going to end up being more linked to the potential headers, since the issue is going to be making clear which footnotes are linked to which headers (or not). May want to hold off on tweaking the style too much right now if we're planning on doing that.

Fair, maybe this should touch on #437...

cookkate:
I'm not sure this PR also needs to do #437 , mostly was thinking that it might make sense to hold off tweaking the style too much until that PR.

Stephen-ONeil:
Hmm, same as the rounded bubble menu items, coming off of the holidays I don't see what I disliked about how these currently are. I'll merge as-is, we'll be refining things when we tackle #437  anyway.

PR #445, opened 2020-01-03
Update dev build babel targets

Stephen-ONeil:
From hardcoded at Chrome 66 to "latest 2 Chrome versions". Less transpiling to do, so dev builds should be quicker.


Stephen-ONeil:
Self approved.

PR #446, opened 2020-01-03
Roll tag explorer forward

Stephen-ONeil:
TODO:
  - [x] ... some amount of general explorer clean up is long over-due, I think
    - Mostly just shuffled and renamed modules to make more sense of what belongs to what. Reducing boilerplate/making the code easier to reason about is out of scope here
  - [x] decouple resource explorers from results docs
  - [x] pin the tag explorer to last actual and first planned year
    - [x] roll the tag explorer actual resources tab forward (should happen in the process)
    - [x] ~redirect old routes that contain references to docs~ nah, there are probably very few bookmarks to the tag explorer, and half of them would be technically invalidated either way
  - [x] remove the sub-program note on the tag explorer page


client/src/panels/panel_declarations/finances/drr_dp_resources/sub_program_resources.js
```diff
@@ -154,6 +152,54 @@ const get_non_col_renderer = ({doc}) => ({node}) => {
   );
 };
 
+const get_col_defs = ({doc}) => [
```
Stephen-ONeil: Commented on this in the relevant commit message, but this is a copy of the old `get_col_defs` from resource_explorer_common.js. I'm fine with a dumb copy paste of it here because this module is slated for deletion soon enough anyway,

cookkate: :+1:

Stephen-ONeil:
> The values for the actual crown corp FTEs are blank, but the sum shows zero. Wonder if there's some way of blanking that out as well?

I'm sure there is, not sure why they currently have different behaviour. Open an issue?

PR #449, opened 2020-01-08
do not show spend or fte row if all are empty or 0 for welcome mat

### LINKED ISSUE #416: Programs missing FTEs have blank graphs
eg https://www.tbs-sct.gc.ca/ems-sgd/edb-bdd/index-eng.html#orgs/program/PCA-ISS03/infograph/financial

Probably should just not show the whole FTE row in this case.

ktw1016:
fixes #416

### DIFF
--- a/client/src/panels/panel_declarations/finances/welcome_mat/welcome_mat.js
+++ b/client/src/panels/panel_declarations/finances/welcome_mat/welcome_mat.js
@@ -160,6 +160,8 @@ const WelcomeMat = (props) => {
   const {
     latest_hist_spend_data,
     oldest_hist_spend_data,
+    is_spend_empty,
+    is_fte_empty,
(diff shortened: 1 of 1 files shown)


ktw1016:
Oops forgot to put the label on

Stephen-ONeil:
Part of my reason for suggesting `has_spending` over `is_spending_empty` is that eyes lock on to the first and last word when skimming, so when you have two multi-word variables that only differ by one word it's nicer to get that word at the start or end. Not that I always remember to do that my self!

Extra pedantic (and this is something I DEFINITELY get sloppy about sometimes), but `is_something_empty` seems like it would be the name of a function or maybe a getter. `something_is_empty` sounds more like a boolean variable. You can ignore this point if you disagree, just something to keep in mind though.

PR #454, opened 2020-01-10
make FTEs and other integer columns use floats

AlexCLeduc:
Just to give an idea of the scale of this fix, Gov-wide FTEs in 2014-15 went from 339,589 to 339,632

This didn't change the planned FTE counts, I guess no one plans using fractional FTEs


AlexCLeduc:
If you want, we should tell Mike too

Stephen-ONeil:
Done, merge away.

PR #458, opened 2020-01-14
collapsed panels

### LINKED ISSUE #401: Stay-collapsed panels?
Would be nice if users could collapse panels they're not interested in and leave them closed. Would help make the longer infographics (e.g. dept financial) a little more useful.

ktw1016:
fixes #401 

Just first initial to let everyone check it out

### DIFF
--- a/client/src/components/Panel.js
+++ b/client/src/components/Panel.js
@@ -64,35 +64,70 @@ const PanelGlossary = ({keys}) => {
   );
 };
 
-export const Panel = ({allowOverflow, title, otherHeaderContent, children, sources, glossary_keys, footnotes }) => (
-  <section className={classNames('panel panel-info mrgn-bttm-md', allowOverflow && "panel-overflow")}>
-    { (title || otherHeaderContent) &&
-      <header className='panel-heading'>
-        { title && <h3 className="panel-title"> {title} </h3> }
-        { otherHeaderContent }
-      </header>
-    }
-    <div className='panel-body'>
-      { children }
-      <div className="mrgn-tp-md" />
-      { _.nonEmpty(sources) && 
-        <div>
-          <PanelSource links={sources} />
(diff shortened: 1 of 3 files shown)


cookkate:
Code looks fine. Comment about functionality though--you should hide the buttons at the top right, both because it doesn't make sense to have them functional and because the PDF download button errors if you click it on a closed panel :)

ktw1016:
Ohh yup true!

cookkate:
Also this is nitpicky but can you align the triangle/arrow to the text? It looks like the text isn't centred vertically on the header but the triangle is.

cookkate:
What's wrong with this applying to datasets panels? Panel is generic, but not that generic. It also has all the footer stuff like glossary keys and footnotes in it, which only apply to the infographics.

ktw1016:
@Stephen-ONeil are there any reasons why datasets shouldn't be collapsed?

Stephen-ONeil:
I thought the intention of this PR was just to add collapsing to infographic panels. It isn't necessarily a bad feature in either place but, in my books, if it isn't intentional then it is a bug.

I don't think it is necessary on the dataset page. Maybe useful on the text diff. Would still prefer it as a panel util for design purposes\*. Like I said though, I can be out-voted on this for now if y'all want to rush in to production. 

\* design thoughts:
  - all the other general panel interactivity is grouped on the right, introducing something elsewhere breaks that convention. I'd prefer not to
  - I know it is supposed to invoke a details drawer but, overall, I don't think it looks enough like our other details drawers to pull that off. The arrow before the title might just look like an attempt at a stylistic bullet point?

@cookkate good point, leaving footnotes and glossary items in the  general Panel component was a mistake :p

ktw1016:
Yup no problem, I'll merge this in for now then!

PR #463, opened 2020-01-20
added search filter within DIsplayTable

### LINKED ISSUE #261: DisplayTable should be filterable
Columns are optionally sortable but there's no functionality for filtering by any means (selecting an option or free-text entry).

ktw1016:
fixes #261 

- [ ] Possibly a search icon that toggles showing the input field?? A lot of people would miss that there is a search if this happens
- [ ] Highlighting a searched query would also be cool

Let me know any feedback

### DIFF
--- a/client/src/components/DisplayTable.js
+++ b/client/src/components/DisplayTable.js
@@ -1,5 +1,5 @@
 import './DisplayTable.scss';
-import text from '../common_text/result_lang.yaml';
+import text from '../common_text/common_lang.yaml';
 import { create_text_maker_component } from './misc_util_components.js';
 
 import classNames from 'classnames';
@@ -11,11 +11,14 @@ const { text_maker, TM } = create_text_maker_component(text);
 export class DisplayTable extends React.Component {
   constructor(props){
     super();
+    const {
+      sort_keys,
+      col_search,
+    } = props;
     this.state = {
-      sort_by: "label",
+      sort_by: sort_keys[0],
       descending: true,
-      indicator_query: "",
-      activity_query: "",
+      col_search: col_search,
     };
   }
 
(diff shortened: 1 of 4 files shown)


client/src/components/DisplayTable.js
```diff
@@ -64,12 +69,24 @@ export class DisplayTable extends React.Component {
...
                 className="center-text"
-                onClick={ () => this.header_click("label") }
               >
                 {label_col_header || ""}
-                <SortDirections 
-                  asc={!descending && sort_by === "label"} 
-                  desc={descending && sort_by === "label"}
+                <div
+                  onClick={ () => this.header_click("label") }
+                >
```
cookkate: nitpick, but I would put this on one line if theres only one attribute

client/src/components/DisplayTable.js
```diff
@@ -64,12 +69,24 @@ export class DisplayTable extends React.Component {
...
+                    desc={descending && sort_by === "label"}
+                  />
+                </div>
+                <input
+                  type="text"
+                  style={{width: "100%"}}
+                  value={activity_query}
+                  onChange={evt => 
+                    this.setState({ activity_query: evt.target.value })
+                  }
```
cookkate: I think it makes sense to assume the first column of a table is always filterable, but it might not always be activities (could be a table of departments). Can you rename this query to something like 'id_query' or 'first_column_query'?

client/src/components/DisplayTable.js
```diff
@@ -64,12 +69,24 @@ export class DisplayTable extends React.Component {
...
+                  placeholder={text_maker('filter_results')}
```
cookkate: Should genericize this ("Filter data" maybe)

client/src/components/DisplayTable.js
```diff
@@ -79,13 +96,27 @@ export class DisplayTable extends React.Component {
...
-                        />
+                        <div
+                          onClick={ () => _.includes(sort_keys,tick) && this.header_click(tick) }
+                        >
+                          <SortDirections 
+                            asc={!descending && sort_by === tick} 
+                            desc={descending && sort_by === tick}
+                          />
+                        </div>
+                        { tick === "indicator" &&
```
cookkate: Hardcoding this in DisplayTable.js means we can't use DisplayTables for other data! Should probably pass in a list of filterable keys like sort_key

ktw1016:
Good point with generalizing, I meant to do it initially but first wanted to see how people thought about it. DisplayTable should be pretty isolated now

ktw1016:
Hmm for highlighting, I made a separate issue here: #479

PR #464, opened 2020-01-20
fixing labeled table's "labeled-table-item"

### LINKED ISSUE #461: Principles CSS bug
Screen Shot 2020-01-17 at 10 37 53 AM

bill-lliu:
something was misnamed when being updated
fixes #461

### DIFF
--- a/client/src/components/LabeledTable.js
+++ b/client/src/components/LabeledTable.js
@@ -24,18 +24,18 @@ export class LabeledTable extends React.Component {
                 aria-checked={item.active}
                 key={item.name}
               >
-                <div className="labeled-table__item__header">
+                <div className="labeled-table__item-label">
                   { styles && styles.header ? <span className={styles.header}>{item.name}</span> : item.name }
                 </div>
-                <div className="labeled-table__item__description">
+                <div className="labeled-table__item-description">
                   { styles && styles.desc ? <span className={styles.desc}>{item.desc}</span> : item.desc }
                 </div>
               </a> :
               <div className="labeled-table__item" key={item.name}>
-                <div className="labeled-table__item__header">
+                <div className="labeled-table__item-label">
                   { styles && styles.header ? <span className={styles.header}>{item.name}</span> : item.name }
                 </div>
-                <div className="labeled-table__item__description">
+                <div className="labeled-table__item-description">
                   { styles && styles.desc ? <span className={styles.desc}>{item.desc}</span> : item.desc }
                 </div>
               </div>
--- a/client/src/components/LabeledTable.js
+++ b/client/src/components/LabeledTable.js
@@ -13,7 +13,7 @@ export class LabeledTable extends React.Component {
         <div className="labeled-table__header">
           {title}
         </div>
-        <div className="labeled-table__items">
+        <div className="labeled-table-items">
           {_.map(content, item =>
           item.href ?
               <a href={item.href}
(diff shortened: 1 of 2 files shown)


client/src/components/LabeledTable.js
```diff
@@ -31,7 +31,7 @@ export class LabeledTable extends React.Component {
                   { styles && styles.desc ? <span className={styles.desc}>item.desc</span> : item.desc }
```
ktw1016: Can you just wrap `item.desc` to `{item.desc}`? this seems like 'potential' bug. Same with line 39

Stephen-ONeil:
On a side note, the diff is hard to read on this PR. Not your fault, diffs just don't do a great job with things like large blocks of indentation changing. Specifically, the changes in `LabeledTable.scss` are really hard to follow from the diff.

I gather the only thing you changed in there was fixing the bad indentation? Good catch by the way! If you changed any styles at the same time though, from this one commit alone, it's like finding a needle in a haystack for me to find the change!

One general tip for making commits is to try and do only one "thing" per commit. The commit itself can be as big as it needs to be, and how you scope a "thing" is up to you. In this case, if "Fixed indentation in LabeledTable.scss" was it's own commit, then when reviewing I could rely on looking at diffs from individual commits/trusting the commit messages rather than trying to work out what happened from one big diff blob.

penguinofwinter:
Thanks for the explanation, it did clear things up a little! Kate also showed me to the BEM naming standards (http://getbem.com/naming/) that I read over. 

I've committed a renaming of the "labeled-table-items" to be "labeled-table__items", which is what I believe you intended. 

There was, however, yet another class that is used around called "labeled-table-item" (item is singular). This is then used to name two other sub-elements called "labeled-table-item__header" and "labeled-table-item_description". I've tentatively changed this too, in the second renaming commit, but I wasn't sure if you meant for the "item" part of the name to be a part of the Block or the Element of BEM. We can revert this commit separately if you intended the former. 

I've also committed Taewan's "item-desc" precaution.

Stephen-ONeil:
**NOTE:** as you can tell by the class names I picked that caused this bug in the first place, I usually don't spend as much time picking class names as I'm about to in this post. I'm only digging in so deep here because I feel like it's a good chance to explain, and maybe rethink, how I've been using BEM. If this makes sense to everyone, I'll write it up in our wiki as the new project guideline on BEM usage.

Good eye, catching the singular version of that class name mixed in there! I'd overlooked/forgotten about that. In my understanding/application of BEM, you don't write things like `block__element__child-element` though, feels like it would get unwieldily and be harder to maintain (e.g. if a new middle element was added to the later on then you'd have to update all the classes under it too).

My approach, instead, has usually been to think of lower elements that also have children as new block contexts, so I thought of `labeled-table__items` as an element within `labeled-table`, but I thought of the `labeled-table-item`s inside it as their own blocks.

Gave it a bit of a search just now, and it looks like some people advocate an alternate way of applying BEM, where you ignore depth/nesting in the names of elements, and just give them names relative to the highest parent block of the component you're building. So, the structure would be something like this:
```
<div className="labeled-table">
  <div className="labeled-table__header" />
  <div className="labeled-table__items">
    <a className="labeled-table__item" />
      <div className="labeled-table__item-header" />
      <div className="labeled-table__item-description" />
    </a> 
  </div>
</div>
```

I think that's better! Shorter names and less coupling of DOM structure to class names so likely easier to modify later. This example's not perfect though because there's still two classes differing only by one character, and also two types of "header" inside the `labeled-table`.

In this case `labeled-table__item` is sort of a redundant class name anyway, since it could be assumed that every direct child of `labeled-table__items` is a "item", so we could drop the class on that component and just select the "items" for styling with `labeled-table__items > *`.

I guess we could even solve the multiple headers issue by changing `labeled-table__item-header` to `labeled-table__item-label`.

Stephen-ONeil:
I'll re-review after lunch!

ktw1016:
Didn't know about BEM stuff, good to know!

PR #490, opened 2020-02-03
Change how nivo tooltips are displayed for mobile devices

### LINKED ISSUE #155: Improving Nivo tooltips on mobile
Some tooltips spill out over the edge of the screen on mobile, even though Nivo is supposed to handle that (and mostly does)

Capture d’écran 2019-08-19 à 09 42 03

cookkate:
Fixes #155

We're already wrapping the nivo tooltips with our own default (to include custom formatters etc). I made a modification to make the font size a little smaller on small devices, and, when that wasn't good enough, to rearrange how tooltip values are displayed from a horizontal to vertical arrangement. Some screenshots to illustrate:

before: Screen Shot 2020-02-03 at 11 45 23 AM

after: Screen Shot 2020-02-03 at 11 45 28 AM

before: Screen Shot 2020-02-03 at 11 43 52 AM

after: Screen Shot 2020-02-03 at 11 43 59 AM

Note that some graphs use custom tooltips, this doesn't affect them: Screen Shot 2020-02-03 at 11 53 13 AM


client/src/charts/NivoCharts.js
```diff
@@ -40,18 +41,28 @@ const get_scale_bounds = (stacked, raw_data, zoomed) => {
...
+                    <div style={{padding: '3px 5px'}}> {tooltip_item.id} </div>
+                    <div style={{padding: '3px 5px'}} dangerouslySetInnerHTML={{__html: formatter(tooltip_item.value)}} />
+                  </td>
+                  :
+                  <Fragment>
+                    <td style={{padding: '3px 5px'}}> {tooltip_item.id} </td>
+                    <td style={{padding: '3px 5px'}} dangerouslySetInnerHTML={{__html: formatter(tooltip_item.value)}} />
+                  </Fragment>
+              }
             </tr>
```
Stephen-ONeil: In practice, the approach of using conditionals on `window.innerWidth` at render time is probably fine in this case, since the popups are fairly transitory and screen resizes are uncommon\*. In general, the problem with this approach is that there's nothing to make this re-render in response to screen size changing.

\* there is one case where you could have a popup active across a screen resize, and it's when you tap to open a popup on a mobile device and then rotate the device while the popup is still open. Probably relevant case to consider here.

Suggestion 1: move the popup styling in to css so you can use a media query for the condition currently on line 47.

Suggestion 2: use the `MediaQuery` react component in the place of the check on line 55, it'll handle re-rendering it's children on screen resize.

client/src/charts/NivoCharts.js
```diff
@@ -42,58 +43,52 @@ const get_scale_bounds = (stacked, raw_data, zoomed) => {
...
-                    <div style = {{padding: '3px 5px'}} dangerouslySetInnerHTML = {{__html: formats.percentage1(tooltip_item.value/total)}}/>
-                  </td>
-                  :
-                  <Fragment>
-                    <td style = {{padding: '3px 5px'}}>{tooltip_item.id}</td>
-                    <td style = {{padding: '3px 5px'}} dangerouslySetInnerHTML = {{__html: formatter(tooltip_item.value)}}/>
-                    <td style = {{padding: '3px 5px'}} dangerouslySetInnerHTML = {{__html: formats.percentage1(tooltip_item.value/total)}}/>
-                  </Fragment>
-              }
+              <MediaQuery minDeviceWidth={breakpoints.minSmallDevice}> {/* Have I gotten min and max mixed up here?? */}
```
Stephen-ONeil: RE: comment, nope, got it right! In media queries, min means actual => min, so it's the right choice for selecting cases where the screen is larger than "small". As for the breakpoint, `minSmallDevice` is the correct option for a `minDeviceWidth` check.

PR #491, opened 2020-02-03
Split up state and props for column search of display tables

cookkate:
Using the prop as the state was causing this to break for other uses of DisplayTable, so I split it to function similarly to how the sort keys works.

### DIFF
--- a/client/src/components/DisplayTable.js
+++ b/client/src/components/DisplayTable.js
@@ -100,7 +100,7 @@ export class DisplayTable extends React.Component {
                             onChange={evt => {
                               const new_query = _.chain(col_search_keys)
                                 .keyBy()
-                                .mapValues((key)=>col_search[key] || "")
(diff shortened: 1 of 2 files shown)


client/src/components/DisplayTable.js
```diff
@@ -88,13 +92,17 @@ export class DisplayTable extends React.Component {
...
+                        { _.includes(col_search_keys, tick) &&
                           <input
                             type="text"
                             id={tick}
                             style={{ width: "100%", fontWeight: "normal" }}
                             onChange={evt => {
-                              const new_query = _.clone(col_search);
+                              const new_query = _.chain(col_search_keys)
+                                .keyBy()
+                                .mapValues((key)=>col_search[key] || "")
```
ktw1016: I do wish this was spaced a little bit better for readability `.mapValues( (key) => col_search[key] || "" )

cookkate: :+1:

PR #494, opened 2020-02-04
Periodic audit

Stephen-ONeil:
SHOULD be routine, but something's off in CI after these package upgrades. 

TODO:
  - [x] add `make` to the CI build image


Stephen-ONeil:
Ok, at first glance, looks like the latest node-sass libraries failed to install in CI because `make` is missing from our CI images. Pretty surprising is that's accurate though.

PR #498, opened 2020-02-07
working fix - pie chart data clarification

### LINKED ISSUE #457: Average share for Population by Employee Type is confusing
The doughnut chart and accompanying table for Population by Employee Type is a bit confusing since it shows average share across the 5 years rather than the latest year. I was about to use the Report a Problem widget to report that the 3.4% from the top text does not match the chart/table's number of 3.2%

Suggest using latest year for the chart and table. Alternatively, better labeling. Ideally both?

image

bill-lliu:
After learning that there are different kinds of Pie Charts used, these few lines should do the job for only the Pie Charts that average data. 

fixes #457

### DIFF
--- a/client/src/panels/panel_declarations/people/AverageSharePie.js
+++ b/client/src/panels/panel_declarations/people/AverageSharePie.js
@@ -10,6 +10,8 @@ import {
   util_components,
 } from "../shared.js"; 
 
+import { LabeledBox } from '../../../components/index.js';
+
 const { TabularPercentLegend } = declarative_charts;
 const { Format } = util_components;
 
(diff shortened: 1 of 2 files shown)


client/src/panels/panel_declarations/people/AverageSharePie.js
```diff
@@ -70,6 +72,9 @@ export const AverageSharePie = ({panel_args, sort_func}) => {
                 </span>
               }
             />
+            <div className="centerer">
+              <p><i>*Chart data is averaged over the period listed</i></p>
```
Stephen-ONeil: Content has to be bilingual. A phrase like this should be stored in yaml and make use of our `text_maker` ecosystem too. Let me know if you haven't been shown that yet!

Stephen-ONeil:
Before you close the related issue, ask Ross if this change addresses the problem he was highlighting.

PR #500, opened 2020-02-11
Panel table view

cookkate:
Implemented something I was playing with in the services branch. Since the graphs accept data in a standard way it should in theory be possible to also display a table version of that graph automatically.

Currently I just implemented it on the (vertical) bar graphs, and the results are decent although not perfect, e.g. the super simple bar graphs look a little weird in table form

Open for comments / suggestions!

TODO:

- [x] figure out what's going on with the legend mutation pattern on all charts
- ~~rewrite get_formatter to allow for percent as well as money / int~~ not sure why I decided this was important, the smart percent is better.
- [x] fix padding on auth_exp_planned_spending- [x] split up NivoCharts.js
- clean up/organize component parameters
   - [x] pie
   - [x] bar
   - [x] line
- [x] convert commonDonut to NivoResponsivePie, delete commonDonut
- [x] Broken panels:
  - [x] gap-year key is showing up on auth_exp_planned_spending

- API/UI
  - [x] switch button to icon
  - [x] move table button and zoom button into same wrapper
  - [x] ~align the buttons a little better~ (looks like there was the ability for some customization of the position on the zoom button--is this breaking those charts?) -- turned out to be non-issue
  - [x] styling on buttons - colour them blue instead of grey, use sibling selectors on the borders/corners for the case where there are two buttons
  - [x] allow graphs to turn off table button
  - [x] settle on 'default on' for both zoom button and show table button
  - [x] move graph height into a prop on all the nivo graphs
- wait on the displayTable api change in the drr branch
  - [x] deal with 'label' header
  - [x] reduce font size in table 
  ~~headers are centred, values are left justified--looks pretty weird.~~ denied, it looks worse to left-justify everything.
  - [x] fix table name TODO
- format values
  - [x] format the table data, default to $
  - [x] change the format for non-$ graphs (people etc)
- ~~The bootstrap docs say: "Be sure to add role="dialog" and aria-labelledby="...", referencing the modal title, to .modal, and role="document" to the .modal-dialog itself. Additionally, you may give a description of your modal dialog with aria-describedby on .modal." Make sure we're doing this~~ semi-done, rest depends on the long-awaited update to bootstrap 4 :(
- [ ] graphs to add:
  - [x] bar
  - [x] hbar
  - [x] line
  - [x] CommonDonut
~~LineBarToggleGraph~~ never mind, get this one for free
~~liquid fill gauge?~~ to be replaced in #530
  - [ ] Canada chart?
  - [x] GOCO?
  - [ ] flat treemap? Another candidate to switch to nivo


bill-lliu:
It kinda seems like an upgraded version of the a11y table; 

Could this possibly be considered to replace the a11y tables? Things like the Sort-By feature at the top of the tables may help for super large tables such as Program Spendings

bill-lliu:
screenshot-localhost_8080-2020 02 14-15_41_26

Also see what you mean now by the button being weird; it just needs to be spaced out a little more. This also happened when I tried to add more rows from the side legend menu.

client/src/charts/NivoCharts.scss
```diff
@@ -7,4 +7,13 @@
...
+}
+
+.btn-table-view { 
+  background-color: $backgroundColor;
+  border: 2px solid $secondaryColor;
+  box-shadow: none;
+  -webkit-box-shadow: none;
+  color: $secondaryColor !important;
+  border-radius: $standardBorderRadius;
+}
```
Stephen-ONeil: Does this need a new class, or even a non-standard appearance? I think it should either be an icon or look the same as any other button.

client/src/charts/NivoCharts.js
```diff
@@ -94,6 +112,37 @@ const fixedSymbolShape = ({
...
+      table,
+    } = this.props;
+
+    return (
+      <Fragment>
+        { show_table ? table : graph }
+        <TableSwitchButton
+          showing_table={show_table}
+          onClick={() => this.setState({ show_table: !show_table }) }
+        />
```
Stephen-ONeil: `TableSwitchButton` is plain enough, and specific enough as a sub-component of `TableSwitchableGraph`, that I don't think it needs to be standalone. Might as well in-line it here to reduce the boilerplate and scrolling required in reading this code.

I'd recommend placing the button above the graph/table. Two reasons:
  - height is almost certain to change after it's clicked, so if it's below the graph then it's going to be pushed out from under the mouse. Always a bit disorienting. The text also changes so it might be harder to track it to it's new position. Especially bad if the user immediately wants to switch back to the previous state
  - I believe it's a general a11y practice to place controls above the DOM they modify, or at least clarify (in text or label) their position relative to the content they control. I know this is not likely to be turned on in the text-only version but, for screen reader users who stay on the standard site, this would be a valuable button to do properly.

client/src/charts/NivoCharts.scss
cookkate: Yeah I'm still not 100% sure how this should look. I thought an icon would be fun (and relatively unobtrusive) but didn't want to put in the work (yet)

ktw1016:
RIP `liquidfillgauge`

client/src/charts/nivo/NivoCharts.js
```diff
@@ -0,0 +1,1037 @@
...
+      borderWidth,
+      disable_table_view,
+    } = this.props;
+
+    legends && (legends[0].symbolShape = fixedSymbolShape);
+
+    const table_data = bar_table_data(data, indexBy, get_formatter(is_money, text_formatter, true, true));
+    const table_header_keys = _.concat([indexBy],keys);
+
+    const table = !disable_table_view && <DisplayTable data={table_data} column_keys={table_header_keys} sort_keys={table_header_keys} table_data_headers={table_header_keys} table_name={"TODO"}/>;
```
ktw1016: I think you should break this up! Same for all below

client/src/charts/nivo/NivoCharts.js
```diff
@@ -0,0 +1,1037 @@
...
+      label: row.id}));
+    const table_header_keys = _.concat(['label'], _.chain(data)
+      .map( d=>_.map( d.data, d=>d.x ) )
+      .flatten()
+      .uniq()
+      .value() );
+    
+    const table = !disable_table_view && <DisplayTable data={table_data} column_keys={table_header_keys} sort_keys={table_header_keys} table_data_headers={table_header_keys} table_name={"TODO"}/>;
+
+    const zoom_button = (!disable_yaxis_zoom && !enableArea) ?
```
ktw1016: Could this still work written `(!disable_yaxis_zoom && !enableArea) &&` ?

client/src/core/format.js
```diff
@@ -232,10 +232,8 @@ _.each(
   }
 );
 
-const dollar_formats = _.pickBy( formats, (format, key) => /^compact([0-9?]?)+_raw$/.test(key) );
 export { 
   formatter, 
   list_formatter, 
   formats,
-  dollar_formats,
```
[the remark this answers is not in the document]
cookkate: I deleted it. It seems to be redundant, most places that were importing it were importing both formats (which includes the dollar formats) and dollar formats.

client/src/extended_bootstrap_css/bootstrap-wet-fixes-extensions.scss
```diff
@@ -304,6 +304,22 @@ a.btn-ib-light {
   color: $backgroundColor !important;
 }
 
+.btn-ib-array {
```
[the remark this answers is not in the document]
cookkate: I could have but it seemed to me to be a more general "button array" with this styling:

Screen Shot 2020-02-27 at 10 24 24 AM

So, currently just used for the graphs but meant as a more general style.

client/src/charts/nivo/NivoCharts.js
```diff
@@ -0,0 +1,1037 @@
...
+      0 
+    );
+
+    const table_data = _.chain(data)
+      .map( row => _.assign(row, {percentage: formats.percentage_raw(row.value/legend_total), formatted_value: get_formatter(is_money, text_formatter, true, true)(row.value)}) )
+      .map( row => ({col_data: row, label: row["label"]}) )
+      .value();
+    const table_header_keys = ["label", "formatted_value", "percentage"];
+
+    const table = !disable_table_view && <DisplayTable data={table_data} column_keys={table_header_keys} sort_keys={table_header_keys} table_data_headers={table_header_keys} table_name={"TODO"}/>;
```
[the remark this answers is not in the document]
cookkate: The DisplayTable API has changed as of the DRR update so this will change

client/src/charts/nivo/NivoCharts.js
cookkate: Sorry, I'm not sure what you mean here

client/src/charts/nivo/NivoCharts.js
cookkate: Which line? As I said above line 431 is going to change because the DisplayTable API changed.

ktw1016: ohhh okay didn't see that

client/src/charts/nivo/NivoCharts.js
ktw1016: I just thought it would look cleaner short circuiting with && instead of ternary here like this:
```
    const zoom_button = (!disable_yaxis_zoom && !enableArea) &&
      <button
         ...
      </button>;
```
but not sure if it would break because `zoom_button` value would be `false` instead of `undefined`

ktw1016:
I'm guessing it's known that GOCO is not working for this?

cookkate:
Yeah, GOCO's one of the one off graphs that isn't working. Up in the air whether I try to get it to work or just make sure that it's disabled--I forget if it's attempting to show a button right now or not.

ktw1016:
Okay, I'm familiar with GOCO code so if you're busy, I can try to see what's up with GOCO for this. 

Also was going to mention that the tabled data are not formatted. E.g.: 
$310,141,662,006.00 for Authorities on Auth, Exp, Planned spending panel.

client/src/panels/panel_declarations/finances/budget_measures/budget_measures_panel.js
```diff
@@ -876,18 +876,16 @@ class BudgetMeasureHBars extends React.Component {
             </div>
           </div>
           <MediaQuery minWidth={992}>
-            <div className="centerer" style={{height: `${data.length*30 + 150}px`}}>
```
ktw1016: Hmm, so I noticed `centerer` class is gone, that doesn't affect anything?

client/src/charts/nivo/NivoCharts.js
```diff
@@ -694,21 +695,27 @@ export class NivoResponsiveLine extends React.Component {
...
       .flatten()
       .uniq()
-      .value() );
-    const column_names = _.fromPairs(_.zip(ordered_column_keys,ordered_column_keys)); //TODO: fix this
-    
+      .value();
+    const ordered_column_keys = _.concat(['label'], last_column_keys);
+  
+    const column_names = _.chain(ordered_column_keys)
+      .zip( _.concat([table_first_column_name ? table_first_column_name : text_maker("label")], last_column_keys) )
```
ktw1016: Super minor nit pick on my personal preference to short circuiting:
`_.concat([table_first_column_name || text_maker("label")], last_column_keys)`

cookkate:
> Also was going to mention that the tabled data are not formatted. E.g.:
$310,141,662,006.00 for Authorities on Auth, Exp, Planned spending panel.

This is by design. It is formatted ($ sign and showing down to the cents) to show the most detailed data for someone who wants to copy and paste into excel.

client/src/charts/nivo/NivoCharts.js
```diff
@@ -421,17 +445,20 @@ export class NivoResponsiveBar extends React.Component{
...
 
     legends && (legends[0].symbolShape = fixedSymbolShape);
 
     const table_data = bar_table_data(data, indexBy, get_formatter(is_money, text_formatter, true, true));
-    const table_header_keys = _.concat([indexBy],keys);
-
-    const table = !disable_table_view && <DisplayTable data={table_data} column_keys={table_header_keys} sort_keys={table_header_keys} table_data_headers={table_header_keys} table_name={"TODO"}/>;
+    const ordered_column_keys = _.concat([indexBy],keys);
+    const column_names = _.chain(ordered_column_keys)
+      .zip( _.concat([table_first_column_name ? table_first_column_name : text_maker("label")], keys) )
```
ktw1016: I think short-circuiting could also apply here:
`_.concat([table_first_column_name || text_maker("label")], keys)`

client/src/panels/panel_declarations/finances/budget_measures/budget_measures_panel.js
cookkate: nope

client/src/charts/nivo/NivoCharts.js
cookkate: :+1:

Stephen-ONeil:
Gov level gender panel's graph is missing

client/src/charts/nivo/nivo_bar.js
```diff
@@ -0,0 +1,273 @@
...
+} from '../../components/index.js';
+
+
+const bar_table = (data, keys, indexBy, table_view_format, table_name, table_first_column_name) => {
+  const table_data = _.map(data, row => ({
+    display_values: _.chain(row)
+      .toPairs()
+      .map(r => r[0]===indexBy ? [indexBy, r[1]] : [r[0],table_view_format(r[1])] )
+      .fromPairs()
+      .value(),
```
Stephen-ONeil: Could be
```
display_values: _.mapValues(
  row,
  (values, key) => key === indexBy ? values : table_view_format(values)
),
```

client/src/charts/nivo/nivo_bar.js
```diff
@@ -0,0 +1,273 @@
...
+      _.map(legends,
+        legend => _.chain(legend)
+          .clone()
+          .assign({symbolShape: fixed_symbol_shape})
+          .value() )
+      ) :
+      undefined;
+
+    const table = !disable_table_view && bar_table(data, keys, indexBy, get_formatter(is_money, text_formatter, true, true), table_name, table_first_column_name);
+
```
Stephen-ONeil: Is this the same code from lines 81-92? Please de-dupe if so, so the logic doesn't become inconsistent later by accident.

client/src/charts/nivo/nivo_line.js
```diff
@@ -0,0 +1,222 @@
...
+    } = this.state;
+
+    const IE_fixed_legends = legends ? (
+      _.map(legends,
+        legend => _.chain(legend)
+          .clone()
+          .assign({symbolShape: fixed_symbol_shape})
+          .value() )
+      ) :
+      undefined;
```
Stephen-ONeil: Oh, here too, guess this should be pulled back to `../nivo_shared.js`

client/src/charts/nivo/nivo_shared.js
```diff
@@ -0,0 +1,195 @@
...
+        <StatelessModal
+          show={ show_table }
+          title={ table_name || graph_text_maker("default_table_name") }
+          body={ table }
+          on_close_callback={() => this.setState({show_table: false})}
+          additional_dialog_class = { (() => {
+            if (window.innerWidth < breakpoints.minMediumDevice) { return 'modal-sm'; }
+            if (window.innerWidth > breakpoints.maxLargeDevice ) { return 'modal-xl'; }
+            return 'modal-lg';
+          })() }
```
Stephen-ONeil: This isn't going to respond to window resizes. Probably fine since since it will be re-called every time the modal's open/closed, but it would be preferable if it was automatic (I could see someone clicking the button on their phone, then deciding to rotate to try and see more of the table, which might not work great as-is).

You don't have to work that out now, but if you don't then please open an issue about it when this PR merges.

client/src/components/modals_and_popovers/bootstrap_modal_exstension.scss
```diff
@@ -36,4 +36,15 @@ $margin: 30px;
...
+
+.modal-sm {
+  width: 300px;
+}
+.modal-lg {
+  width: 800px;
+}
+.modal-xl {
+  width: 1140px;
 }
```
Stephen-ONeil: Oh, these are ours, that makes it easier to add the responsiveness. Maybe a third class `modal-responsive` the appropriate media queries build in?

client/src/panels/panel_declarations/finances/goco/goco.js
```diff
@@ -277,9 +331,11 @@ class Goco extends React.Component {
             <NivoResponsiveBar
               { ...nivo_default_props }
               data={ node.data.children }
+              custom_table={_.find(child_tables, ['key',node.indexValue]).table}
               onMouseEnter={ (child_node, e) => handleHover(child_node, e.target, node.data.children) }
               onMouseLeave={ (child_node, e) => handleHover(child_node, e.target, node.data.children) }  
               onClick={ (child_node, e) => window.open(tick_map[child_node.indexValue], '_blank') }
+              graph_height = '500px'
```
Stephen-ONeil: Seeing `graph_height = '400px'` in most places (is that the default? maybe we shouldn't add it explicitly everywhere). Does this graph have a special reason to get more height?

client/src/components/modals_and_popovers/bootstrap_modal_exstension.scss
cookkate: These classes are copied from bootstrap 4's modals, but I don't know if there's a case where we wouldn't want it to be responsive, so might as well just have that, no?

client/src/charts/nivo/nivo_line.js
cookkate: Good catch!

cookkate:
>  lots of tables have empty cells, but some have zeros (e.g. first graph in this panel), at least sometimes. Zeros might be better because the table's, I think, harder to parse when there are gaps in it (since columns aren't a consistent width and there's no indication of where a column breaks within a row)

Filling with zeros is problematic for the `auth_exp_planned_spending` graph because it's not zeroes for those cases, we just aren't displaying the values...

client/src/panels/panel_declarations/finances/goco/goco.js
cookkate: Most yes but not overwhelmingly so. I can make 400px the default but its overridden a fair amount (e.g. for donuts).  Haven't tested it but everything is different for the gocographic so why not this? :P

client/src/components/DisplayTable.js
```diff
@@ -78,6 +78,18 @@ export class DisplayTable extends React.Component {
       .tap( descending ? _.noop : _.reverse )
       .value();
 
+    const all_sort_keys = _.chain(rows)
+      .map( row => _.keys(row.sort_values) )
+      .flatten()
```
Stephen-ONeil, line 83: Real minor, but could just be `flatMap` here and below.

client/src/components/DisplayTable.scss
```diff
@@ -1,5 +1,9 @@
 @import '../common_css/common-variables';
 
+table.display-table {
+  width: 99%;
+}
```
Stephen-ONeil, line 5: What's this a fix for? Maybe add a "why" comment for future reference, because it looks like something I'd be tempted to delete.

cookkate:
Oh shoot, I thought the blanks thing was fixed. I should be able to figure out what's going on.

client/src/components/DisplayTable.scss
cookkate: The border is outside the width which is set in bootstrap's `.table` to be 100%, which was putting a very slight scroll on the table. There's probably a better fix for this though? If not I can comment.

client/src/components/DisplayTable.js
cookkate: ah TIL :+1:

client/src/components/DisplayTable.scss
Stephen-ONeil: Oh, gross, hate scrolls like that. Not in a CSS headspace right now so can't give a better fix off the top of my head.

client/src/components/DisplayTable.js
```diff
@@ -74,8 +74,18 @@ export class DisplayTable extends React.Component {
           .every()
           .value()
       )
-      .sortBy( ({sort_values}) => sort_values[sort_by] )
-      .tap( descending ? _.noop : _.reverse )
+      .sortBy( ({sort_values}) => sort_values[sort_by] ? sort_values[sort_by] : Number.NEGATIVE_INFINITY )
```
Stephen-ONeil, line 77: Oop, looks like this will sort values of 0 as - Inf

client/src/components/DisplayTable.js
```diff
@@ -16,17 +16,18 @@ export class DisplayTable extends React.Component {
...
 
-    const searches = _.mapValues(
-      search_values,
-      () => "",
-    );
+    const searches = _.chain(rows)
+      .map( row => _.keys(row.search_values) )
+      .union().keyBy()
+      .mapValues( () => "" )
+      .value();
```
Stephen-ONeil, line 30: Not clear what this is doing. Is it equivalent to this?
```
const searches = _.chain(rows)
  .flatMap( 
    ({search_values}) => _.chain(search_values)
      .keys()
      .map( (search_key) => [search_key, ""] )
      .value()
  )
  .fromPairs()  // left out the uniq-ing step, happens implicitly here since the resulting object will only have one instance of each key
  .value();
```

client/src/panels/panel_declarations/people/employee_totals.js
```diff
@@ -72,36 +72,34 @@ export const declare_employee_totals_panel = () => declare_panel({
...
+                  is_money = {false}
+                  yScale = {{toggle: true}}
+                  tooltip = {slice =>  
+                    <div style={{color: window.infobase_color_constants.textColor}}>
+                      <table style={{width: '100%', borderCollapse: 'collapse'}}>
+                        <tbody>
+                          { slice.data.map(
+                            tooltip_item => ( 
+                              <tr key = {tooltip_item.serie.id}>
+                                <td style= {{padding: '3px 5px'}}>
```
Stephen-ONeil, line 89: This padding shows up a lot in this diff for nivo tooltips. Even a class that applies it (is supposed to apply it?), `nivo-tooltip__content`. Move to CSS is possible please!

client/src/components/DisplayTable.js
Stephen-ONeil: Because 0's falsey. Probably want `_.isNumber(sort_values[sort_by])`.

cookkate: I guess blank < 0 then?

client/src/components/DisplayTable.js
cookkate: Basically it needs to get all the search keys from all rows (it was assuming the first row had all the data) and setting up the `searches` object like so:

```
searches = {
  key1: "",
  key2: "",
  key3: "",
}
```
I think your code is equivalent, yeah.

PR #504, opened 2020-02-12
Show in-year estimates in "Authorities, Expenditures and Planned Spending" panel

### LINKED ISSUE #421: Authorities, Expenditures and Planned Spending should use in-year Authorities for the current year
Right now we're using end of year authorities, but there's no reason to wait until the year closes to show it.

cookkate:
Does what it says on the tin. Closes #421

TODO:

- [x] fix the A11y table version
- [x] double check that my logic on the years is right, based on the order that things (should be) tabled. -- I think we're good, if mains come out before DPs then its fine, if DPs come out before mains (unlikely) it would look weird but I *think* would still be fine. We should check on this graph when those updates go out though.
~~- [ ] possibly add another sentence of text?~~ Don't think this is necessary, it's already pretty wordy.

### DIFF
--- a/client/src/panels/panel_declarations/finances/historical_auth_exp/auth_exp_prog_spending.js
+++ b/client/src/panels/panel_declarations/finances/historical_auth_exp/auth_exp_prog_spending.js
@@ -11,7 +11,6 @@ import {
   create_text_maker_component,
   NivoResponsiveLine,
   newIBCategoryColors,
-  formatter,
   util_components,
 
   declare_panel,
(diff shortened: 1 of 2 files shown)


client/src/panels/panel_declarations/finances/historical_auth_exp/auth_exp_prog_spending.js
```diff
@@ -90,31 +94,19 @@ class AuthExpProgSpending extends React.Component {
...
+        .concat(exp_ticks, plan_ticks)
+        .uniq()
+        .value();
+
+      const data = _.map(all_ticks, tick => ({
+        label: tick,
+        data: [
+          auth[_.findIndex(auth_ticks, d=>d===tick)] || null,
+          exp[_.findIndex(exp_ticks, d=>d===tick)] || null,
+          progSpending[_.findIndex(plan_ticks, d=>d===tick)] || null,
```
Stephen-ONeil: Any cases where we'd want to preserve 0's? A null would be a gap in the graph, right?

cookkate: Hmmmm, maybe. I'll alter this to display a zero if there is one.

Issue #470, opened 2020-01-23
Google search snippets for GC infobase include "please enable javascript" error page


cookkate:
Weirdly only for the tag explorer snippet (none of the others).

Might not be much we can do.

![Screen Shot 2020-01-23 at 2 33 51 PM](https://user-images.githubusercontent.com/10406459/73017283-866c2800-3ded-11ea-95e1-4e4052603796.png)

Stephen-ONeil:
Did some googling on this, looks like maybe Google always indexes the content of a noscript tag, [at least in 2010](https://webmasters.stackexchange.com/a/1688). I doubt they've changed their underlying reasoning/methodology since then, even if websites that don't work without JS are far more common now.

Could be something else about the particular page(s) this happens for (I'm seeing it for rpb in current searches). Appears to have proper metadata for title and description though. Can't think of any other obvious reasons google would give so much priority to the noscript.

Note that the noscript content DOES always show up at the top of the body, so if google is crawling it then it's possible that if the meta tags aren't set it **would** have high priority.

Stephen-ONeil:
Suggested fix could be nuking the noscript tag from JS (maybe in app bootstrap or something). That should keep it out of search indexes (unless google really is choosing to crawl and index the site without JS, in which case we're probably out of luck).

Issue #443, opened 2020-01-02
More gracefully handle traffic to dead infographics


Stephen-ONeil:
In particular, a fair number of users have been trying to visit infographics for programs and CRSOs that fell out of our window recently. Currently, trying to route to one gets you an error page. 

A better behaviour would be to, for example, fall back to the parent department (or home page?) with a banner explaining why you can't visit the page you were linked to.

Easiest approach would be to bail back home on any failing infographic route. I don't like this because it could hide actual bugs down the road.

Best, but probably the most involved, approach would be to keep a list of recently dead subject ids and their appropriate fallback subjects. Then we'd have something to check against whenever an infographic route fails to match any of the current subjects. On the other hand, that would be a new dataset for the pipeline to produce and yet another thing to maintain. Would it be worth the effort?

bill-lliu:
I believe this is solved with #508 ^ right?
Good to be closed?

Stephen-ONeil:
#508 may not have fixed things, or maybe it will just take time for Google to re-crawl. Dead programs still show up in my google searches at least.

#508 also only fixed one way that this could happen, still want a graceful handling to deal with stale links and bookmarks.

PR #508, opened 2020-02-14
Add a robots noindex meta tag to the error boundary 

Stephen-ONeil:
Should clear/prevent future indexing of broken/dead routes in search engines. For example, you can search for a now-dead program and get an InfoBase link that takes results in an error boundary page.

This might be a considerable source of traffic to dead infographics, so possibly the root cause of #443. If I get around to digging through analytics to check that assumption, I'll update here with the answer.

### DIFF
--- a/client/src/core/ErrorBoundary.js
+++ b/client/src/core/ErrorBoundary.js
@@ -1,6 +1,13 @@
 import { get_static_url, make_request } from '../request_utils.js';
 import { log_standard_event } from './analytics.js';
 
+
+const NoIndex = () => ReactDOM.createPortal(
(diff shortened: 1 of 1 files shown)


cookkate:
Also this should close #470 no?

Stephen-ONeil:
Won't close #470. Unclear if #470 is caused by google actually crawling the site with JS disabled or because they might just always crawl the content of noscript tags. Either way, this fix a) requires JS and b) only applies when an errors thrown, so it doesn't fit in either of those potential cases.

cookkate:
Oops, misremembered. I thought it was crawling the error boundary itself.

PR #510, opened 2020-02-14
Negative Common Donut Fix

### LINKED ISSUE #509: Common Donut does not generate if there are negative numbers
Screen Shot 2020-02-14 at 12 24 22 🏙

Found in certain programs that have a negative total expenditure, as their revenues were greater than expenses, such as here:
https://www.tbs-sct.gc.ca/ems-sgd/edb-bdd/index-eng.html#orgs/program/AGR-BWN05/infograph/financial

bill-lliu:
This fixes the display of negative values on the common donut. 

The code that was modified should be cleaned up though, as pointed out by @Stephen-ONeil 

Fixes #509

### DIFF
--- a/client/src/charts/NivoCharts.js
+++ b/client/src/charts/NivoCharts.js
@@ -61,7 +61,7 @@ const smalldevice_percent_tooltip_content = (tooltip_item, formatter, total) => 
   <td>
     <div className="nivo-tooltip__content">{tooltip_item.id}</div>
     <div className="nivo-tooltip__content" dangerouslySetInnerHTML = {{__html: formatter(tooltip_item.value)}}/>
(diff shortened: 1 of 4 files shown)


Stephen-ONeil:
Taking a step back from this, I think we're solving the problem at the wrong level. Rather than try and make pie charts that (sort-of) know how to handle negatives, or deciding to hide the panel in cases where the sign of the values doesn't fit nicely in the chosen visualization, we should just pick a new visualization for it all together.

A one-tick stacked bar graph, either of absolute values or percentages, works about as well as a pie chart AND handles negative values just fine. The main benefit of a pie over a stacked bar is that "slivers" are displayed slightly better in the pie layout, but these graphs only compare two values so the sliver problem's minimized anyway.

PR #514, opened 2020-02-18
Internal Services Panel now hides the bar graph when in A11y Mode

### LINKED ISSUE #506: Graph Still Displaying in Accessibility Mode
Screen Shot 2020-02-13 at 4 20 18 🏙
This one graph still appears when the page is set to accessibility mode
Found here:
https://www.tbs-sct.gc.ca/ems-sgd/edb-bdd/index-basic-eng.html#orgs/dept/1/infograph/financial

bill-lliu:
The bar graphs were displaying in a11y mode for specifically:
level: dept, category: finance
panel: internal services (full name: Full-time employees in Internal Services, under declare_internal_services_panel)

I followed the same fixing format that I found in another file, specifically for:
level: dept, category: finance
panel: spend_rev_split (Full name: Expenditures and Revenues, under declare_spend_rev_split_panel)

Although the bar graph is now hidden, the data is not shown in a table. This is the case for a lot of other bar graphs on the page; as they are set to simply hide the data when in a11y mode. This should be fixed altogether in a future update. 

For now though, there shouldn't be any issues with graphs in a11y mode. 

fixes #506

### DIFF
--- a/client/src/panels/panel_declarations/finances/internal_services/internal_services.js
+++ b/client/src/panels/panel_declarations/finances/internal_services/internal_services.js
@@ -103,17 +103,11 @@ export const declare_internal_services_panel = () => declare_panel({
         return result;
       }, []);
 
-      const to_render = <div>
-        <div className="medium_panel_text" style={{marginBottom: "15px"}}>
-          <TM
-            k="internal_service_panel_text"
-            args={{
-              subject,
-              isc_fte_pct: isc_fte / total_fte,
-              gov_isc_fte_pct: gov_isc_fte / gov_fte_total,
-            }}
-          />
-        </div>
+      let graph_content;
+      if (window.is_a11y_mode) {
+        graph_content = null;
+      } else {
+        graph_content = 
         <div className="frow md-middle"> 
           <div className="fcol-md-3">
             <div className="well legend-container">
@@ -137,7 +131,21 @@ export const declare_internal_services_panel = () => declare_panel({
               }}
             />
           </div>
+        </div>;
+      }
+
+      const to_render = <div>
+        <div className="medium_panel_text" style={{marginBottom: "15px"}}>
+          <TM
+            k="internal_service_panel_text"
+            args={{
+              subject,
+              isc_fte_pct: isc_fte / total_fte,
+              gov_isc_fte_pct: gov_isc_fte / gov_fte_total,
+            }}
+          />
         </div>
+        {graph_content}
       </div>;
   
       return (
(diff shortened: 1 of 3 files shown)


client/src/panels/panel_declarations/finances/internal_services/internal_services.js
```diff
@@ -137,7 +131,21 @@ export const declare_internal_services_panel = () => declare_panel({
...
+          <TM
+            k="internal_service_panel_text"
+            args={{
+              subject,
+              isc_fte_pct: isc_fte / total_fte,
+              gov_isc_fte_pct: gov_isc_fte / gov_fte_total,
+            }}
+          />
         </div>
+        {graph_content}
```
Stephen-ONeil: Works, here's a couple alternate suggestions though.

1) Always good to avoid a `let` in my opinion. If you can re-write it as a `const` and avoid side effects, it makes the code easier/safer to reason about and more future-proof. Even if the `let` declaration is immediately followed by its assignment, it still opens up the possibility that maybe it's being modified elsewhere, and a future dev reading/working with this code will have to confirm that before continuing.

Ternaries are the easiest syntax for reducing a conditional assignment in to a single expression. If it's a complicated case, you can use an immediately invoked function (IIF), e.g.
```
const graph_content = (() => {
  if (window.is_a11y_mode) {
    return null;
  } else {
    return (
      <div className="frow md-middle"> 
        <div className="fcol-md-3">
          <div className="well legend-container">
            <GraphLegend
              items={legend_items}
            />
          </div>
        </div>
        <div className="fcol-md-9" style = {{height: '300px'}} aria-hidden = {true}>
          <NivoResponsiveBar
            data = {bar_data}
            indexBy = "date"
            colorBy = {d => colors(d.id)}
            keys = {label_keys}
            is_money = {false}
            margin = {{
              top: 15,
              right: 30,
              bottom: 40,
              left: 50,
            }}
          />
        </div>
      </div>
    );
  }
})();
```
Alternatively to an IIF, if it's reasonable, declare a separate `get_graph_content` function. Not really necessary here though.

2. `graph_content` doesn't really need to be its own variable here. Could put a ternary or IIF directly inside the braces at line 148. Braces in JSX are for inserting arbitrary JS, most often used to just drop in a variable but also commonly used to drop in a conditional like this.

bill-lliu: I've changed the case for the internal services panel, but I found there were many other cases that did use a let statement for this kinda stuff. Do you think it'd be worth going through and changing?

PR #527, opened 2020-02-24
Standardizing People Panels

### LINKED ISSUE #507: Standardizing Panel Displays Across Site
Just something I discussed with Ross, but what do you guys think about standardizing some of the panels for the people tab? Just because we noticed that some of the panels are inconsistent in that they sometimes show the average pie chart, sometimes the morphing bar graph, and sometimes both. It’s not always consistent between panel to panel. 

Another notice was just about the usefulness of the average pie chart in general, since if the multiple morphing graph existed, do we still really need the average pie chart to show the same data? (another thing to note is that the morphing bar graph shows the progress year to year, while the pie chart just averages the data across all five years then generates a pie chart afterwards - arguably not as useful)

I'll be going through anywhere that uses either the average donut display or the morphing graph, and standardize the panels to show the data using both - then, if we feel that the donut is useless, I can just remove them all from the panels

bill-lliu:
All panels now display both the pie chart and the morphing graph for more information 

(people can now also see the exact numbers on the morphing graph, as they could not before only seeing the pie chart, or had to change to text only mode)

Fixes #507

### DIFF
--- a/client/src/panels/panel_declarations/people/employee_executive_level.js
+++ b/client/src/panels/panel_declarations/people/employee_executive_level.js
@@ -1,4 +1,5 @@
 import text from "./employee_executive_level.yaml";
+import { AverageSharePie } from './AverageSharePie.js';
 import {
   formats,
   run_template,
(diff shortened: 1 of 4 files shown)


bill-lliu:
Changes moved to Kate's branch for Table Panels

Issue #376, opened 2019-11-25
Deeper level to more refine infographs within a bubble


ktw1016:
For example, after selecting `Finances` bubble, have some sort of a filter (like show only transfer payments panels, or FTE panels, etc.)

![Screen Shot 2019-11-25 at 2 40 30 PM](https://user-images.githubusercontent.com/29746626/69572192-9dc79a00-0f91-11ea-9730-279d01b05e3f.png)
Suggestion from Infobase Heuristic Report

ktw1016:
We know what tables for each panel so we can filter depending on what data each panel contains.
Problems with that:
1. Budget not tabled
2. Key concepts, empty table (this isn't actually a problem)

ktw1016:
I think #529 should have closed it. Reopen if needed

ktw1016:
This should never have closed actually

Stephen-ONeil:
Duplicate of #642

PR #529, opened 2020-02-26
Filter panels and filter icon

ktw1016:
related to: #376 , not sure if this PR should close the issue though

This is rough initial work


cookkate:
I like it!

Some comments about functionality:

When the details panel is closed but some filters are applied, can we indicate that somehow?

I don't like the upsidedown icon much 😂

Also I think the text should be tweaked to give some more explanation and mirror the search filter. My suggestion:
"Filter panels" (or maybe "Filter graphs?" "Infograph" isn't really a word we use)
And then when you open it have a sentence of explanation:
"Filter panels to only show graphs using the following datasets"

ktw1016:
> I don't like the upsidedown icon much 😂

ohhh I thought I was being clever with the filter icon😂I thought filter icon might help catch users looking for filter functionality quicker without them having to read it. Do you think I should get rid of it?

> When the details panel is closed but some filters are applied, can we indicate that somehow?

I'm not sure what you mean by this?

cookkate:
I do like the use of the "filter" icon (it is meaningful to me, might not translate well to others so may be worth polling the team or something to see if they know what it means?) I just don't think it needs to be rotated 180 when the panel is open.

If I unclick some of the filter options (so e.g. I'm only seeing program spending-related panels) and then close the details panel, there's no visual indication that some of them have been unchecked. Something like a "(filters applied)" text next to the "Filter infographs" or _maybe_ even just a * would signal that something has changed.

cookkate:
Hey! Showed to a couple of people in my section and they said they really liked it. There was the suggestion to increase the font size of the "Filter Panels" (and "Search Filters") text a little, it's pretty small. Also, my suggestion would be to move the "Showing X of Y panels" text to next to the "Filter Panels" title to save space. I also have not yet looked at what it looks like on mobile (where space is going to be a lot more limited).


Also, it doesn't work on results :) Since there is no way to usefully filter results panels, at this point I would disable the filter entirely on that view.

client/src/components/Details.js
```diff
@@ -32,7 +38,7 @@ export class Details extends React.Component {
         aria-labelledby={label_id}
       >
         <span aria-hidden className="IBDetails__TogglerIcon">
-          { isOpen ? "▼" : "►" }
+          { isOpen ? opened_drawer_icon || "▼" : closed_drawer_icon || "►" }
```
Stephen-ONeil: Alternatively, could use 
```
Details.defaultProps = {
  opened_drawer_icon: "▼",
  closed_drawer_icon: "►",
}
```

Don't actually feel too strongly one way or the other, but I think we (myself included) probably under-use default props.

client/src/infographic/Infographic.js
```diff
@@ -188,6 +195,29 @@ class InfoGraph_ extends React.Component {
         include_tables: false,
       }}
     />;
+    const filter_icon_props = {
+      color: window.infobase_color_constants.primaryColor,
+      title: "Filter icon for panels",
```
Stephen-ONeil: Needs translation

client/src/infographic/Infographic.js
```diff
@@ -305,8 +371,26 @@ class InfoGraph_ extends React.Component {
...
     }).then( () => {
       if ( shallowEqualObjectsOverKeys({active_bubble_id, subject, level}, this.state, ['subject','active_bubble_id','level']) ){
+        const panel_filter = _.chain(panel_keys)
+          .map(panel_key => tables_for_panel(panel_key, subject.level))
+          .flatten()
+          .uniq()
+          .reduce((result, table_id) => {
+            result[table_id] = true;
+            return result;
+          }, {})
```
Stephen-ONeil: A pretty benign mutation, because `result` is totally scoped to within this reduce. Maybe cleaner like this thoung:
```
.reduce(
  (result, table_id) => ({
    ...result,
    [table_id]: true
  }),
  {}
)
```

Alternatively replace the reduce with
```
  .map( (table_id) => [table_id, true] )
  .fromPairs()
```
?

client/src/infographic/Infographic.js
```diff
@@ -188,6 +195,29 @@ class InfoGraph_ extends React.Component {
...
+      title: "Filter icon for panels",
+      width: 15,
+      height: 15,
+      vertical_align: 5,
+    };
+    const panel_renderers = !loading &&
+      _.chain(panel_keys)
+        .map((panel_key) => {
+          const panel_obj = PanelRegistry.lookup(panel_key, subject.level);
+          const panel_filter_arr = _.map(panel_filter, (value, key) => panel_filter[key] && key);
```
Stephen-ONeil: `value === panel_filter[key]`? 

This also leaves a bunch of `false`s in the resulting array; unlikely to cause a bug but could if something causes `panel_obj.depends_on` to return something containing `false`. If you replace this with a chain than there's a few options to deal with that (this map followed by a `.compact()`, or even simpler just `_.chain(panel_filter).filter().keys().value()` I think).

Wasn't initially clear what `panel_filter_arr` is from the name, but it's something like `active_panel_filter_keys` (or `active_panel_data_dependencies`? since that's what filter keys are), right? If one of those is right, or you can think of an even clearer name, please update it.

client/src/infographic/Infographic.js
ktw1016: ah yes yes

client/src/infographic/Infographic.js
```diff
@@ -227,15 +261,60 @@ class InfoGraph_ extends React.Component {
...
+                  key="opened_filter"
+                  rotation={180}
+                />
+              }
+              summary_content={
+                <div>
+                  <span style={{fontSize: 16}}>
+                    <TM k="filter_panels"/>
+                  </span>
+                  <span> </span>
```
ktw1016: Not sure how I should separate this two text. Is this good enough?

client/src/infographic/Infographic.js
```diff
@@ -188,6 +195,33 @@ class InfoGraph_ extends React.Component {
...
+    const is_filter_panels = active_bubble_id === "financial" || active_bubble_id === "people";
+    const panel_renderers = !loading &&
+      _.chain(panel_keys)
+        .map((panel_key) => {
+          const panel_obj = PanelRegistry.lookup(panel_key, subject.level);
+          const active_panel_filter_keys = _.chain(panel_filter)
+            .map((value, key) => value && key)
+            .compact()
+            .value();
+          const is_filtered = _.intersection(active_panel_filter_keys, panel_obj.depends_on).length > 0 || !is_filter_panels;
```
ktw1016: Not sure if `is_filtered` and `!is_filter_panels` naming seem make sense?

client/src/infographic/Infographic.js
Stephen-ONeil: Kind of unrelated to your exact question but I just added something, in a PR merged to master today, that helps here a bit. If you rebase, you'll be able to pass `style` and `className` props directly to `TM`s., so you won't need to wrap them in bonus `span`s.

I think you could also get the space in as a text node by replacing `<span> </span>` with `{" "}`. Still a bit wonky, but at least clearer that there's an actual space character being inserted. Can't think of anything better than that right now.

ktw1016:
@Stephen-ONeil 
> Standard `details` element convention has the chevron pointing right when collapsed and down when expanded. Maybe the rotating filter drawer icon would feel better if it matched that pattern?

So for me, the reason I put the filter icons instead of chevrons was so users catch on to filter functionality quicker by seeing familiar icon. I personally think if I rotate it 90 degrees, regular chevron would be better?


> There's a `MultiColumnList` component in `components/misc_util_components.js`, could use that to let the filter items use more horizontal space/less vertical when expanded.

Screen Shot 2020-03-03 at 4 37 06 PM

So I tried it out like you can see above, I personally prefer the current one column list since it is collapsed by default and expandable. Also expanded list is not long (6-7 elements usually). Let me know what you guys prefer though, open to recommendations!

client/src/infographic/Infographic.js
Stephen-ONeil: That was part of what had me scratching my head earlier. I think right now having a value of `true` in `panel_filter` means a key is NOT being filtered out, i.e. is being included in results. The name `panel_filter` kind of makes it feel like it should be the other way around.

The text associated with this feature on the site is "Panel Filters", which I think is a good user-facing name for it, but internally `panel_filter`, the component state, is more like a `data_sources_to_display` object. Maybe that changes/is complicated if we add more filter options on the site, but I think it's a much easier name to use/think in terms of for now.

cookkate:
> Standard details element convention has the chevron pointing right when collapsed and down when expanded. Maybe the rotating filter drawer icon would feel better if it matched that pattern?

IMO the filter icon doesn't read as such when 'pointing' right.

cookkate:
Oh, missed the screenshot. I rather like the multi-column list. Even though the you can hide it I think it's better to lessen the amount that the page content is bumped up and down the page when you open and close the drawer.

Stephen-ONeil:
Yeah, I can see how the right-pointing filter icon is kind of meaningless. I think the upside down version looks strange too though, and I don't like breaking the details drawer convention. Maybe stick to a chevron and just add the filter icon to the summary text? Not as fancy, design wise, but might be that this is a case where keeping to the basics is for the best.

ktw1016:
I will get some feedback on this during the team meeting tomorrow or when email is back up!

ktw1016:
I'm having two problems with displaying total number of panels (`showing X of Y panels`):
- I can't get accurate sources for each panels
- I can't get accurate panel objects that are actually displayed for each subjects

Example is Parliamentary Protective Service. In Finance, there are 8 panels shown but according to my code, it's 10 panels. This is mostly due to the 2nd point mentioned

PR #530, opened 2020-02-26
RIP LiquidFillGauge

cookkate:
Its time has come. Replacing with a slightly hacked Nivo bubble chart in an effort to reduce custom d3 in the infographics to only the Canada graph, and get Nivo's nice tooltips etc.

TODO:

- [x] adjust visuals: remove border, make sure the colours are full opacity
- [x] check minimum size of inner circle, make sure it's functional (ish) on mobile but as small as possible (to avoid misrepresenting data as much as possible) -- not functional on mobile but easily visible
- [x] instead of labeling inner and outer, just label with a header that says something like "X as a fraction of Y" 
  - ~~*maybe* integrate with a legend??????~~ Don't think this is necessary given the current description + the important numbers are in the panel text.
- [x] fix unused variable warnings

Looks very similar to the old bubble charts:

Screen Shot 2020-03-09 at 9 21 19 AM


ktw1016:
😂👌

Stephen-ONeil:
Maybe give it a custom tooltip that shows both values no matter where you hover inside the graph?

cookkate:
Oh that's a good idea, I hadn't thought of that. Definitely would solve the problem, thanks.

client/src/charts/NivoCharts.js
```diff
@@ -1,13 +1,19 @@
...
 import { formats, dollar_formats } from "../core/format.js";
 import { Fragment } from 'react';
 import { IconZoomIn, IconZoomOut } from '../icons/icons.js';
+import { create_text_maker_component } from '../components/index.js';
 import { trivial_text_maker } from '../models/text.js';
 import { breakpoints } from '../core/breakpoint_defs.js';
+import { newIBCategoryColors } from '../core/color_schemes.js';
 import MediaQuery from 'react-responsive';
 import './NivoCharts.scss';
+import graph_text from './NivoCharts.yaml';
```
Stephen-ONeil: Not a totally strict code-style rule, but I try to keep imports from same-named yaml and scss modules as the the very first line of same-named js file (usually followed by a lone break and all npm modules then a line break and local modules, unless some other grouping of imports is more meaningful in the file).

client/src/charts/NivoCharts.js
```diff
@@ -608,6 +614,119 @@ NivoResponsiveLine.defaultProps = {
...
+      legends,
+      text_formatter,
+      labelSkipWidth,
+      height,
+      value,
+      name,
+      totalValue,
+      totalName,
+    } = this.props;
+    legends && (legends[0].symbolShape = fixedSymbolShape);
```
Stephen-ONeil: Mutations :(

Actually, what's going on with `legends` and `fixedSymbolShape`? Why does only the first element of `legens` need it's `symbolShape` property overridden? I don't see `fixedSymbolShape` defined in the PR diff, is this a pattern in NivoCharts?

client/src/handlebars/helpers.side-effects.js
```diff
@@ -390,6 +390,22 @@ function subject_name(subject){
...
 Handlebars.registerHelper('subj_name', subject_name);
 
+function subject_acronym_or_name(subject){
+
+  if (subject.is("dept")){
+    return (
+        !_.isEmpty(subject.abbr) ?
+        subject.abbr :
+        subject.fancy_name
+    );
```
Stephen-ONeil: Huh, surprised the linter's not flagging this indentation. IMO, 397 should have two less spaces.

client/src/charts/NivoCharts.js
```diff
@@ -608,6 +614,119 @@ NivoResponsiveLine.defaultProps = {
...
+      margin,
+      is_money,
+      legends,
+      text_formatter,
+      labelSkipWidth,
+      height,
+      value,
+      name,
+      totalValue,
+      totalName,
```
Stephen-ONeil: I'd put forward these prop names as more meaningful, thoughts?
```
child_value,
child_name,
parent_value,
parent_name,
```

cookkate: Yeah sure, but I want to make the point that this is set up as a visualization for 1 parent and 1 child only, and not a generic parent-child chart.

I might add a comment stating as such just to make sure someone doesn't think it's our generic wrapper around the Nivo bubble chart.

cookkate: Or maybe rename the whole component?

cookkate:
Yeah, I can have the tooltip show the whole thing. I used the acronyms because the phrase "Employment and Social Development Canada transfer payment spending as a fraction of Employment and Social Development Canada total spending" was a bit of a mouthful.

client/src/charts/NivoCharts.js
cookkate: I still plan on breaking up NivoCharts, so it would be a generic text to go along with those multiple js files.

client/src/charts/NivoCharts.js
cookkate: It's a `NivoCharts.js` pattern. (?????????) Will investigate on #500.

client/src/handlebars/helpers.side-effects.js
cookkate: :+1:

Stephen-ONeil:
Maybe if the tooltip was less of a legend and more of a sentence? Like:
"`parent_name` of `parent_value` (`parent_coloured_square`), `child_value` of which (`child_value_%`) is `child_name` (`child_coloured_square`)"

Dunno

cookkate:
How standard is it that we have text on the left, graphs on the right? It seems pretty standard (so forcing the tooltip to the left should be fine)....

I don't think the 100% is that weird, it seems clear enough to me. I could remove the value from the 100% one, but it would be nice to be able to use the standard tooltip and not customize it. If you think it's really problematic I'll change it.

Stephen-ONeil:
If text and graphs are side by side, at least right now, the graph's always on the right, yeah.

Hmm, guess we could run screen shots of the tooltip by the team and ask what their immediate interpretation is. Or, even better, find some rando's who aren't familiar with that panel.

Stephen-ONeil:
Ready for review! Want to rename and cleanup some chart stuff (kind of started a little here), but to keep things in easier to review I'll stop working on that here and do it in a fresh PR.

TODO:
  - [ ] translation for a couple fragments

client/src/core/format.js
```diff
@@ -129,7 +129,7 @@ const percentage = (precision, val, lang, options) => {
 };
 
 // 'smart' percent formatter that scales decimal places (up to 3) based on the value
-const percentage_smart = (min_precision, val, lang, options) => {
+const smart_percentage = (min_precision, val, lang, options) => {
   const smart_precision = val < 0.001 ? 3 :
     (val < 0.01 ? 2 : 1);
```
bill-lliu, line 134: bracket consistency might help with clarity (just adding brackets to the top statement)

client/src/charts/wrapped_nivo/nivo_line.js
```diff
@@ -152,13 +167,13 @@ export class NivoResponsiveLine extends React.Component {
             motionDamping,
             motionStiffness,
           }}
-          legends={ IE_fixed_legends }
+          legends={ fix_legend_symbols(legends) }
           tooltip={ (d) => tooltip( d, get_formatter(is_money, text_formatter, false) ) }
```
bill-lliu, line 171: maybe change the `d` variable to be a slight bit more explicit

client/src/charts/wrapped_nivo/nivo_line.js
```diff
@@ -185,16 +200,20 @@ export class NivoResponsiveLine extends React.Component {
 }
 NivoResponsiveLine.defaultProps = {
   ...general_default_props,
-  tooltip: (slice, tooltip_formatter) => default_tooltip(
-    slice.data.map( 
+  tooltip: (slice, formatter) => {
+    const tooltip_items = slice.data.map( 
       (d) => ({
```
bill-lliu, line 205: `d` to more explicit variable here too

client/src/charts/wrapped_nivo/nivo_common.js
```diff
@@ -0,0 +1,238 @@
...
+      .clone()
+      .assign({symbolShape: fixed_symbol_shape})
+      .value()
+  ) :
+  undefined;
+
+
+const DefaultTooltipColorLegend = ({tooltip_item}) => (
+  <td className="nivo-tooltip__content">
+    <div style={{height: '12px', width: '12px', backgroundColor: tooltip_item.color}} />
```
bill-lliu, line 44: don't know if this is long enough, but maybe expand this out

client/src/charts/wrapped_nivo/nivo_bubble.js
```diff
@@ -56,79 +110,83 @@ export class CircleProportionChart extends React.Component{
...
-
-
-    const title = <div>{nivo_common_text_maker("bubble_title",{outer: parent_name, inner: child_name})}</div>;  
-    const table = !disable_table_view && <DisplayTable rows={table_data} column_names={column_names} ordered_column_keys={ordered_column_keys} name={nivo_common_text_maker("bubble_title")} />;
+    const title = <div
+      dangerouslySetInnerHTML={{
+        __html: text_maker(
+          "bubble_title", 
+          {outer: parent_name, inner: child_name}
+        ),
```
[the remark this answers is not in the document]
Stephen-ONeil: This one could be a `TM` component.

client/src/charts/wrapped_nivo/nivo_common.js
Stephen-ONeil: Thought I moved the height and width to CSS, must have done that somewhere else and missed this.

client/src/charts/wrapped_nivo/nivo_common.js
```diff
@@ -88,15 +100,21 @@ const DefaultPercentTooltip = ({tooltip_items, formatter, total}) => (
...
           <MediaQuery maxDeviceWidth={breakpoints.maxSmallDevice}>
             <td>
-              <div className="nivo-tooltip__content">{tooltip_item.name || tooltip_item.id}</div>
-              <div className="nivo-tooltip__content" dangerouslySetInnerHTML = {{__html: formatter(tooltip_item.value)}}/>
-              <div className="nivo-tooltip__content" dangerouslySetInnerHTML = {{__html: formats.percentage1(Math.abs(tooltip_item.value)/total)}}/>
+              <div className="nivo-tooltip__content">
+                {tooltip_item.name || tooltip_item.id}
+              </div>
+              <div className="nivo-tooltip__content" dangerouslySetInnerHTML={{__html: formatter(tooltip_item.value)}} />
+              <div className="nivo-tooltip__content" dangerouslySetInnerHTML={{__html: formats.percentage1(Math.abs(tooltip_item.value)/total)}} />
```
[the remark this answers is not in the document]
Stephen-ONeil: Yeah, hmm, I'll need to double check if either of these are necessary. Some of the formatters do return HTML strings, but they usually have a "raw" option to avoid that (might just be able to use raw here and skip the headache, if that's the case).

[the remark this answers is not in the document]
ktw1016: Hmm, `DefaultPercentTooltip` and `DefaultTooltip` does seem pretty redundant, except this line (108 and 117). I wonder if you could add `children` (or config) props to `DefaultTooltip` and basically pass in line 117? And you could still have it as separate variable like so.

```
const const DefaultPercentTooltip = ({tooltip_items, formatter, total}) => <DefaultTooltip (...) children={
  ({tooltip_item}) => <div (...) dangerouslySetInnerHTML={{__html: 
    formats.percentage1(Math.abs(tooltip_item.value)/total)}} />
}
```
I didn't actually think this through though so not sure if it'll work. But something along this line is worth a try?

client/src/charts/wrapped_nivo/nivo_bubble.js
```diff
@@ -0,0 +1,242 @@
...
+        {
+          id: child_name,
+          name: child_name,
+          value: child_value,
+          color: color_scale(child_name),
+        },
+      ],
+    };
+
+    const tooltip = () => (
```
ktw1016, line 132: Is it possible to get `TooltipFactory` or `DefaultTooltip` from `nivo_common` and make use of those here in this case?

client/src/charts/wrapped_nivo/nivo_line.js
Stephen-ONeil: Ehh, can't really think of one, maybe `datum`? It's a vague D3 wrapped chunk of graph data, with a bunch of other stuff crammed in there. The D3 style (which I don't love) is to just call that `d` (I feel like that betrays D3's rough edges, that they spend so much time dealing with messy variables that can't be easily named).

client/src/charts/wrapped_nivo/index.js
```diff
@@ -1,12 +1,14 @@
 import { NivoResponsiveBar, NivoResponsiveHBar } from './nivo_bar.js';
 import { NivoResponsiveLine } from './nivo_line.js';
 import { NivoResponsivePie } from './nivo_pie.js';
-import { LineBarToggleGraph } from './nivo_misc.js';
+import { CircleProportionChart } from './nivo_bubble.js';
+
+export { NivoLineBarToggle } from './NivoLineBarToggle.js';
```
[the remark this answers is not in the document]
Stephen-ONeil: Yeah, that was part of me starting to go overboard on the cleanup. As I pick better names for the modules/components here (in a future branch) I'll be switching to the direct re-export syntax.

Stephen-ONeil:
Ahh, haha, hadn't noticed things were broken in French (I knew translation was not finished, guess I would have noticed while doing that) or the a11y tables (oops). Thanks for double checking on me!

PR #539, opened 2020-02-27
make new checkbox back to standard checkbox for a11y

ktw1016:


client/src/components/CheckBox.js
```diff
@@ -4,19 +4,32 @@ export class CheckBox extends React.Component{
...
+      <label style={a11y_label_style}>
+        <input 
+          type='checkbox'
+          checked={active}
+          onChange={onClick}
+          style={{marginRight: 5}}
+        />
+        { label }
+      </label>
+    </div>
```
cookkate: I'm confused about why you are applying a `style` at all to an a11y-specific version component....is there a particular reason for it?

ktw1016: Do you mean `a11y_label_style`? It's because some text for checkbox are bolded for A11y version. However, this bold is not needed for standard version because it is `link-styled `(blue, underlined text)

If you meant `marginRight`, this is actually already applied to the new fancy checkbox with `legend-color-checkbox`

ktw1016: `a11y_label_style` may be unnecessary though. I've only put it up to bold the text and I hated myself for doing it for adding another messy `style` props..

cookkate: I don't think the bold style is necessary.

PR #544, opened 2020-03-03
eslint setting need to be fixed because of new update

ktw1016:


cookkate:
Does this actually make code changes on save? That's not something we had enabled before (as far as I know)....maybe I'm missing something.

ktw1016:
> Does this actually make code changes on save? That's not something we had enabled before (as far as I know)....maybe I'm missing something.

ohhh yes this does.. I guess it was only me had it locally in vscode settings

PR #547, opened 2020-03-04
Adv search design change

### LINKED ISSUE #496: Advanced Search design change
Screen Shot 2020-02-06 at 1 30 52 PM

How about something like this? This is probably unnecessary but.. if I add panel filter, it might look confusing with two filters in a close by space and this would clarify the search filter usability, also search filter will be usable in the home page as well.

Not sure how customizable `Typeahead` can be, but this could also possibly be implemented alongside the potential changes to our search

related to: #137, #340

ktw1016:
The problem I'm facing is `TypeAhead` search component. Since it is using `input type='text'` internally, I have to insert string of HTML into the DOM to create the filter button. Code may get ugly since I need to use other components such as `StatelessModal` but could probably still get it to work.

I'm thinking of getting position of filter button by `getBoundingClientRect()` and then placing modal somewhere near there to make it look like it's coming out of the filter button. Then just insert the existing Advanced Search into modal.

closes #496


ktw1016:
Closing for now. Will come back to it later

PR #550, opened 2020-03-05
Be explicit about files to include when deploying GCloud functions

Stephen-ONeil:
GCloud function deploys are pointed at a directory and then deploy everything inside of it, unless that directory contains a `.gcloudingore` in its root which can explicitly **exclude** files.

Didn't like that, was too easy to add a new file to a project root and accidentally include it in all deploys, and was hard to retroactively detect when this had happened.

Instead, I've set up our `.gcloudignore` files to ignore EVERYTHING, and then explicitly **include** just the files actually required in a deploy. This way, if someone adds a new file that's actually required and forgets to include it, things should at least fail-fast.

### DIFF
--- a/email_backend/.gcloudignore
+++ b/email_backend/.gcloudignore
@@ -1,23 +1,9 @@
-.gcloudignore
-.git
-.gitignore
+# ignore everything
+/[!.]*
+/.?*
 
-node_modules
-*.DS_Store
-*.swp
-*.swo
-.vscode/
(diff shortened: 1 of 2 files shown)


Stephen-ONeil:
... caught on to this because prod deploys had become outrageously slow ever since I'd moved the server's mongo dir in to the project. Took me too long to realize that was because I was shoving all of the mongo files in to each function bucket.

PR #553, opened 2020-03-09
Results summary (ibcolors version)

cookkate:
One possibility for improving the results summary.

Adds a donut plot to the main DRR summary, controllable with a graph legend in the usable way, that lets a user decide what the denominator should be for calculating result status %s.

### DIFF
--- a/client/src/panels/panel_declarations/results/drr_summary.js
+++ b/client/src/panels/panel_declarations/results/drr_summary.js
@@ -210,13 +210,15 @@ class PercentageViz extends React.Component {
       })
     );
     
+    const all_data_total = _.sumBy(all_data, 'value');
 
     const graph_data = _.filter(all_data, d=>_.includes(selected,d.id));
     const graph_total = _.sumBy(graph_data, 'value');
 
     const new_summary_text_args = {
       year: current_drr_year,
-      drr_total: graph_total,
+      drr_subset: graph_total,
+      drr_total: all_data_total,
       drr_indicators_met: _.includes(selected, 'met') && counts.met,
       drr_indicators_not_met: _.includes(selected, 'not_met') && counts.not_met,
       drr_indicators_not_available: _.includes(selected, 'not_available') && counts.not_available,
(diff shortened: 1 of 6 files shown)


client/src/panels/panel_declarations/results/drr_summary.js
```diff
@@ -180,6 +181,95 @@ const StatusGrid = props => {
...
+  }
+
+  render(){
+    const { counts } = this.props;
+    const { selected } = this.state;
+
+    const data = _.chain(counts)
+      .toPairs()
+      .map(pair => ({label: result_statuses[pair[0]].text, value: pair[1], id: pair[0]}))
+      .value();
```
Stephen-ONeil: Patterns like this are a useful case for array destructuring, to avoid having to access by index/to add clarity
i.e. `.map(pair =>` to `.map( ([key, value]) =>`

Alternatively, when mapping an object, the second argument of the iterate func becomes is the key. So, this could be written as
```
const data = _.map(
  counts,
  (value, key) => ({
    label: result_statuses[key].text,
    id: key,
    value
  })
);
```

cookkate: :+1:

cookkate:
Depends on #559 or similar fix (to avoid me having to do something super awkward, at least)

client/src/panels/panel_declarations/results/results_common.js
```diff
@@ -185,6 +186,16 @@ const filter_and_genericize_doc_counts = (counts, doc_key) => {
...
 };
 
+const result_color_scale = d3.scaleOrdinal()
+  .domain(["met","not_met","not_available","future"])
+  .range([
+    newIBCategoryColors[0],
+    newIBCategoryColors[1],
+    newIBCategoryColors[2],
+    newIBCategoryColors[3],
+  ]);
```
Stephen-ONeil: Why not just an object mapping keys to colours? What's the d3 wrapping add?

cookkate: For nivo?

PR #554, opened 2020-03-09
RPB changes

### LINKED ISSUE #551: Rough draft of modified RPB mock up.
Screen Shot 2020-03-05 at 12 53 42 PM
Have it multi-step of Table Selection as shown above. I think you could make it into 3 step process by adding `Select organization` as step 2.

Screen Shot 2020-03-05 at 12 54 03 PM
This is more of a table functionality. I think all of the Filter, Group Column Selection, share, export, etc. can be moved right above the table and inside the table. This has to do with #478 . Also some wording changes would be nice

I think also renaming "report builder" to something simpler and intuitive like "View Dataset" or something along this line?

ktw1016:
- [x] Multi-step Table Picker (implemented DeptSearch as Step 2 but commented it out)
- [x] Removed slow scroll
- [x] Removed `Currently selected` text in `Select your data`
- [x] Removed `All of Government of Canada` button in `Select organization`
- [x] Moved FTE from people to money category
- [x] Remove `Select the type of report you want` into table
- [x] Remove `Sort and filter your report` into other places

- [x] Make column toggle more visible? (Or enable all columns by default
- [x] Where to put "Filter data"
- [x] Check screen reader for `DropdownMenu` control
- [ ] Top of the page title to probably include the topic
- [ ] Change text from tables to datasets.. or something similar to that (?)
- [ ] Change name of "Report Builder" (?)

closes #551


ktw1016:
Another possible improvement is to click outside of dropdown to close dropdown. But I'll leave that out for now.. until someone asks for it

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -174,6 +193,38 @@ export class DisplayTable extends React.Component {
...
+                          .pickBy((col, key) =>
+                            _.chain(visible_ordered_col_keys)
+                              .toggle_list(clicked_key)
+                              .includes(key)
+                              .value()
+                          )
+                          .map(({ index }, key) => [index, key])
+                          .sortBy(_.first)
+                          .map(_.last)
+                          .value(),
```
Stephen-ONeil, line 222: This would be made a lot simpler if the array of visible keys was stored in state un-ordered, e.g. as just `visible_col_keys`. `all_ordered_col_keys` is declared during render, so the ordered version of visible col keys may as well also be calculated (e.g. `_.without(all_ordered_col_keys, visible_col_keys)`) rather than stored in state. Then updating the state would be as easy as `_.toggle_list(visible_col_keys, clicked_key)`.

Not the most common/direct use of it, but this is sort of a single-source-of-truth thing. Column order is derived from `col_configs_with_defaults` within the render function, so encoding it back in to state with `visible_ordered_col_keys` is just an inefficient duplication.

client/src/components/DropdownMenu.js
```diff
@@ -0,0 +1,44 @@
...
+        <button
+          className={button_class_name}
+          onClick={() => this.setState({ isOpen: !isOpen })}
+        >
+          <IconComponent
+            title={button_description}
+            color={icon_color}
+            alternate_color={false}
+          />
+        </button>
```
Stephen-ONeil, line 32: Gonna need to double check this with a screen reader, I think it will read out that it's on a button but won't read out the title of the icon inside it without the user drilling down. Guessing this is a mistake made in other parts of the code base though, only just occurred to me (oops, sure it's a mistake I've also made). Well, test if it seems properly usable at least, I could be wrong about what the screen reader will do.

client/src/components/DropdownMenu.js
```diff
@@ -0,0 +1,44 @@
...
+        <div
+          className={classNames(
+            "dropdown-content",
+            isOpen && "dropdown-is-open"
```
Stephen-ONeil, line 36: BEM would be `dropdown-content--is-open` for the modifier.

Stephen-ONeil: Oh, `dropdown__content` too, since there's a parent `dropdown`

client/src/panels/panel_declarations/index.js
```diff
@@ -21,9 +21,8 @@ export {
   declare_in_year_stat_breakdown_panel,
   declare_in_year_voted_stat_split_panel,
   declare_vote_stat_split_panel,
-  declare_welcome_mat_panel,
-} from //declare_tp_by_region_panel,
-"./finances/index.js";
+  declare_welcome_mat_panel, //declare_tp_by_region_panel,
+} from "./finances/index.js";
```
Stephen-ONeil, line 26: Hmm, this and some other changes look like weird Prettier artifacts. Any idea what's happened? Could be from back when I rebased this branch...

client/src/rpb/TablePicker.js
```diff
@@ -10,7 +10,8 @@ import { AlertBanner, GlossaryIcon } from "../components";
 import {
   categories,
   concepts_by_category,
-  concept_filter,
+  concept_white_filter,
+  concept_filter_by_type,
```
Stephen-ONeil, line 14: Trying to figure out what these are by prop name alone, and having a pretty hard time. Better names you can give these?

client/src/rpb/index.js
```diff
@@ -276,20 +277,17 @@ class RPB extends React.Component {
...
-                  }}
+                  className="container app-font modal-container"
                 >
-                  <TablePicker
-                    onSelect={(id) => this.pickTable(id)}
-                    broken_url={broken_url}
-                  />
+                  {!dataset_type && <Step1 />}
+                  {/*dataset_type && !selected_subject && <Step2/>*/}
+                  {dataset_type /* && selected_subject */ && (
```
Stephen-ONeil, line 284: Here and a bit above there's some commented out stuff. Looks like this `Step2` stuff's currently not used? While I'm at it, `Step1` and `Step2` are kind of vague names.

client/src/panels/panel_declarations/index.js
ktw1016: I think I may have changed this, as I was rebasing as it gave me conflicts. I'm not 100% sure as I went through like 60 different files to rebase this lol

client/src/rpb/TablePicker.js
ktw1016: `concept_whitelist_filter`
`concept_filter_by_categories` ?

ktw1016: I can leave a comment in the code for `concept_filter_by_type` if it's not clear

ktw1016:
- Actually never thought of dropping "Select organization" control, as I was actually mostly working on `simple_view` on my local branch, then dropped it afterwards. Makes sense to drop it since all organizations are listed anyways in the table and they can filter it using the table filter.

- For enabling all columns by default for RPB, this was my initial approach as well but wanted to get some more feedback on how this should work. I'll try to get more feedback on this one

- How do you think `filter data` dropdown control should go into heading utils? I suppose I could put it in as a different icon, or even straight into the same control with column toggle? For dividing it up with download/copy, I can put it at the left most heading utils for column toggle

- dollar signs are not changed from live site. I didn't modify anything for that

client/src/rpb/index.js
ktw1016: Ah yes, so I was considering making it a 2 step table picker process, with selecting organizations but saw some disadvantages to it. I'll drop this completely

client/src/panels/panel_declarations/index.js
Stephen-ONeil: Ah, ouch, ok!

Stephen-ONeil:
- responded on slack, think we should keep all applicable footnotes for all subjects in one big list for now, work on better organizing all footnote lists as follow-up
- :+1:
- when I was first writing that comment, I hadn't realized yet that the column selection was a new generic `DisplayTable` util, thought it was in some way an RPB specific thing. Hadn't looked at the code yet.  Moving the column selection out or the RPB filter in doesn't make sense, yeah. Left-aligning the column selection could work, although I kind of think we should start using that empty left hand of the table headers for the actual table titles. Maybe separate the table control utils and the table copying utils with a vertical bar? Or bigger margins and labels for the two control groups? That's kind of assuming we'll keep adding utils, but I think that's likely
- ah, huh! Wonder if that's a bug or if we intentionally removed them at some point, hmm

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -71,25 +100,22 @@ export class DisplayTable extends React.Component {
         },
       }
       */,
-      disable_header_utils,
+      enable_utils,
+      custom_utils,
```
ktw1016, line 104: I think how I handle `enable_utils` and `custom_utils` can be improved here. Any suggestions welcome

client/src/components/DropdownMenu.js
ktw1016: Just tried it out on Mac voiceover, it works as expected (although I got rid of icon anyways)

client/src/components/DisplayTable/DisplayTable.js
Stephen-ONeil: I'd suggest dropping `enable_utils` and just having one prop for `util_components`. Each user of `DisplayTable` will need to import and pass in any of the shared utils it wants from `DisplayTableUtils.js` along with its own custom utils. A bit more boilerplate everywhere, but fully flexible and consistent between custom and shared utils.

`DisplayTable` will have to pass some reasonable set of props to the util components it renders. That's mostly for the shared utils (particularly download and copy which want the table-processed data), the custom ones can ignore those props and smuggle in any variables or callbacks from the higher scope where they're defined.

Stephen-ONeil: At least, that's how I'd do it. Could still have set the default value for `util_components` as the current set of shared components, but anywhere that wants a custom set of utils will need to re-insert the shared utils along with the custom ones it passes in.

ktw1016: Indeed that's more consistent and flexible, however, that means utils need to be imported and pass in its own copy/download csv everywhere, as well as calculating its own csv data. I personally don't like the idea of having to do more work everytime you make a new table for features when `DisplayTable` is capable of doing it all

ktw1016: how about `util_components`, with shared components as booleans, and just leave a room for custom utils in there, and use that?

or, make separate `shared_utils` that store shared util components like `<DisplayTableDownloadCsv>`, have prop like `util_components` as above. If value is `true`, render what's in `shared_utils`, if it's not boolean, just render it? I know that sounds confusing but it really isn't

Stephen-ONeil: 1) RE: importing and passing in shared utils everywhere. All of the default shared utils can be imported by `DisplayTable` and set as the default prop value for `util_components`. Only instances that want any sort of custom `util_components` will have to import and include whichever of the defaults they still want to keep
  2) RE: having to calculate its own data. Can be avoided if we set the standard that the rendering `DisplayTable` passes some consistent set of props, like display-ready data, in to all of the `util_components` its rendering. Both the shared utils and the custom utils could make use of those but, as commented above, the custom utils can also access anything from the scope they're defined in (meaning they can smuggle in things like callbacks and don't necessarily need to make use of any of the props `DisplayTable` tries to pass)

Stephen-ONeil: If any of that's unclear still, I could implement my suggested approach pretty quickly to show what I mean.

ktw1016: ohhhh okay, this is actually in align with what I was thinking of, basically

ktw1016:
One minor problem that I'm having hard time fixing is: when there is no data shown on the table, table div height is small but dropdown content height is big so you have to scroll through dropdown content in a small area
I tried setting `overflow-y: visible` everywhere and setting `min-height` of table div but didn't wanna mess around with getting rendered element's height. Any ideas around this?

Stephen-ONeil:
Could remove `.dropdown__content`'s `position: absolute` and play around with letting the dropdown expand the header as needed. Would need a bit of tweaking and styling to make that feel decent though (i.e. maybe a transition for the drawer sliding in, and better connection between the button and the attached drawer?), and it's questionable since it might be making everything worse just to make one edge case better.

Where's the `overflow-y` coming from and what' it necessary for? I don't remember.

Side note, I noticed that the `.dropdown__content` divs are `aria-hidden`, but the dropdown buttons themselves aren't. Shouldn't be hard to unsure these controls are accessible, if they aren't already (`aria-hidden` aside), so that `aria-hidden` should be dropped. 

`aria-hidden` isn't necessary when the drawer is closed either because it uses `display: none`. Anything with `display: none` or `visibility: hidden` is ignored by screen readers anyway.

ktw1016:
I played around with `position: absolute` removed, but I couldn't do it with it decently looking whatsoever. So I just increased min height of no table data message to compensate the edge case problem

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -147,167 +168,225 @@ export class DisplayTable extends React.Component {
...
+                              col_configs_with_defaults[col_key].formatter(
+                                total_row[col_key]
+                              )
+                            )
+                          ) : (
+                            total_row[col_key]
+                          )
+                        ) : (
+                          ""
+                        )}
```
Stephen-ONeil, line 381: Not a necessary change, but this ternary's gotten a bit big. Would suggest refactoring the as an immediately invoked function:
```
<td key={col_key}>
  {(()=>{
    const has_total_row = total_row[col_key];
    if (has_total_row){
      // various nested cases
    } else {
      return "";
    }
  })()}
</td>
```

client/src/components/TwoLevelSelect.js
```diff
@@ -1,33 +0,0 @@
...
-          <option key={choice.id} value={choice.id}>
-            {choice.display}
-          </option>
-        ))}
-      </optgroup>
-    ))}
-  </select>
-);
-
-export { TwoLevelSelect };
```
Stephen-ONeil, line 33: I don't know about deleting this, feel like it could be useful in the future. Did you see one of your new components as totally replacing this functionality?

client/src/extended_bootstrap_css/bootstrap-wet-fixes-extensions.scss
```diff
@@ -268,25 +268,32 @@ div.label {
...
-/* to be used with light-background */
-a.btn-ib-light {
-  color: inherit;
+.btn-ib-opened {
+  background-color: $backgroundColor;
+  border: 2px solid $backgroundColor;
+  color: $primaryColor !important;
+  border-radius: $standardBorderRadius;
+  margin: 0px 5px 0px 0px;
 }
```
Stephen-ONeil, line 277: Maybe this should be named `.btn-ib-light--reversed` or something? It's specific to the light buttons, and, outside of the context of the buttons for those dropdowns, a button being "opened" doesn't mean much. Should also give it its own, `:hover` styles, etc.

client/src/extended_bootstrap_css/bootstrap-wet-fixes-extensions.scss
```diff
@@ -268,25 +268,32 @@ div.label {
...
+
+/* to be used with dark-background */
 .btn-ib-light {
-  background-color: inherit;
-  border: 2px solid $buttonPrimaryColor;
-  color: inherit;
-  border-radius: $standardBorderRadius; 
+  background-color: $primaryColor;
```
Stephen-ONeil, line 281: This had a use, `background-color: inherit;` was so that we could use this class on any dark background, not just `$backgroundColor`, and have it look like it fit in. Was it necessary to change?

client/src/extended_bootstrap_css/bootstrap-wet-fixes-extensions.scss
```diff
@@ -268,25 +268,32 @@ div.label {
...
+  border: 2px solid $backgroundColor;
+  color: $backgroundColor;
+  border-radius: $standardBorderRadius;
+  margin: 0px 5px 0px 0px;
```
Stephen-ONeil, line 285: This margin is specific to the use case in the RPB, right? Please make it some styling local to the RPB use case, the styles here should be for the most generic use case (e.g. same margin defaults as other buttons, which I think is none?) to keep it reusable.

client/src/rpb/TablePicker.scss
```diff
@@ -213,3 +213,17 @@
   transition: max-height 500ms ease-out, opacity 500ms ease-out;
 
 }
+.back-button {
```
Stephen-ONeil, line 216: Unused?

client/src/rpb/granular_view.js
```diff
@@ -153,112 +68,96 @@ class GranularView extends React.Component {
...
-                        : (e) => {
-                            if (e.keyCode === 13 || e.keyCode === 32) {
-                              on_set_page(ix);
-                              this.refs.table.focus();
+      <DisplayTable
+        data={table_data}
+        column_configs={column_configs}
+        util_components={{
+          rpb_group_data: (
+            <DropdownMenu
```
Stephen-ONeil, line 107: Big JSX pyramid below here. Could possibly define inner chunks as separate components to break up and flatten this a bit?

client/src/rpb/granular_view.js
```diff
@@ -153,112 +68,96 @@ class GranularView extends React.Component {
...
+                            onClick={(evt) => {
+                              const [
+                                dim_key,
+                                filt_key,
+                              ] = evt.target.value.split("__");
+                              on_set_filter({
+                                filter: filt_key,
+                                dimension: dim_key,
+                              });
+                            }}
```
Stephen-ONeil, line 137: Don't need to access the value off the event object, it's directly available in scope as `child.id`. Code's a bit more clear as
```
onClick={() => {
  const [
    dim_key,
    filt_key,
  ] = child.id.split("__");
  on_set_filter({
    filter: filt_key,
    dimension: dim_key,
  });
}}
```

That said, storing two values in `id` and then doing all these splits/comparisons using `__` is a bit hacky. I know that's how the RPB was originally written, but could you update the `state_and_data.js` and this code so that `dim_key` and `filter_key` are available individually as properties of `child`?

client/src/rpb/rpb.scss
```diff
@@ -6,22 +6,29 @@
   min-width: 75%;
 }
 
+.modal-container {
```
Stephen-ONeil, line 9: Could this be `rpb-modal-container`? Just to keep things scoped. Fine right now, but feels like the kind of class name that someone will accidentally step on eventually.

client/src/rpb/rpb.scss
```diff
@@ -6,22 +6,29 @@
...
+  overflow: auto;
+  line-height: 1.5;
+  padding: 0px 20px 0px 20px;
+  border-radius: 5px;
+  font-weight: 400;
+  display: flex;
+  flex-direction: column;
+}
+
+.trinity-container {
```
Stephen-ONeil, line 20: `rpb-trinity-container`? See above, just being cautious

client/src/rpb/rpb.scss
```diff
@@ -93,35 +97,32 @@ li.select-list-item {
...
-  justify-content: center;
-  & > .pagination > li > span {
-    padding: 10px 16px;
-  } 
-}
-
-.row-opition-content-search > div > div > .input-lg {
-  height: 38px;
-  line-height: 0.9;
+.full-width {
```
Stephen-ONeil, line 100: A small utility class like this should probably be in `common_css`. I think we've mostly stuck them in `site.scss` till now, but I feel like we should start collecting generic utility classes in something like `common_css/utility_classes.scss` or something.

client/src/components/DisplayTable/DisplayTableUtils.scss
```diff
@@ -13,4 +13,7 @@ button.display_table__util-icon {
   &:active {
     transform: scale(1.15);
   }
+}
+.no-right {
```
Stephen-ONeil, line 17: Oh, later on I make a comment about utility classes that would apply to this too. They should be defined globally.

client/src/rpb/rpb.scss
```diff
@@ -93,35 +97,32 @@ li.select-list-item {
...
+    transform: scale(1.15);
   }
-}
\ No newline at end of file
+}
+.group_filter_dropdown {
+  display: flex;
+  flex-direction: column;
+}
+.normal-rd-label {
```
Stephen-ONeil, line 125: What's "rd"? I could go look where it's used to figure out, but it's nicer to spell things out in the class name itself. If the place it's used was deleted, this would just be a mystery class.

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -147,167 +168,225 @@ export class DisplayTable extends React.Component {
...
+        />
+      ),
+    };
+    const util_components_with_defaults = _.chain({
+      ...util_components_default,
+      ...util_components,
+    })
+      .map()
+      .reverse()
+      .value();
```
Stephen-ONeil, line 235: I still have opinions about the implementation of this, but we went back and forth on it enough already, haha. Fine for now. If you still want to know how I was thinking of doing it let me know next week and I'll make a quick draft!

client/src/components/TwoLevelSelect.js
ktw1016: It does not replace this, I can keep it alive for now

client/src/extended_bootstrap_css/bootstrap-wet-fixes-extensions.scss
ktw1016: It's not a necessary change, it's more like a clean up as there's no actual button that use this class, it's only used in anchor tag in igoc explorer which is not even used on dark background. I think it looks much better as `btn-ib-primary`, so I decided to basically repurpose it to use it like this

ktw1016: I can revert it if you think it's necessary, but I don't see the point in leaving this class

Stephen-ONeil: Hmm, ok. If it's for a particular use case, move it out of the common css and in to local css. Maybe give it a more specific name too.

client/src/extended_bootstrap_css/bootstrap-wet-fixes-extensions.scss
ktw1016: agreed with naming, but for hover styles, I intentionally left them blank to give it a more of "opened" look, so that colour stays "opened" all the time. Just tried it out with hover styles applied and it looks a bit off

client/src/extended_bootstrap_css/bootstrap-wet-fixes-extensions.scss
ktw1016: Well, I think the same style can potentially be used elsewhere in the future for more buttons with dark background? Even now, it's used in DisplayTable and rpb

Stephen-ONeil: Doesn't `background-color: inherit;` === `background-color: $backgroundColor;`, in both cases? My point was that inheriting the parent's background colour would probably almost always mean having a background colour of `$backgroundColor`, but would also make it fit in seamlessly on other dark backgrounds if we happened to have any.

Stephen-ONeil: Unless we had some jumbled HTML somewhere so that the inherited background colour wasn't the colour of the element it was positioned over, which was why I asked if this was a necessary change.

ktw1016: ohhh yes I misinterpreted, it is both `$backgroundColor`, however, I needed it to be more explicit because of `btn-ib-light--reversed` so that the color is reversed explicitly

client/src/components/DisplayTable/DisplayTable.js
ktw1016: Huh, I thought this was kinda what we both had in mind😂, default utils that can be changed, and within the same prop, can add any new custom utils. Only using one object prop.. Feel free to make changes !

client/src/extended_bootstrap_css/bootstrap-wet-fixes-extensions.scss
ktw1016: So my point stands that it can stay in common css since this can still be used more in other places IMO, let me know if you think otherwise

ktw1016:
I agree with the problems you mentioned with dropdown and experimented a lot of different things. Anyways, I like the current look (with the recent change and here's why):
- Increasing `min-width` will have the dropdown burst out of page, forcing user to scroll. I prefer the current look than that. You can argue to use `getBoundingClientRect` to dynamically change the direction of which dropdown opens but I prefer to not do hacky things with DOM and also it will burst out of screen on mobile regardless the direction
- To close the dropdown when a selection is made, I'd have to lift the state up and I would just prefer to keep Dropdown API simpler. This also means I have to move column selection util into `DisplayTable` rather than `DisplayTableUtils`, as well as creating a state inside table for a dropdown which also clutters DisplayTable states. For jumpyness, I can just remove displaying group filter text in `dropdown_trigger_txt`, although I like showing that. I guess it's a tradeoff

haha this was a lot of excuses to not do work but I experimented a lot of different things and came back to just keeping it as is. If you have implementation suggestions on how to achieve these, let me know. Thanks!

Stephen-ONeil:
Ok, gave it a closer look. The actual cause of the radio icon and label ending up on different lines is the `display: inline-block` on the label. Add `display: inline` to `.normal-radio-btn-label`, will wrap as expected after that. No need for the min-width stuff then.

I'll take your word for it on the dropdown closing point, haha. Feels fine now, although I agree that it would be nice to fit the current group by selection on the screen somewhere.

PR #558, opened 2020-03-10
Refinements to last minute gov result panel banners

Stephen-ONeil:
Wanted short banners INSIDE the gov level results panels, to re-state that totals are not final when any departments are late. Slapped together in master to release with the DP, but refined and made automatic (for both DRR and DP) here.

### DIFF
--- a/client/src/panels/panel_declarations/results/result_components.yaml
+++ b/client/src/panels/panel_declarations/results/result_components.yaml
@@ -29,3 +29,8 @@ indicator_display_desc:
 report:
   en: Report
   fr: Rapport
+
+result_late_depts_warning:
+  transform: [handlebars,markdown]
+  en: Totals below are incomplete, data is not yet available for **{{late_dept_count}}** {{plural_branch late_dept_count "organization" "organizations"}}. Updates will follow.
+  fr: Les totaux ci-dessous sont incomplets, les données ne sont pas encore disponibles pour **{{late_dept_count}}** {{plural_branch late_dept_count "organisation" "organisations"}}. Des mises à jour suivront.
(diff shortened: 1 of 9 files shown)


client/src/panels/panel_declarations/results/result_components.yaml
```diff
@@ -28,4 +28,9 @@ indicator_display_desc:
...
 report:
   en: Report
-  fr: Rapport
\ No newline at end of file
+  fr: Rapport
+
+result_late_depts_warning:
+  transform: [handlebars,markdown]
+  en: Totals below are incomplete, data is not yet available for **{{late_dept_count}}** {{plural_branch late_dept_count "organization" "organizations"}}. Updates will follow.
+  fr: Les totaux ci-dessous sont incomplets, les données ne sont pas encore disponibles pour **{{late_dept_count}}** {{plural_branch late_dept_count "organisation" "organisations"}}. Des mises à jour suivront.
```
cookkate: Is there a reason this has different text than the other alert?

Are there other panels that should be showing this?

Stephen-ONeil: This one is gov level only, but, unlike the other gov level warning, I didn't want it displaying the full length list of late departments (since that's already at the top of the page and in footnotes). This basically just exists so users jumping straight to a gov level panel from the home page are made aware when departments are late with their results.

Arguably, the same banner could go on the gov level results intro panel, but nothing (of ours) links directly to that and it's right underneath the main, full length, warning banner.

Dept level (and below) panels don't need it I figure, they aren't actively misleading when a department is late (aside from maybe implying that subject WON'T have that year's doc) and aren't linked to directly so users seeing them should will have scrolled past the main warning banner on the way there (again, not directly linked to by us at least).

PR #560, opened 2020-03-11
Late resource orgs footnote/banner refinement 

Stephen-ONeil:
More chasing down of this that I had to rush for the DP update. Arguably, the only necessary change here was correcting "late planned spending data" to "late planned spending and planned FTE data". Everything else was just coding principles.

### DIFF
--- a/client/src/models/footnotes/dynamic_footnotes.js
+++ b/client/src/models/footnotes/dynamic_footnotes.js
@@ -8,10 +8,6 @@ import { create_text_maker } from '../text.js';
 const all_subject_classes= [Gov, Dept, CRSO, Program];
 const text_maker = create_text_maker(text);
 
-
-const depts_with_late_planned_spending = [];
-
-
 const expand_dept_cr_and_programs = (dept) => [
   dept,
   ...dept.crsos,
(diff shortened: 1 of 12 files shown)


client/src/models/footnotes/dynamic_footnotes.yaml
```diff
@@ -29,15 +29,15 @@ late_results_warning_crso:
   fr: Les données du {{result_doc_name}} de cette responsabilité essentielle ne sont pas encore disponibles. Une mise à jour suivra.
 
 
-late_planned_spending_warning_gov:
-  en: Planned spending data does not include values from the organizations listed below, as their data is not yet available. Updates will follow.
+late_resources_warning_gov:
+  en: Planned spending and planned FTE data does not include values from the organizations listed below, as their data is not yet available. Updates will follow.
```
cookkate: Technically should probably be and/or, right? But I don't think we should change it.

client/src/models/footnotes/dynamic_footnotes.js
```diff
@@ -31,85 +27,70 @@ const get_dynamic_footnotes = () => {
...
+            [`${_.toUpper(doc_type)}_RESULTS`],
+          text: `<p>${text_maker(`late_${result_or_resource}_warning_gov`, {result_doc_name: text_maker(`${doc_type}_name`, {year})} )}</p>${
+            `<ul>${
+              _.reduce(
+                late_orgs,
+                (elements, org_id) => `${elements}<li>${Dept.lookup(org_id).fancy_name}</li>`,
+                ''
+              )
+            }</ul>`
+          }`,
```
cookkate: Oof, this is hard to read. Suggest adding a couple of line breaks:

```
text: `<p>
    ${text_maker(`late_${result_or_resource}_warning_gov`, {result_doc_name: text_maker(`${doc_type}_name`, {year})} )}
  </p>
  ${ `<ul>
    ${
      _.reduce(
        late_orgs,
        (elements, org_id) => `${elements}<li>${Dept.lookup(org_id).fancy_name}</li>`,
        ''
      )
    }
  </ul>` }`,
```

client/src/models/footnotes/dynamic_footnotes.yaml
Stephen-ONeil: It's possible someone fills out one and not the other, but I think we expect it to be all or nothing?

client/src/models/footnotes/dynamic_footnotes.js
Stephen-ONeil: Haha, thought I could slip that by, I agree it's rough. A bit better with syntax highlighting, but I'll take the suggestion. Thanks!

client/src/models/footnotes/dynamic_footnotes.yaml
cookkate: I mean, hopefully...

PR #562, opened 2020-03-13
fancy_name cleanup

Stephen-ONeil:
~Be more consistent with using it as the default field when displaying a name (renamed it to `display_name` to highlight that), although I don't think I'll hunt down every single place we display a subject name.~

Drop the department subject's `fancy_name` (briefly known as `display_name`), have the department `name` property itself behave as `fancy_name` used to.

At the same time, shift to prioritizing applied titles of departments over legal titles.

Also expand the name of the government from "Government" to "the Government of Canada" (a bit annoying, but it's up to devs to upper case the first character as-needed).

TODO:
  - [x] maybe add a toggle in the IGOC (and elsewhere, estimates comparison? don't want to go overboard) so that users can go back to prioritizing legal names
  - [x] Add glossary entry for applied title (coordinate with pipeline)
    - Applied title: approved name used to identify a department, program or activity. This title is used in all communications
    - Titre d’usage: titre approuvé, utilisé pour identifier un ministère, un programme ou une activité. Ce titre est utilisé dans toutes les communications.
  - [x] global estimates and PA footnote explaining that those are tabled by legal name, but displayed in the InfoBase by applied title
  - [x] edit glossary and footnote file in pipline inputs
  - [x] translations
  - [x] oh, should make sure client Subject changes are mirrored in the API models. **Edit:** aha, looks like the org subject in the API dropped `fancy_name` a long time ago, all good there


client/src/handlebars/helpers.side-effects.js
```diff
@@ -459,7 +427,7 @@ Handlebars.registerHelper("le_dept",function(context){
   if (article.length > 0 && _.last(article) !== "'") {
     article += " ";
   }
-  return article + (dept.applied_title || dept.name);
+  return article + (dept.name);
```
ktw1016, line 430: I suppose this could be written as string interpolation but such a minor pick since it's just simple string concatenation

Stephen-ONeil: I think I didn't bother cleaning it up more since the handelbars helpers file is such old junky code. Broken windows and all that.

client/src/models/organizational_entities.js
```diff
@@ -33,12 +26,11 @@ const Gov = {
   level: 'gov',
   get has_planned_spending(){ return true;},
```
bill-lliu, line 27: I think there's a missing space at the end of the return true

client/src/common_text/common_lang.yaml
```diff
@@ -279,6 +282,9 @@ open_data_datasets:
 link:
   en: Link
   fr: Lien
+the_goc:
+  en: the Government of Canada
+  fr: le gouvernement du Canada
```
bill-lliu, line 287: I think the G was meant to be capital in the french version too

client/src/IgocExplorer/explorer_view.js
```diff
@@ -179,7 +190,13 @@ const get_non_col_content_func = createSelector(
...
     super();
-    this.state = { _query: "" };
+    this.state = {
+      _query: "",
+
+      // note use_legal_titles exists totally outside the redux here. No need for it to be redux-state,
+      // and not worth all that added boiler plate to get it in there
+      use_legal_titles: false,
+    };
     this.debounced_set_query = _.debounce(this.debounced_set_query, 500);
```
bill-lliu, line 200: Sorta more just a personal question for clarification, but why is this constructor debounced 500ms? Shouldn't it only run whenever the `ExplorerForIgoc` is called, which seems to be only once when `ExplorerContainer` is run

client/src/common_text/common_lang.yaml
Stephen-ONeil: Double checked, other uses in the IB have lower case g in French

client/src/IgocExplorer/explorer_view.js
Stephen-ONeil: The constructor of a React component is only called once right before it first mounts, yeah. Nothing is being debounced within the constructor though, `_.debounce` takes a function and a debounce time and returns a new, debouncing, version of that function. What's happening here is that a new debouncing function is being created, it's not called until later.

I tried to clean the code up a bit anyway, still a mess because there's a weird thing going on with the loading state in this component...

bill-lliu: Ohh okay. Is that the `.debounced_set_query` part that's the new function?

Stephen-ONeil: Yeah, although it was a bit messy because the class had a pre-defined function named `debounced_set_query` that was not initially debounced but was being overwritten with a debounced version in the constructor. I cleaned that mess up.

bill-lliu: Ohhh makes a lot more sense. I don't think that's messy per-se, I just didn't get the whole debouncing at first

PR #568, opened 2020-03-18
Periodic audit

Stephen-ONeil:
`npm audit fix` doesn't seem to be having the expected behaviour, multiple runs claim to fix the same number of issues each time. Possible that the local package in `scripts/common_node_scripts` is causing some sort of confusion?


Stephen-ONeil:
Woof, only one point of data here but, after the testcafe update, the CI browser tests step has more than doubled in run time (... and it hasn't finished running yet).

PR #573, opened 2020-03-24
Fix server logs

Stephen-ONeil:
Those messy hotfixes on master last Friday didn't fix anything.

This code's always been a real headache because, it turns out, I've had an inconsistent picture of what/how these request objects should look in practice (doesn't help that there are so many dang cases, and hacky things going on with the compressed GET requests).

Should probably rethink all of this, or at least capture and normalize all the special cases in a single place, instead of juggling every special case in every different util function/test. That's out of scope for this PR though.


Stephen-ONeil:
TODO: double check that everything works on the dev link once live. The variables parsed out of the compressed GET requests were being left string encoded, surprised that GraphQL wasn't choking on that! They're parsed in to objects now, but I'll need to double check now that expressGraphQL doesn't ACTUALLY expect variables in POST bodies to be stringified objects or something.

PR #587, opened 2020-04-08
Rework DisplayTable API

### LINKED ISSUE #584: Rework DisplayTable API
Rework so that data is just raw values and config object prop would change how columns should display/sort/search/sum (total)

related PR: #570

ktw1016:
closes #584 

- [x] Some API documentation
- [x] Still need to do more clean up
- [x] sort function
- [x] search func?

### DIFF
--- a/client/src/components/DisplayTable.js
+++ b/client/src/components/DisplayTable.js
@@ -29,9 +29,7 @@ export class DisplayTable extends React.Component {
     
     const searches = _.chain(column_configs)
       .pickBy(col => col.is_searchable)
-      .keys()
-      .map(search_key => [search_key, ""])
-      .fromPairs()
+      .mapValues( () => "" )
       .value();
 
     this.state = {
@@ -112,10 +110,13 @@ export class DisplayTable extends React.Component {
     );
     const is_total_exist = !_.isEmpty(total_row);
 
-    const ordered_column_keys = _.reduce( column_configs, (ordered_arr, column, column_key) => {
-      ordered_arr.splice(column.index, 0, column_key);
-      return ordered_arr;
-    }, [] );
+    const ordered_column_keys = _.chain(column_configs)
+      .map( ({index}, key) => [index, key] )
+      .sortBy( _.first )
+      .map( _.last )
+      .value();
+
+    const default_true = (val) => _.isUndefined(val) || val ? true : false;
 
     return (
       <div style={{overflowX: "auto", marginTop: "20px", marginBottom: "20px"}}>
@@ -148,7 +149,7 @@ export class DisplayTable extends React.Component {
                 _.map(
                   ordered_column_keys,
                   (column_key) => {
-                    const sortable = column_configs[column_key].is_sortable;
+                    const sortable = default_true(column_configs[column_key].is_sortable);
                     const searchable = column_configs[column_key].is_searchable;
 
                     const current_search_input = (searchable && searches[column_key]) || null;
@@ -244,3 +245,6 @@ export class DisplayTable extends React.Component {
     );
   }
 }
+DisplayTable.defaultProps = {
+
+};
--- a/client/src/components/DisplayTable.js
+++ b/client/src/components/DisplayTable.js
@@ -17,12 +17,18 @@ export class DisplayTable extends React.Component {
 
     const {
       unsorted_initial,
-      column_config,
+      column_configs,
     } = props;
 
-    const sort_by = unsorted_initial ? null : _.first(column_config.sort);
+    const sort_by = unsorted_initial ? null : _.chain(column_configs)
+      .pickBy(col => col.is_sortable)
+      .keys()
+      .first()
+      .value();
     
-    const searches = _.chain(column_config.search)
+    const searches = _.chain(column_configs)
+      .pickBy(col => col.is_searchable)
+      .keys()
       .map(search_key => [search_key, ""])
       .fromPairs()
       .value();
@@ -48,31 +54,8 @@ export class DisplayTable extends React.Component {
   render(){
     const {
       table_name, // Optional: Name of table
-      column_names, // { column_key: column_name, column_key2: column_name2 }
-      ordered_column_keys, // [column_key, column_key2]
-      data, /* 
-      [
-        {
-          column_without_any_display_formats: {
-            value: 1, **required
-          }
-          column_with_display_format_in_column_config: {
-            format_type: "big_int"
-            value: 10, **required
-          }
-        }
-      ]
-      */
-      column_config, /* Everything Optional
-        {
-          sort: [column_key, column_key2]
-          search: [column_key]
-          total: {column_key2: "big_int"}
-          display: {
-            column_key: ({format_type, value}) => <Format type={format_type} content={value}/>
-          }
-        }
-      */
+      data,
+      column_configs,
     } = this.props;
     const {
       sort_by,
@@ -83,17 +66,17 @@ export class DisplayTable extends React.Component {
     const clean_search_string = (search_string) => _.chain(search_string).deburr().toLower().trim().value();
     const sorted_filtered_data = _.chain(data)
       .filter( (row) => _.chain(row)
-        .map( (column, column_key) => (
+        .map( (column_value, column_key) => (
           _.isEmpty(searches[column_key]) ||
             _.includes(
-              clean_search_string(column.value),
+              clean_search_string(column_value),
               clean_search_string(searches[column_key])
             )
         ) )
         .every()
         .value()
       )
-      .sortBy( row => ( row[sort_by] && _.isNumber(row[sort_by].value) )
+      .sortBy( row => ( row[sort_by] && _.isNumber(row[sort_by]) )
         ? row[sort_by].value : Number.NEGATIVE_INFINITY )
       .tap( descending ? _.reverse : _.noop )
       .value();
@@ -102,13 +85,23 @@ export class DisplayTable extends React.Component {
       sorted_filtered_data,
       (totals, row) => _.mapValues(
         totals,
-        (total, col_key) => total + row[col_key].value
+        (total, col_key) => total + row[col_key]
       ),
-      _.mapValues(column_config.total, () => 0)
+      _.chain(column_configs)
+        .pickBy(col => col.is_summable)
+        .mapValues(col => 0)
+        .value()
     );
+    const is_total_exist = !_.isEmpty(total_row);
+
+    const ordered_column_keys = _.reduce( column_configs, (ordered_arr, column, column_key) => {
+      ordered_arr.splice(column.index, 0, column_key);
+      return ordered_arr;
+    }, [] );
+
     return (
       <div style={{overflowX: "auto", marginTop: "20px", marginBottom: "20px"}}>
-        <table className={classNames("table", "display-table", !column_config.total && "no-total-row")}>
+        <table className={classNames("table", "display-table", !is_total_exist && "no-total-row")}>
           <caption className="sr-only">
             <div>
               { 
@@ -122,23 +115,23 @@ export class DisplayTable extends React.Component {
             <tr className="table-header">
               {
                 _.map(
-                  column_names,
-                  (col_name, i) => <th
+                  ordered_column_keys,
+                  (column_key, i) => <th
                     key={i} 
                     className={"center-text"}
                   >
-                    {col_name}
+                    {column_configs[column_key].header}
                   </th>
                 )
               }
             </tr>
             <tr className="table-header">
-              { sorted_filtered_data.length > 0 &&
+              {
                 _.map(
                   ordered_column_keys,
                   (column_key) => {
-                    const sortable = _.includes(column_config.sort, column_key);
-                    const searchable = _.includes(column_config.search, column_key);
+                    const sortable = column_configs[column_key].is_sortable;
+                    const searchable = column_configs[column_key].is_searchable;
 
                     const current_search_input = (searchable && searches[column_key]) || null;
 
@@ -187,25 +180,32 @@ export class DisplayTable extends React.Component {
                 <tr key={i}>
                   {_.map(
                     ordered_column_keys,
-                    col => (
-                      <td style={{fontSize: "14px"}} key={col}>
-                        {column_config.display[col] ?
-                          column_config.display[col](row[col]) : row[col].value}
+                    col_key => (
+                      <td style={{fontSize: "14px"}} key={col_key}>
+                        {column_configs[col_key].formatter ?
+                          _.isString(column_configs[col_key].formatter) ?
+                            <Format type={column_configs[col_key].formatter} content={row[col_key]} /> :
+                            column_configs[col_key].formatter(row[col_key]) : 
+                          row[col_key]}
                       </td>
                     ))}
                 </tr>
               )
             )}
-            { column_config.total &&
+            { is_total_exist &&
               <tr key="total_row">
                 <td>{text_maker("total")}</td>
                 { _.chain(ordered_column_keys)
                   .tail()
-                  .map(col => (
-                    <td key={col}>
-                      { total_row[col] ? 
-                        <Format type={column_config.total[col]} content={total_row[col]}/> : 
-                        ""
+                  .map(col_key => (
+                    <td key={col_key}>
+                      { total_row[col_key] ?
+                          column_configs[col_key].formatter ?
+                            _.isString(column_configs[col_key].formatter) ?
+                              <Format type={column_configs[col_key].formatter} content={total_row[col_key]} /> :
+                              column_configs[col_key].formatter(total_row[col_key]) :
+                            total_row[col_key] :
+                          ""
                       }
                     </td>
                   ))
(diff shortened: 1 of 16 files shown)


Stephen-ONeil:
Good start, but I think column name and order could also go in to the column config. I think something like this would work well:

`rows` prop that looks like this:
```
[
  { // this object is row 1
    col_key_1: a_raw_value,
    col_key_2: a_raw_value,
    // etc.
  },
  { // this object is row 2
    col_key_1: a_raw_value,
    col_key_2: a_raw_value,
    // etc.
  },
  // etc.
]
```

Note that me saying "row 1" only applies if the data is sorted, and `col_key_1` doesn't mean that value is in the first column. Column order would be in the `column_configs` prop, which could look like:
```
{
  col_key_1: {
    column_index: /* order in table  */,
    header: /* string, name for header  */,
    is_sortable: /* boolean  */,
    sort_func: (a, b) => {
      /*
        A function that defines how to compare two raw values, a and b from this column.
        Will be passed as the sort function to a _.sortBy() at some point.
        Default to _.Identity.
      */
    },
    is_summable: /* boolean  */,
    sum_func: (sum, value) => {
      /*
        A function that defines how to sum the raw values in this column.
        Will be passed as the iterate to a _.reduce() over the raw data at some point.
        Default to (sum, value) => sum + value.
        I think it will almost always be left as a default, but I think with a little trickery I could supply something to 
        calculate an average instead, for instance
      */
    },
    sum_initial_value: /* anything, default to 0 */,
    is_searchable: /* boolean  */,
    // I don't think we need a search func, just run values through the formatter and do string matches
    formatter: //*function OR string matching the name of one of our formatters, default to _.identity */,
  },
}
```

Can figure out if a total row is needed for the table with `_.some(column_configs, 'is_summable')`. Values in the total row would use their column's formatter.

Anything missing, or seem unwieldy in that?

Stephen-ONeil:
The only other props to `DisplayTable` would be the table name and maybe some flags to turn on/off an additional table-level features like the show-raw-data button or etc.

Stephen-ONeil:
Thinking ahead a bit, to the idea of a button that pivots a table's header and first column, I think that would either need to disable searching and sorting OR require that every non-index 0 column config has the same formatter and search func. That should be doable behind the scenes though, I don't think we need to think too much about that while designing this API.

client/src/components/DisplayTable.js
```diff
@@ -71,50 +64,49 @@ export class DisplayTable extends React.Component {
...
-      _.mapValues(total_row_config, () => 0)
+      _.chain(column_configs)
+        .pickBy(col => col.is_summable)
+        .mapValues(col => 0)
+        .value()
     );
+    const is_total_exist = !_.isEmpty(total_row);
+
+    const ordered_column_keys = _.reduce( column_configs, (ordered_arr, column, column_key) => {
+      ordered_arr.splice(column.index, 0, column_key);
```
ktw1016, line 98: Trying to make ordered column keys here. Not sure if this is the right approach

client/src/panels/panel_declarations/finances/vote_stat/in_year_vote_stat_breakdown.js
```diff
@@ -76,54 +76,45 @@ const planned_vote_or_stat_render = vs => function ({ calculations, footnotes, s
...
-  const rows_with_complement_amt = _.concat(rows, [{
-    display_values: {
-      name: "",
-      voted_stat: complement_desc,
-      amount: <Format type={amt_type} content={complement_amt} />,
+  
+  const subj_map = _.chain(top_10_rows)
+    .map( obj => [Subject.Dept.lookup(obj.dept).name, infograph_href_template(Subject.Dept.lookup(obj.dept))] )
+    .fromPairs()
+    .value();
```
ktw1016, line 84: Mapping subjects by name here and multiple other places with these `map`. I think these are bad and potentially dangerous but not sure what to do. I can put dept code as the value and use that but I don't really like having to define sort, search every time..

client/src/panels/panel_declarations/results/gov_dp.js
```diff
@@ -65,42 +65,43 @@ export const declare_gov_dp_panel = () => declare_panel({
...
-          display_values,
-          sort_values,
-          search_values,
-        };
-      });
-      const total = _.chain(column_keys)
-        .keys()
-        .map(key => [key, "big_int"])
+      const subj_map = _.chain(dept_counts)
+        .map(row => [ Dept.lookup(row.id).name, link_to_results_infograph(Dept.lookup(row.id)) ])
```
ktw1016, line 69: another map here

client/src/panels/panel_declarations/results/gov_drr.js
```diff
@@ -86,41 +86,49 @@ export const declare_gov_drr_panel = () => declare_panel({
...
-          display_values,
-          sort_values,
-          search_values,
-        };
-      });
-      const total = _.chain(column_keys)
-        .keys()
-        .map(key => [key, "big_int"])
+      const subj_map = _.chain(dept_counts)
+        .map(row => [ Dept.lookup(row.id).name, link_to_results_infograph(Dept.lookup(row.id)) ])
```
ktw1016, line 90: another map here

client/src/panels/panel_declarations/results/result_flat_table.js
```diff
@@ -74,56 +73,67 @@ const subject_link = (node) => (
...
-    date_to_achieve: text_maker("date_to_achieve"),
-    status: text_maker("status"),
+  const subject_link_map = _.chain(indicator_list)
+    .map(ind => [ind.parent_node.data.name, subject_link(ind.parent_node)])
+    .fromPairs()
+    .value();
+  const indicator_id_map = _.chain(indicator_list)
+    .map(ind => [ind.indicator.name, `#indicator/${ind.indicator.id}`])
+    .fromPairs()
+    .value();
```
ktw1016, line 84: more maps here

client/src/rpb/simple_view.js
```diff
@@ -199,57 +199,57 @@ class SimpleView extends React.Component {
...
+        first_col_nick === "dept" ?
+          granular_rpb_link_for_org(this.props, Dept.lookup(row_first_col_nick_value)) :
+          granular_rpb_link_for_filter(this.props, row_first_col_nick_value) ];
+      })
+      .fromPairs()
+      .value();
+    const data_type_map = _.chain(columns)
+      .map(col => [col.nick, col.type])
+      .fromPairs()
+      .value();
```
ktw1016, line 215: more maps here

client/src/panels/panel_declarations/finances/goco/goco.js
```diff
@@ -95,49 +95,80 @@ class Goco extends React.Component {
...
+        const total_spending = total_fte_spend[sa.id].total_child_spending;
+        return [total_spending, spending_table_formatter(total_spending) ];
+      })
+      .fromPairs()
+      .value();
+    const ftes_map = _.chain(Tag.gocos_by_spendarea)
+      .map(sa => {
+        const total_ftes = total_fte_spend[sa.id].total_child_ftes;
+        return [total_ftes, spending_table_formatter(total_ftes) ];
+      })
```
ktw1016, line 109: more maps here

client/src/components/DisplayTable.js
```diff
@@ -71,50 +79,54 @@ export class DisplayTable extends React.Component {
...
-          ) )
-          .every()
-          .value()
+        ) )
+        .every()
+        .value()
       )
-      .sortBy( ({sort_values}) => _.isNumber(sort_values[sort_by]) ? sort_values[sort_by] : Number.NEGATIVE_INFINITY )
+      .sortBy( row => row[sort_by] && (
+        column_configs[sort_by].sort_func ?
```
ktw1016, line 95: Lodash currently doesn't support comparator natively: https://github.com/lodash/lodash/pull/3764
I think I can add mixin to make it work though

Stephen-ONeil: Always forget that. The JS built in array sort takes a comparator, could do
```
  // ...
  .thru( (unsorted_array) => unsorted_array.sort(
    (a, b) => {
      // whatever
    }
  ) )
  // ...
```

Stephen-ONeil: A mixin would be handy though!

client/src/panels/panel_declarations/results/result_flat_table.js
```diff
@@ -74,56 +74,68 @@ const subject_link = (node) => (
...
+      formatter: (val) => _.isDate(val) ?
+        `${months[val.getMonth() + 1].text} ${val.getFullYear()}` : val,
+    },
+    status: {
+      index: 5,
+      header: text_maker("status"),
+      is_sortable: true,
+      formatter: (value) => <Fragment>
+        <span aria-hidden="true" className="copyable-hidden">{result_statuses[value].text}</span>
+        {window.innerWidth < breakpoints.mediumDevice ? status_icons[value] : large_status_icons[value]}
```
Stephen-ONeil, line 122: Should use React `MediaQuery` components so this automatically re-renders to the appropriate size when the screen resizes. Also `breakpoints.mediumDevice` is undefined, it's either `minMediumDevice` or `maxMediumDevice` (depending if you're doing a `<MediaQuery minWidth={minMediumDevice}>` or `<MediaQuery maxWidth={maxMediumDevice}>` respectively).

Looking at it now, that stuff's super poorly documented in the InfoBase, my bad. If you want to know why breakpoints are stupid like this ask some other time (long story short: they're a nightmare).

client/src/charts/nivo/nivo_line.js
```diff
@@ -93,19 +93,32 @@ export class NivoResponsiveLine extends React.Component {
...
+        header: table_first_column_name || graph_text_maker("label"),
+        is_sortable: true,
+        is_searchable: true,
+      },
+      ..._.chain(table_ordered_column_keys || _.map(data,'id'))
+        .map( (col, idx) => [col, {
+          index: idx + 1,
+          header: col,
+          is_sortable: true,
+          formatter: (value) => value ? get_formatter(is_money,text_formatter,true,true)(value) : "",
```
Stephen-ONeil, line 112: Just noting that this will make values of zero empties. This is a tricky subject because empties get turned in to zeros when the datasets are read in... but I think we tend to just display them as zeros always because empties are more confusing/zero is usually technically correct still?

Also, the bar chart should have the same check added, saw some tables for bar charts full of stuff like "$NaN", :rofl:

client/src/components/DisplayTable.js
```diff
@@ -16,31 +16,27 @@ export class DisplayTable extends React.Component {
...
-        ({search_values}) => _.chain(search_values)
-          .keys()
-          .map( (search_key) => [search_key, ""] )
-          .value()
-      )
+    const searches = _.chain(column_configs)
+      .pickBy(col => col.is_searchable)
+      .keys()
+      .map(search_key => [search_key, ""])
       .fromPairs()
```
Stephen-ONeil, line 34: I think these three lines could be just `.mapValues( () => "" )`

client/src/models/results.js
```diff
@@ -190,7 +190,7 @@ class Indicator {
   get target_date(){
     const { target_month, target_year } = this;
     if( _.isNumber(target_month) && _.isNumber(target_year) ){
-      return `${months[target_month].text} ${target_year}`;
+      return new Date(target_year, target_month);
```
Stephen-ONeil, line 193: This whole date thing's rough, it makes the `_.isNumber(target_month) && _.isNumber(target_year)` case for `target_date` values a weird special case that every consumer of the Indicator target dates in the InfoBase will now have to write special case code to handle. 

`result_flat_table.js` has access to the raw indicator month and year values (try `_.pick(indicator, ['target_month', 'target_year'])`), can leave this code as it was and keep the special case requiring `Date` objects more contained.

client/src/components/DisplayTable.js
Stephen-ONeil: Definitely an interesting approach, works I think, but not one I would have thought of! Array will have blanks in it if an index is skipped by accident in the column config (I commented on a case like that further down, should check if that actually breaks anything in here).

I'd probably write it something like 
```
  _.chain(column_configs)
    .map( ({index}, key) => [index, key] )
    .sortBy( _.first )
    .map( _.last )
    .value()
```

client/src/panels/panel_declarations/finances/drr_dp_resources/dp_rev_split.js
```diff
@@ -46,53 +46,80 @@ export const declare_dp_rev_split_panel = () => declare_panel({
...
-        return {
-          display_values: display_values,
-          sort_values: sort_values,
+          gross: find_year_data_ends_with(year_data, "_gross"),
+          spa: find_year_data_ends_with(year_data, "_spa"),
+          rev: find_year_data_ends_with(year_data, "_rev"),
         };
       });
+      const amt_type = window.is_a11y_mode ? "compact1_written" : "compact1";
+      const column_configs = {
```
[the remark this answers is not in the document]
ktw1016: This was definitely a fluke

client/src/charts/nivo/nivo_bar.js
```diff
@@ -14,26 +14,35 @@ import {
...
-  const column_names = _.chain(ordered_column_keys)
-    .zip( _.concat([table_first_column_name || graph_text_maker("label")], keys) )
-    .fromPairs()
-    .value();
+  const column_configs = {
+    [indexBy]: {
+      index: 0,
+      header: table_first_column_name || graph_text_maker("label"),
+      is_searchable: true,
+      is_sortable: true,
```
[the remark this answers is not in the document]
ktw1016: Do you mean have only index 0 column searchable and sortable by default? Right now, they're all false by default

client/src/charts/nivo/nivo_line.js
```diff
@@ -93,19 +93,32 @@ export class NivoResponsiveLine extends React.Component {
...
-      .map( row=> ({
-        display_values: _.mapValues(
-          row,
-          (values, key) => key === 'label' ? values : get_formatter(is_money,text_formatter,true,true)(values)
-        ),
-        sort_values: row,
-        search_values: {label: row.label},
+      .map( row => ({
+        ..._.mapValues(row),
       }))
```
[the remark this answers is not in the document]
ktw1016: Oops, this one was a mistake

client/src/charts/nivo/nivo_line.js
ktw1016: Were you seeing $NaN for bar charts in this branch? I'm seeing values of zero, displayed as $0 for bar charts, and empties with line chart (I believe because of this check). Which should I be going for?

client/src/charts/nivo/nivo_bar.js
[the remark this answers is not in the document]
Stephen-ONeil: I guess only column 0 makes sense to sort by default, but I think any should be sortable by default?

client/src/charts/nivo/nivo_line.js
Stephen-ONeil: I think we've always gone with zeros, double check how the current master versions do it. I'd try and find the NaNs I was seeing, but I'm getting problems on the dev links. Not your fault, CORS rules are causing a problem; Google changed their default permissions a while back but I swore I fixed it... Maybe rebase from master and force push up? That might fix the dev links.

client/src/components/DisplayTable.js
ktw1016: ohhh ha I never thought of sorting like this. But this is definitely something I wanted! Thank you

client/src/panels/panel_declarations/finances/goco/goco.js
```diff
@@ -95,49 +95,80 @@ class Goco extends React.Component {
...
+      [sa_text]: {
+        index: 0,
+        header: sa_text,
+        is_searchable: true,
+      },
+      [spending_text]: {
+        index: 1,
+        header: spending_text,
+        is_sortable: true,
+        formatter: (value) => spend_map[value],
```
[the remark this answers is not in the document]
ktw1016: You're totally correct, I did goco at last and definitely tried to rushed through it too quickly. My bad

client/src/charts/nivo/nivo_line.js
ktw1016: Okay!

client/src/panels/panel_declarations/results/result_flat_table.js
ktw1016: I played around with it and tested it around, I think this one looks better with `large_status_icons` at all times. Thanks for letting me know with breakpoints though and putting this one on spotlight!

client/src/models/results.js
ktw1016: Okay, that was my initial direction as well but I decided to change the source `get` as `target_date` was only being used at 2 places. I'll contain this date thing within this.

client/src/charts/nivo/nivo_line.js
ktw1016: In master, bar is displayed as 0, line is displayed as empty (which is exactly the same case as this branch)

client/src/panels/panel_declarations/finances/vote_stat/in_year_vote_stat_breakdown.js
ktw1016: Hmm, I'm puzzled on `Dept` subject with `lookup_by_name` function though. What're you suggesting with this one?

PR #595, opened 2020-04-29
toggle all for graphlegend

### LINKED ISSUE #578: Add a button to show all graph items in the modal table even if some are toggled off in the legend
This is particularly for the graphs with tons of items in the legend and only one enabled by default, tedious to manually click them all on just to get them showing up in the table view

ktw1016:
closes #578


client/src/panels/panel_declarations/finances/detailed_program_spending_split/detailed_program_spending_split.js
```diff
@@ -124,6 +125,10 @@ class HistoricalProgramBars extends React.Component {
                   selected: _.toggle_list(selected, id),
                 });
               }}
+              onToggleClick={() => this.setState({
+                selected: selected.length < all_keys.length ? all_keys : [],
```
Stephen-ONeil, line 129: Most of the legend onClicks have a check to prevent the final item from being toggled off, right? What happens when no items are selected?

Stephen-ONeil, line 129: I believe all our legends usually stop the user from toggling off the final item. I don't remember if this was because graphs broke when they had no data or if we just didn't want them to be able to (since they would at least *look* broken). Is being able to have no selected items valuable? Does that clash with the normal behaviour too much?

client/src/panels/panel_declarations/finances/transfer_payments/historical_g_and_c.js
```diff
@@ -179,6 +179,7 @@ class DetailedHistTPItems extends React.Component {
     
     const raw_data = _.flatMap(graph_series, value => value);
 
+    const all_tp_idx = _.map(prepped_rows, (row, idx) => idx);
```
Stephen-ONeil, line 182: There's a lodash util called `range` that would give the same result by doing `_.range(prepped_rows.length)`, just a tip.

client/src/components/DisplayTable.js
```diff
@@ -20,7 +20,7 @@ export class DisplayTable extends React.Component {
       unsorted_initial,
     } = props;
 
-    const { sort_values } = _.first(rows);
+    const { sort_values } = !_.isEmpty() && _.first(rows);
```
Stephen-ONeil, line 23: Huh, thought trying to destructure `false` would possibly give an error, but at lest in the console this works. Learn something new every day!

client/src/panels/panel_declarations/finances/transfer_payments/historical_g_and_c.js
ktw1016: Ah, never knew this util existed. Noted, thanks!

ktw1016:
@Stephen-ONeil so I was aware that we don't allow de-selecting last item in our legends and that this functionality was de-selecting everything (including last item). To argue why returning to previous selections may be a bad idea, is that I only applied this feature on legends with lots of items. User could've already selected lots of items, and just want to de-select everything that they've selected. So I implemented it with that use case in mind. This doesn't break anything (as far as I know) and Nivo just disappears which I think looks fine.

I actually find this more intuitive, Select all, then de-select all when you click it again. And seems like how most of these "Select All" checkboxes work (At least this is how Excel works exactly and I find it intuitive)

Problem I find with this is that, it's inconsistent that we don't allow to de-select last item when user de-selects manually, but I'm allowing it here in this case. So 4 options are:
1. Leave it as is ("Select All" can de-select all, but user can't de-select last item)
2. Allow user to de-select last items at all times
3. "Select All" to return to previous selections
4. "Select All" de-selects everything except few items depending on graph

I'm voting on 2 because it seems like it's the most logical. (Click on a checkbox to de-select it)

Stephen-ONeil:
Ugh, Excel, haha.

Maybe make it two buttons then, "select all" and "deselect all", and have them always be present side-by-side. Saves a click if you have some but not all selected and want to go to having none selected, and makes the path to get there clear at all times. Would stand out more than a checkbox in a list of checkboxes too.

Letting the graph totally disappear feels bad, but I guess having hidden logic that sometimes stops you from deselecting an item is the same kind of secret behaviour that makes me grumpy about the checkbox. Is there any way to have no data but still render the graph axis?

ktw1016:
Microsoft spends millions on improving Office usability and I find it decent enough..! I think having two checkboxes just to select/deselect all is a bit intrusive and take too much space. And I think this is pretty standard usability, click select all to select it all, click it again to deselect it all.. I could make some dotted lines and paddings to make "Select All" clearer :)

Stephen-ONeil:
I don't think Microsoft's designers would use a checkbox for select all in their filter menu if they were designing it today (or at least de-selecting the checkbox would have an obvious behaviour). Microsoft's mixed blessing/curse is that they have a very deep legacy to all their systems, so we're all used to their weird design flaws (but also they have a really hard time changing anything).

Based on the web standards around what a checkbox control is and does, I would expect deselecting it to return you to your previous state; that's based on my understanding of form elements. Someone very familiar with Excel might expect it to have the Excel-like behaviour, as implemented. That's two different types of expert user though, we need to think about what the most baseline user would expect, and I have no idea what that is!

That's why I think a checkbox is a poor fit, it's got no clear behaviour on the de-select here. Two side by side buttons clearly labeled with what their behaviour is are the clearest UI I can think of. You're probably right that it would be too cluttered though, in which case I'd suggest a button that says "select all" when everything's not selected and "deselect all" when everything is selected. That would be the exact same behaviour as the current checkbox, but with no ambiguity to what would happen when it's clicked.

No need to keep going back on forth on this though, haha, we've probably both spent too long thinking about it now. If you're still unconvinced then I'll stop pushing!

client/src/panels/panel_declarations/finances/detailed_program_spending_split/detailed_program_spending_split.js
```diff
@@ -103,29 +103,44 @@ class HistoricalProgramBars extends React.Component {
...
+                color: colors(label),
+              }))
+              .value()
+            }
+            onClick={id => {!(selected.length === 1 && selected.includes(id)) &&
+              this.setState({
+                selected: _.toggle_list(selected, id),
+              });
+            }}
+            Controls={
```
ktw1016, line 123: How about making the redundant component (line 124 to 143) into a reusable component itself somewhere in `legends`, and pass in something like SelectAllOnClick and SelectNoneOnClick ?

Stephen-ONeil: Sounds good to me!

Stephen-ONeil: Want to do that before I re-review my self?

ktw1016: sure, I'll do that

client/src/charts/legends/SelectAllControl.js
```diff
@@ -0,0 +1,22 @@
...
+      <TrivialTM k="all"/>
+    </span>
+|
+    <span
+      onClick={SelectNoneOnClick}
+      className="link-styled" 
+      style={{marginLeft: "5px"}}
+    >
+      <TrivialTM k="none"/>
+    </span>
```
Stephen-ONeil, line 21: So, these spans-as-buttons are going to need some adjustments for a11y purposes.

Screen readers aside, the spans currently can't be used by keyboard navigators on the standard site. Spans aren't tab-focusable by default, and even if they were you wouldn't be able to properly fire the onClick from the keyboard.

If you want to keep them as spans you need to:
  - give them `tabIndex={0}`
    - explanation: tab index sets the local tabbing order if non-zero, but we rarely use that (order in the DOM's the better way to control tab order). When zero, it makes any element tab focusable and puts adds i to the default tabbing order. Negative tab index can be used to remove tab navigation from an element that has it by default
  - add `onKeyDown={ ({keyCode}) => (keyCode === 13 || keyCode === 32) && SelectAllOnClick()}`, etc.
    - keyboard navigation expects both enter (13) space bar (32) to work on a button

Alternatively (preferable) is to use buttons instead of spans here. Buttons are tab-navigable by default AND they automatically trigger their onClick when a space or enterkey is pressed on them. You could use the `button-unstyled` class and a bit of css tweaking to keep them looking like the spans currently do.

Not as important here because I think these controls won't be anywhere on the a11y build, but the span's would also need `role='button'` for proper screen reader semantics.

I think the ultimately correct approach would be something like the following, although I don't usually go all out with groups and labels (probably should be)
```
<div role="group" aria-labelledby={unique_id}>
  <div id={unique_id}>
     <TrivialTM k="select" />:
  </div>
  <button onClick={SelectAllOnClick}>
     <TrivialTM k="all" />
   </span>
|
  <button onClick={SelectNoneOnClick}>
     <TrivialTM k="none" />
  </button> 
</div>
```
With whatever styling you want on that

ktw1016: Ahh A11y😩

ktw1016: Thanks for the explanations!

PR #602, opened 2020-05-05
testing lighthouse ci

ktw1016:
Currently testing ci


ktw1016:
Hmm.. tests sometimes fail or successful. Running it locally was successful. Given that errors are timeout errors, failures are due to the docker image, probably incapable to run the tests fast enough. I might need to look for faster docker image?

Stephen-ONeil:
If it's a resource availability issue, changing the image is unlikely to help. It's a pretty lightweight image already, not a ton of other processes should be running during these tests. To actually get more resources available to our CI containers we'd have to upgrade our CircleCI plan.

I'll take a look at the error output and see if I've got a clue.

Stephen-ONeil:
The timeout's associated to `Method: Emulation.setScriptExecutionDisabled`, guessing a bit here but see if you can use the config to skip tests with disabled JS? We don't need to test with JS disabled anyway since that's just a page that says "please enable JS". Would be better to figure out what's actually wrong here, but that's at least a shot in the dark to try.

Weird that it would just be on the partition though, if it was related to disabled JS.

ktw1016:
> The timeout's associated to `Method: Emulation.setScriptExecutionDisabled`, guessing a bit here but see if you can use the config to skip tests with disabled JS? We don't need to test with JS disabled anyway since that's just a page that says "please enable JS". Would be better to figure out what's actually wrong here, but that's at least a shot in the dark to try.
> 
> Weird that it would just be on the partition though, if it was related to disabled JS.

I'll give it a shot, but go back to few commits back, you'll see partition passed before, also gov/people route passed, but failed after, etc... Results are going back and forth, hence why I believe it's resource problem (I guess circleci problem)

ktw1016:
```PROTOCOL_TIMEOUT refers to the situation when Lighthouse has been waiting for Chrome to respond to a single command for more than 30 seconds.```

**Comment by ktw1016** [comment]
and I couldn't find any config that allow us to do that

**Comment by Stephen-ONeil** [comment]
Try rerunning those older passing commits a few times from the CircleCI menu, they could have also been intermittently pass/failing if it's an inconsistent thing.

The CI images have Chromium, not Chrome, and likely a version from at least a few months back. Their internals are basically the same though. If you don't dig up any other ideas, to be through you could try updating chromium in the CI images and running with chromium locally instead of Chrome.

**Comment by Stephen-ONeil** [comment]
Well, the two failing runs I see on CI (not much data to work with) fail on different routes but fail both times from a timeout related to `Method: Emulation.setScriptExecutionDisabled`. No clue what the underlying cause is (could be resources, could be some flaky bug in the version of chromium we use, could be something else).

The log doesn't use the same names to refer to steps as the documentation, ugh, but I'm guessing that adding
```
"assertions": {
    "without-javascript": "off",
}
```
to the config will disable the audit step that's using ` Emulation.setScriptExecutionDisabled`. At least worth adding that and seeing if the error output changes.

May also be worth rerunning the job a handful more times in CI, just to collect some more data. If it's any sort of flaky test, resource based timeout or otherwise, more logs should give us something.

**Comment by ktw1016** [comment]
Where did you find `"without-javascript": "off"`? I couldn't find it anywhere in their docs. Anyways, I re-tried with that on and same failure. Just to answer the questions you listed, because more tests were ran in previous commits:
1. Tests are running in series
2. Always same error `setScriptExecutionDisabled`
3. Never the same routes.. Same route sometimes succeed or sometimes fail for unknown reasons
4. I can't draw anything in common between them, especially since the same route sometimes fail or have success

**Comment by ktw1016** [comment]
Error:
```
Runtime error encountered: Waiting for DevTools protocol response has exceeded the allotted time. (Method: Emulation.setScriptExecutionDisabled)
 
LHError: PROTOCOL_TIMEOUT
 
    at Timeout._onTimeout (/home/test-image-user/InfoBase/node_modules/lighthouse/lighthouse-core/gather/driver.js:409:21)
 
    at listOnTimeout (internal/timers.js:531:17)
 
    at processTimers (internal/timers.js:475:7)
```
Failed routes:
#compare_estimates
#partition/dept/exp
#orgs/gov/gov/infograph/people
#treemap

Note: All of the failed routes have also ran successfully as well.

**Comment by Stephen-ONeil** [comment]
> Where did you find `"without-javascript": "off"`? I couldn't find it anywhere in their docs. Anyways, I re-tried with that on and same failure.

Their doc's are a bit annoying, they don't list all the options but [here](https://github.com/GoogleChrome/lighthouse-ci/blob/master/docs/assertions.md#audits) they link to the default config, [here](https://github.com/GoogleChrome/lighthouse/blob/v5.5.0/lighthouse-core/config/default-config.js#L375-L407). 

Annnd now that I look closer, there's no `assertions` property in the default config, so I think the docs are out of date actually, ha.

**Comment by Stephen-ONeil** [comment]
Do the tests run in a consistent order? Are the ones it fails on always roughly the same number in, or is it, e.g., sometimes the second and sometimes the first? That might indicate if it's a resource depletion error or just a random flaky bug.

**Comment by ktw1016** [comment]
> Do the tests run in a consistent order? Are the ones it fails on always roughly the same number in, or is it, e.g., sometimes the second and sometimes the first? That might indicate if it's a resource depletion error or just a random flaky bug.

Tests run in a consistent order and it runs in order specified in `route_load_tests_config` (except I took out `treemap` to see if it was only that route causing the issue, which it wasn't)

**Comment by Stephen-ONeil** [comment]
`Emulation.setScriptExecutionDisabled` only shows up twice in their code base, in [this](https://github.com/GoogleChrome/lighthouse/blob/9e8ad2db893ecbfc184353450935bbba1fe4a811/lighthouse-core/gather/driver.js) file. Looks to be the core module for their "gatherers" (what visits the pages and runs collects the diagnostics). It's used when visiting a page with JS disabled, but ALSO used when exiting from a promise that's gone over the max timeout.

The error log doesn't make it clear which of the two it is. If we can figure out how to disable the JS disabled tests then we'd know for sure that it's an error that get throws in the process of a promise timing out (which could be resource exhaustion or an even deeper error somewhere else that freezes things up).

**Comment by Stephen-ONeil** [comment]
Sorry, not ANY promise timing out, didn't read the context that function was in. It's specifically the timeout promise used when a page fails to be considered "loaded" (uses a heuristic since "loaded" is undefined in general, looks for things like network and CPU going idle) after `maxWaitForLoadedMs`.

... damn inconsistent naming, had to trace back through three renames of the same value, but `maxWaitForLoadedMs` can be set in the config file as
```
{
  settings: {
    maxWaitForLoad: /*value in milliseconds*/,
  }
}
```

Seems to default to either 45000 or 35000 in different default configs. Try boosting that number up and see if it give the pages time to load?

Issue #597, opened 2020-04-30
integrate prettier


AlexCLeduc:
Here's what we need before we call this done and start formatting everything

1. an npm command to format everything, we'll need to run this at least once for each active branch. This can either be separate client/server/email or just one 🤷 
    * There should be a similar "dry-run" command that we can run to make sure our code already passes formatting
2. a strategy to minimize the impact of merge conflicts on existing PRs. I'm hoping if we run the above "format everything" command on master and the PR separately, then we won't get any worse conflicts than what we started with. Note that rebasing becomes difficult so we can either squash or just have plain merges
3. a pull-request level CI step that "fails" when the dry-run formatting command returns non-null changes
4. VSCode integration

Note that it's popular to use git hooks for this kind of stuff, but I'd prefer we _don't_ use a pre-commit hook because it becomes impossible to commit staged changes without also commiting unstaged changes in the same files. I like to tinker with commits too much for this not to piss me off

ktw1016:
Can this be closed?

AlexCLeduc:
🎉

PR #603, opened 2020-05-05
Prettier! 

AlexCLeduc:
supports #597

you'll want to add this in your _workspace_ settings (`infobase/.vscode/settings.json`) 

```json
{
    "editor.formatOnSave": true
}
```

 and install the prettier extension in vscode: `esbenp.prettier-vscode`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,15 +26,24 @@ jobs:
       - restore_cache:
           keys:
             - email-backend-dependencies-{{ checksum "email_backend/package-lock.json" }}
+            - top-level-dependencies-{{ checksum "package-lock.json" }}
       - run: 
           command: |
             [ -e "email_backend/node_modules" ] || (cd email_backend && npm ci)
+      - run: 
+          command: |
+            [ -e "node_modules" ] || npm ci
       - save_cache:
           paths:
             - email_backend/node_modules
             - scripts/common_node_scripts/node_modules
           key: email-backend-dependencies-{{ checksum "email_backend/package-lock.json" }}
+      - save_cache:
+          paths:
+            - node_modules
+          key: top-level-dependencies-{{ checksum "package-lock.json" }}
 
+      - run: npm run check_format email_backend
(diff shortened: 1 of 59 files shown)


AlexCLeduc:
Tested CI, it works. [Here's a passing commit](https://github.com/TBS-EACPD/infobase/commit/2fa7bcc39feff078339708bd06b249be1712a356) with all files formatted.


I added the check in CI and made it fail right before tests are run. I think it's worth it to check formatting in CI, because it'll break more often than we think.
It made the `test_server` and `test email_backend` step a little bit longer, but not by much. The `test_end_to_end` step's duration is too volatile for the `client/` format check addition to really matter.

a couple notes

- checking `email_backend/` is part of `test_email_backend`
- checking `server/` is part of `test_server`
- checking `client/` is part of `test_end_to_end`
- ignore rules are all in the top-level `.prettierignore` , maybe we should distribute these accross the app like we do with eslint, but we'd need a top-level one anyway 🤷

Stephen-ONeil:
TODO:
  - [ ] read the docs my self, should try and tweak any rules I might want to before going through the work of getting this on all the open branches
  - [ ] https://prettier.io/docs/en/options.html#arrow-function-parentheses

AlexCLeduc:
actually, since we're only formatting JS, you may want to use 

```
  "[javascript]": {
    "editor.formatOnSave": true
  }

```

instead of top-level  

```
"editor.formatOnSave": true
```

This just screwed me over in the python project when prettier formatted a yaml file and gave me an unexpectedly huge diff (And I forgot to use workspace settings)

Stephen-ONeil:
> This just screwed me over in the python project when prettier formatted a yaml file and gave me an unexpectedly huge diff (And I forgot to use workspace settings)

I'm going to drop .vscode from the root level .gitignore and commit up the appropriate, JS only, `editor.formatOnSave` rule. One less manual dev env set up task, and will save everyone the sort of headache you got from putting it in their global settings.

Issue #604, opened 2020-05-06
department DRR donut labels messed up 


AlexCLeduc:
![Screen Shot 2020-05-05 at 8 41 43 PM](https://user-images.githubusercontent.com/7769215/81128993-d7bda480-8f10-11ea-9a32-af3f5372d040.png)

Stephen-ONeil:
Huh, hadn't seen that before, but rebasing a commit that references an issue seems to keep re-logging the reference.
![image](https://user-images.githubusercontent.com/11301919/81229148-5c65fc80-8fbd-11ea-81b7-e0c7baa2958f.png)
(screen shot in case this is a caching thing or GitHub retroactively figures out that only one of those references still exists)

PR #605, opened 2020-05-06
DRR Summary panel review

Stephen-ONeil:
Oof.

TODO:
  - [x] title's using the wrong grammar, and needs to be different between dept and gov panels
  - [x] panel has too many sections with no delineation (at the gov level at least), add horizontal lines to break things up (but also consider trimming some content or breaking this in to multiple panels, the full detailed table is weird to put in a "summary" panel anyway... pretty sure these WERE separate panels and someone merged them?)
  - [x] #604
  - [x] the instructions "Click below to change which results are counted in the full set" are totally disjointed from the "counts" they refer to, that second section needs a better layout
  - [x] icon array is trash on mobile, not using fcols right, gets crushed and overlaps with text

Edit: to move things along, moved some of my bullets to a holding issue, #606


Stephen-ONeil:
Haha, set myself up for a couple merge conflicts again #595 (even if they aren't apparent, I've for sure used a CSS class here that's deprecated by #595, but I'll sorted that out once #595 merges).

PR #614, opened 2020-05-12
FAQ route

Stephen-ONeil:
TODO:
  - [x] convert remaining content of google docs files to `faq.yaml`
  - [x] if the page is long enough, add an index at the top (meaning also give the route a second argument for the key of the question to auto-scroll too)
  - [x] style tweaks (~doesn't need much, going to be a pretty basic route~ ok, maybe a bit to do)
  - [x] aha, good opportunity to remove (some) of the redundant items from the About page, make that route more focused! Edit: wasn't really that much I could justify deleting in the end, ah well
  - [x] move Q/A data to a csv, coordinate with the pipeline on them taking that file over
  - [x] slight tangent, but refactor `FancyUL` to be less of a point of confusion
  - [x] hmm, rethink using `LabeledTable`? Maybe something less fancy would be more suitable. Not sure I actually love the questions being to the left of the answers, vertical layout may be better.
      - eh, convinced my self to stick with the labeled table for now. No one else seems to feel it's a poor fit
      - [x] ~double check a11y of `LabeledTable`, is it properly screen-reader navigable? Definitely has tab-navigation issues~ should be much better now
      - [x] ... but still needs more a11y testing, really need to get our hands on more screen reader options
  - [x] do we really want that team email in the FAQ on contact? I thought we'd hesitated to publish that in the past for some reason. **EDIT:** filter out the contact Q for now, but leave it the data so I can toggle it back on once confirming that's really the email we want to give
  - [x] write route description,  translate new text fragments, etc.


client/src/InfoBase/App.js
```diff
@@ -162,11 +165,12 @@ export class App extends React.Component {
               />
               <Route path="/rpb/:config?" component={ReportBuilder} />
               <Route path="/about" component={About} />
+              <Route path="/faq/:selected_qa_key?" component={Faq} />
               <Route
                 path="/compare_estimates/:h7y_layout?"
                 component={EstimatesComparison}
               />
-              <Route path="/privacy" component={PrivacyStatement} />
+              <Route path="/privacy" component={PrivacyStatement} />s
```
AlexCLeduc, line 173: I didn't realize routes supported pluralization. Is that french friendly, though? 😆

Stephen-ONeil: Frick, always screwing up my quick save hotkey, good catch!

client/src/Faq/Faq.yaml
```diff
@@ -0,0 +1,209 @@
+faq_page_title:
+  en: Frequently Asked Questions
+  fr: TODO
```
AmndaPlant, line 3: I believe the accepted french translation is "Foire aux questions"

client/src/Faq/Faq.scss
```diff
@@ -0,0 +1,17 @@
+@import "../common_css/common-variables";
+
+ul.fancy-ul.faq-index {
+  & > li {
+    border-color: $primaryColor;
+    padding: 0px 15px;
+
+    &:first-child {
```
ktw1016, line 8: I think these could be generalized into fancyul scss for potential future use. This is basically fancy ul with title

client/src/Faq/Faq.js
```diff
@@ -0,0 +1,89 @@
...
+];
+
+const FaqIndex = () => (
+  <FancyUL ul_class="faq-index">
+    {[
+      <TM
+        key="jump_to_question"
+        k="jump_to_question"
+        el="h2"
+        className="heading-unstyled"
```
ktw1016, line 39: Is this better than just
`<TM style={{fontSize: x}}/>` ?
I guess h2 is fixed font size so it's more consistent

Stephen-ONeil: I'm clobbering the `h2` styling with "heading-unstyled". This is more about proper semantic HTML/a11y. Distinct sections/subsection of a route should have (appropriately hierarchical) heading elements because screen readers generally have some form of heading-based navigation (jump to next heading at level/jump to next subheading, etc.).

A real hassle to have well managed headings on more dynamic pages (although I mostly make the effort), but straightforward for a page like this.

Stephen-ONeil:
Definitely better as a csv long term, but I talked my self back and forth between csv and yaml for a while and eventually just went with the quickest one to get to production with (yaml, so I wouldn't have to coordinate with the pipeline). Doesn't seem like there's really a rush for this though so I could always slow my roll and work out the csv version now, yeah.

client/src/Faq/Faq.js
ktw1016: ahh I see!

client/src/Faq/Faq.scss
Stephen-ONeil: Oh, hmm, didn't notice because the CSS is buried in `site.scss`, but there's already a class for `fancy-ul-title-row`. The `FancyUL` styling's a bit of a mess, I think it was pre-react, so the react component for it is pretty slapped together/the styles aren't co-located with it.

I think the `FancyUL` component should be moved from `misc_util_components` to its own file, its styling should be plucked from `site.scss`, and then the whole thing should be refactored a little. Can make a consistent way of adding a title to it at that point.

Bit out of scope for this PR, but I might slip it in since I pretty much thought through how to do that in the process of writing this comment, haha. Don't really want to open more issues right now anyway.

PR #618, opened 2020-05-12
copy/download raw data on DisplayTable

### LINKED ISSUE #561: Panel table views should include a button to copy the data to clipboard
Filing an issue instead of doing this in #500 because I don't want to delay that merge any longer.

### LINKED ISSUE #585: List of new feature request issues for DisplayTable
Implement the features referenced below

ktw1016:
closes #561 
closes #585 

- [x] download csv button


ktw1016:
For departments where we used their dept id, this is where we get in trouble. We can't just use formatter to format all the values because numbers will then have commas everywhere. Having another boolean in `column_configs` would be too messy. Any ideas?

ktw1016:
Also I suggest, instead of copying to clipboard, make it a download to csv (preferable for RPB.. and I think it makes more sense?)

Stephen-ONeil:
I can think of two many cases where one is preferable to the other, depending on what the user wants, haha. Probably we should just do both!

client/src/components/DisplayTableUtils.scss
```diff
@@ -0,0 +1,30 @@
+@import '../common_css/common-variables';
+
+.heading-container {
```
Stephen-ONeil, line 3: Class names in this file are fairly generic, could they reference display table in some way?

client/src/components/DisplayTableUtils.js
```diff
@@ -0,0 +1,21 @@
+import "./DisplayTableUtils.scss";
+
+import { WriteToClipboard } from "./WriteToClipboard.js";
+import { IconCopy } from "../icons/icons.js";
+
+export class DisplayTableUtils extends React.Component {
```
Stephen-ONeil, line 6: For organization's sake, I'd recommend defining this component in `DisplayTable.js` and not exporting it (doesn't seem like it would have much reason to be used in any other context). Alternatively, create a `/DisplayTable` sub-dir to the `/components` dir and group this and display table there.

ktw1016: yeah I think I'd rather make the sub directory to separate it because I expect this component to get larger and I'd rather make `DisplayTable.js` cleaner

client/src/components/DisplayTableUtils.js
```diff
@@ -0,0 +1,21 @@
...
+    const { data_to_csv_string } = this.props;
+
+    return (
+      <div className="heading-container">
+        <WriteToClipboard
+          text_to_copy={data_to_csv_string}
+          button_class_name={"heading-utils"}
+          IconComponent={IconCopy}
+          icon_color={window.infobase_color_constants.backgroundColor}
+        />
```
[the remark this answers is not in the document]
ktw1016: haha that's actually what I originally did, but thought another yaml file just for this wouldn't be  worth it. But considering possibly more text to come, I can add that in !

ktw1016:
1px was intended haha I'm sure people wouldn't like it, I'll just remove that
TBH, for `raw_formatter`, we can re-purpose `search_formatter` config, and use it as `raw_formatter`

client/src/components/DisplayTableUtils.scss
ktw1016: I don't think so, I changed them up however

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -149,10 +152,24 @@ export class DisplayTable extends React.Component {
...
+      )
+    );
+
     return (
       <div
         style={{ overflowX: "auto", marginTop: "20px", marginBottom: "20px" }}
       >
+        <DisplayTableUtils
+          table_name={table_name}
+          data_to_csv_string={d3.csvFormatRows(data_csv)}
```
Stephen-ONeil, line 171: Small nitpick, but based on the name I'd expect `data_to_csv_string` to be a function, as it sounds like it's describing an action. Just `csv_string` feels like a clearer prop name.

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -149,10 +152,24 @@ export class DisplayTable extends React.Component {
...
 
+    const data_csv = [ordered_column_keys].concat(
+      _.map(sorted_filtered_data, (row) =>
+        _.map(ordered_column_keys, (key) =>
+          col_configs_with_defaults[key].raw_formatter
+            ? col_configs_with_defaults[key].raw_formatter(row[key])
+            : row[key]
+        )
+      )
+    );
```
Stephen-ONeil, line 163: Huh, I must have missed this before, but `ordered_column_keys` is an array of the headers, not the keys? Or are we using the plain language headers as keys?

If you set a default `raw_formatter` of `_.identity` then you could skip the ternary check.

client/src/components/DisplayTable/DisplayTableUtils.js
```diff
@@ -0,0 +1,48 @@
...
+      "data:text/csv;charset=UTF-8," + encodeURIComponent(data_to_csv_string);
+
+    const temporary_anchor = document.createElement("a");
+    temporary_anchor.setAttribute(
+      "download",
+      `${table_name ? table_name : "table"}.csv`
+    );
+    temporary_anchor.setAttribute("href", uri);
+    temporary_anchor.dispatchEvent(new MouseEvent("click"));
+  }
```
Stephen-ONeil, line 23: Please test in IE, been a while but I have a vague memory that this approach might not work there. If it doesn't work in IE then don't spend too long trying to find something that does, we can just hide the download button on IE, the copy to clipboard covers the functionality well enough.

client/src/components/DisplayTable/DisplayTableUtils.scss
```diff
@@ -0,0 +1,29 @@
+@import '../../common_css/common-variables';
+
+.dp-utils-heading-container {
+  display: flex;
+  flex-direction: row-reverse;
```
Stephen-ONeil, line 5: Row reverse is a little annoying, because the order in JSX will be reversed from the order on the live site. `justify-content: flex-end` will also push the util icons to the right side, and it won't reverse their order in the process.

client/src/components/DisplayTable/DisplayTable.js
ktw1016: ah good point

client/src/components/DisplayTable/DisplayTable.js
ktw1016: ohh `ordered_column_keys` is keys, as name suggests

client/src/components/DisplayTable/DisplayTableUtils.js
ktw1016: Yup, I'll remove it for IE. It doesn't work

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -149,10 +153,22 @@ export class DisplayTable extends React.Component {
       .map(_.last)
       .value();
 
+    const data_csv = [ordered_column_keys].concat(
```
Stephen-ONeil, line 156: I'm still questioning about this line. If `ordered_column_keys` is keys, then aren't you using internal/uni-lingual keys as the headers of the downloaded/copied csv? I only checked one or two and they looked like they used the correct plain language headers, which is why I asked if we were using the plain language column names as keys.

ktw1016: Oh shoot, I misinterpreted your question earlier. You're totally right, I'm putting keys into CSV as headers, I actually didn't even notice that, I guess I got way too familiar with these DisplayTable keys

Stephen-ONeil: Ah! Ok, I was asking it the other way around. I need to double check the case I tested because I swear the headers WERE plain language and not keys, haha.

Stephen-ONeil: Oh, damn, right. I was looking at graph tables, and for dumb nivo reasons we DO use plain language names as keys in those. The other tables would have been more obvious, yeah.

ktw1016: Oh haha, maybe Kate changed them around? I THINK they were keys ever since I started working on DisplayTable (back when I implemented search filter). Anyways, I pushed changes to fix this one

Issue #599, opened 2020-05-01
Nivo cleanup


Stephen-ONeil:
Wanted to do this immediately following #530 merging, but I've got other priorities in the short term.

I started the cleanup in #530, but left it unfinished for the sake of keeping that PR focused. To start with:
  - [ ] rename our wrapped nivo components, they've all got names like `NivoResponsiveBar`. At a minimum, we should drop responsive from all of their names. We might want to start using `WrappedNivo___` instead, to clarify that these are our own wrapped nivo charts, and not the originals. I did that with renaming the directory to `wrapped_nivo`, although I'm very open to a better naming convention than that
  - [ ] rename all the modules in `client/src/charts/wrapped_nivo` so their names follow code standards and match the component they export (after their component are renamed)
  - [ ] improve `get_formatter` in `client/src/charts/chart_utils.js`, at least give it named arguments so we aren't always double checking and messing up what the booleans it takes do. Preferably,  refactor it or even rethink the whole approach, it's a mess...
  - [ ] lots of old or messy code throughout, do a code standards/clean up sweep
  - [ ] and the big part, do a lot of thinking about how to standardize the APIs of our wrapped graphs. We could do so much better than we currently do...

PR #619, opened 2020-05-13
Fix DRR table percentages

### LINKED ISSUE #612: Percentages showing as infinities in table for gov DRR summary pie chart
image

AmndaPlant:
closes #612 
image

The code responsible for creating the table percentage values is this snippet of code from the NivoResponsivePie component:

```
const table_data = _.map(data, (row) => ({
    label: row.label,
    percentage: row.value / legend_total,
    value: row.value,
}));
```

`legend_total` is calculated using the `legend_data` prop to the component. DRR summary was never passing this prop to its NivoResponsivePie so there was a division by zero. Passing in the `legend_data` prop fixes the percentage calculation however this has the added effect of putting a legend below the pie charts on results pages.
image

If this is a problem it shouldn't be too hard to change NivoResponsivePie to use the graph data instead of the legend data for calculating table data.

### DIFF
--- a/client/src/charts/wrapped_nivo/nivo_pie.js
+++ b/client/src/charts/wrapped_nivo/nivo_pie.js
@@ -32,6 +32,7 @@ export class NivoResponsivePie extends React.Component {
       display_horizontal,
       disable_table_view,
       table_name,
+      show_legend,
     } = this.props;
 
     const color_scale = infobase_colors_smart(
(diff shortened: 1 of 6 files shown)


ktw1016:
Ohhh, hmm I think Kate intentionally left out `legend_data` to add the separate legend because she doesn't wanna show the percentages in the `TabularLegend`

ktw1016:
I think, remove passing in `legend_data` props for `drr_summary`, and calculate `percentage` for `table_data` from `data`

AmndaPlant:
Yeah calculating the percentage using the `data` prop instead of `legend_data` should work. To my knowledge everywhere else that uses NivoResponsiblePie passes the same object for both `data` and `legend_data` so it shouldn't affect any other pages.

Stephen-ONeil:
If there was ever a use case for having separate `data` and `legend_data` props, it's not occurring to me now.

I'd recommend fully dropping the `legend_data` prop right now, since it's currently redundant to the `data` prop in use and any attempts to use `legend_data` in the future would potentially be bugged after this change (using the `data` total for the legend's denominator would likely not be expected if there _was_ a reason for the data to be different, and a future attempts to "fix" that would then regress this fix).

Stephen-ONeil:
It's possible someone wanted to do something like collapse all slices smaller than x% in to one "other" slice on the actual chart, while still displaying them as individual items on the legend. I guess that'd be one use case for distinct graph and legend data, but I don't think it's a worthwhile one. I'm sticking to my suggestion of deprecating `legend_data`, leaving would just be a pitfall for future devs, haha.

ktw1016:
deprecating `legend_data` entirely means the `TabularLegend` will show up. I remember Kate did not want to show the percentages in the legend because it was misleading. Deprecating it means we need to either:
1. Have a boolean flag to show/hide TabularLegend panel by panel
2. Remove `StandardLegend` and make TabularLegend more configurable (non-solidBox and hiding percentages)

I wouldn't mind either

AmndaPlant:
I added a prop named `show_legend` which is just a simple boolean value used to conditionally render the `TabularLegend`

Stephen-ONeil:
I think the legend toggling prop is good for now. Moving the use of `TabularLegend` out of the pie chart component is a good next step but is more in-line with the general Nivo cleanup issue #599 (being a matter of consistency across all charts). Out of scope to this PR.

AmndaPlant:
I don't know why I didn't even think of doing that.

PR #621, opened 2020-05-14
hotfix for rpb displaytable column removal error

### LINKED ISSUE #620: DisplayTable in RPB, error on removal of sorted column
Worked this out from a prod error logged to analytics, so some urgency to fixing it!

Easy to reproduce:
  - go to the RPB and enable two columns of data
  - sort by one of the two
  - disable the currently sorted column
  - crash :(

ktw1016:
fixes #620

### DIFF
--- a/client/src/components/DisplayTable.js
+++ b/client/src/components/DisplayTable.js
@@ -71,7 +71,7 @@ export class DisplayTable extends React.Component {
       */,
     } = this.props;
     const { sort_by, descending, searches } = this.state;
-
+    console.log(column_configs);
     const column_config_defaults = {
       is_sortable: true,
       is_summable: false,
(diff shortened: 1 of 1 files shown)


client/src/components/DisplayTable.js
```diff
@@ -71,7 +71,7 @@ export class DisplayTable extends React.Component {
       */,
     } = this.props;
     const { sort_by, descending, searches } = this.state;
-
+    console.log(column_configs);
```
Stephen-ONeil, line 74: Stray `console.log`

client/src/components/DisplayTable.js
```diff
@@ -111,7 +111,7 @@ export class DisplayTable extends React.Component {
           .value()
       )
       .thru((unsorted_array) => {
-        if (sort_by) {
+        if (col_configs_with_defaults[sort_by]) {
```
Stephen-ONeil, line 114: Could also be `_.has(col_configs_with_defaults, sort_by)`. In this case, I think they're equivalent, just pointing out because using `_.has` to check existence of a key on an object can be useful in other cases (e.g. if the value of the key itself might be falsey, but all you care about is if the object has the key). Maybe worth using here just because it is a better representation of that the check actually cares about (the existence of the column, not the value of the config).

PR #625, opened 2020-05-20
Expand/collapse all

### LINKED ISSUE #615: result drill down Expand all button?
Could be handy to have expand all button for result drill down!

AmndaPlant:
closes #615 

Adds an expand/collapse all button to the result drill down. There may potentially be a way to optimize the `toggle_all` action as on drill downs with more elements it can take a second or two for all of them to open. The button could also potentially use some styling.

### DIFF
--- a/client/src/explorer_common/state_and_memoizing.js
+++ b/client/src/explorer_common/state_and_memoizing.js
@@ -70,8 +70,6 @@ function root_reducer(state = initial_root_state, action) {
       const { root } = payload;
       const shouldExpand = !payload.is_expanded;
 
-      const { userExpanded: oldExpanded, userCollapsed: oldCollapsed } = state;
-
       let nodes = [];
       let search_nodes = (parent) => {
         if (parent.children.length === 0) {
--- a/client/src/explorer_common/state_and_memoizing.js
+++ b/client/src/explorer_common/state_and_memoizing.js
@@ -66,7 +66,6 @@ function root_reducer(state = initial_root_state, action) {
     }
 
     case "toggle_all": {
-      // TODO
       const { root } = payload;
       const shouldExpand = !payload.is_expanded;
 
--- a/client/src/explorer_common/state_and_memoizing.js
+++ b/client/src/explorer_common/state_and_memoizing.js
@@ -70,21 +70,21 @@ function root_reducer(state = initial_root_state, action) {
       const shouldExpand = !payload.is_expanded;
 
       let nodes = [];
-      let search_nodes = (parent) => {
-        if (parent.children.length === 0) {
+      let search_nodes = (source) => {
+        if (source.children.length === 0) {
           return;
         }
-        _.forEach(parent.children, (child) => {
+        _.forEach(source.children, (node) => {
           if (shouldExpand) {
-            if (!child.isExpanded) {
-              nodes.push(child.id);
+            if (!node.isExpanded) {
+              nodes.push(node.id);
             }
           } else {
-            if (child.isExpanded) {
-              nodes.push(child.id);
+            if (node.isExpanded) {
+              nodes.push(node.id);
             }
           }
-          search_nodes(child);
+          search_nodes(node);
         });
       };
       search_nodes(root);
@@ -92,7 +92,7 @@ function root_reducer(state = initial_root_state, action) {
       if (shouldExpand) {
         return { ...state, userExpanded: nodes, userCollapsed: [] };
       } else {
-        return { ...state, userCollapsed: nodes, userExpanded: [] };
+        return { ...state, userExpanded: [], userCollapsed: nodes };
       }
     }
 
--- a/client/src/explorer_common/state_and_memoizing.js
+++ b/client/src/explorer_common/state_and_memoizing.js
@@ -95,6 +95,14 @@ function root_reducer(state = initial_root_state, action) {
       };
     }
 
+    case "clear_expanded_collapsed": {
+      return {
+        ...state,
+        userExpanded: [],
+        userCollapsed: [],
+      };
+    }
+
     default: {
       return state;
     }
@@ -132,6 +140,10 @@ const map_dispatch_to_root_props = (dispatch) => {
       payload: { node },
     });
 
+  const clear_query = () => dispatch({ type: "clear_query" });
+
+  const enable_loading = () => dispatch({ type: "enable_loading" });
+
   const expand_all = (root) =>
     dispatch({
       type: "expand_all",
@@ -144,9 +156,10 @@ const map_dispatch_to_root_props = (dispatch) => {
       payload: { root },
     });
 
-  const clear_query = () => dispatch({ type: "clear_query" });
-
-  const enable_loading = () => dispatch({ type: "enable_loading" });
+  const clear_expanded_collapsed = () =>
+    dispatch({
+      type: "clear_expanded_collapsed",
+    });
 
   return {
     set_query,
@@ -155,6 +168,7 @@ const map_dispatch_to_root_props = (dispatch) => {
     enable_loading,
     expand_all,
     collapse_all,
+    clear_expanded_collapsed,
   };
 };
 
--- a/client/src/explorer_common/state_and_memoizing.js
+++ b/client/src/explorer_common/state_and_memoizing.js
@@ -65,32 +65,38 @@ function root_reducer(state = initial_root_state, action) {
       }
     }
 
-    case "toggle_all": {
+    case "expand_all": {
       const { root } = payload;
-      const shouldExpand = !payload.is_expanded;
-      const { userExpanded: oldExpanded, userCollapsed: oldCollapsed } = state;
+      const { userExpanded: oldExpanded } = state;
 
       const get_ids_to_update = ({ isExpanded, id, children }) => [
-        (shouldExpand && !isExpanded) || (!shouldExpand && isExpanded)
-          ? id
-          : [],
+        !isExpanded ? id : [],
         _.map(children, get_ids_to_update),
       ];
       const ids_to_update = _.flattenDeep(get_ids_to_update(root));
 
-      if (shouldExpand) {
-        return {
-          ...state,
-          userExpanded: oldExpanded.concat(ids_to_update),
-          userCollapsed: [],
-        };
-      } else {
-        return {
-          ...state,
-          userExpanded: [],
-          userCollapsed: oldCollapsed.concat(ids_to_update),
-        };
-      }
+      return {
+        ...state,
+        userExpanded: oldExpanded.concat(ids_to_update),
+        userCollapsed: [],
+      };
+    }
+
+    case "collapse_all": {
+      const { root } = payload;
+      const { userCollapsed: oldCollapsed } = state;
+
+      const get_ids_to_update = ({ isExpanded, id, children }) => [
+        isExpanded ? id : [],
+        _.map(children, get_ids_to_update),
+      ];
+      const ids_to_update = _.flattenDeep(get_ids_to_update(root));
+
+      return {
+        ...state,
+        userExpanded: [],
+        userCollapsed: oldCollapsed.concat(ids_to_update),
+      };
     }
 
     default: {
@@ -130,10 +136,16 @@ const map_dispatch_to_root_props = (dispatch) => {
       payload: { node },
     });
 
-  const toggle_all = (root, is_expanded) =>
+  const expand_all = (root) =>
     dispatch({
-      type: "toggle_all",
-      payload: { root, is_expanded },
+      type: "expand_all",
+      payload: { root },
+    });
+
+  const collapse_all = (root) =>
+    dispatch({
+      type: "collapse_all",
+      payload: { root },
     });
 
   const clear_query = () => dispatch({ type: "clear_query" });
@@ -145,7 +157,8 @@ const map_dispatch_to_root_props = (dispatch) => {
     toggle_node,
     clear_query,
     enable_loading,
-    toggle_all,
+    expand_all,
+    collapse_all,
   };
 };
 
--- a/client/src/explorer_common/state_and_memoizing.js
+++ b/client/src/explorer_common/state_and_memoizing.js
@@ -68,28 +68,28 @@ function root_reducer(state = initial_root_state, action) {
     case "toggle_all": {
       const { root } = payload;
       const shouldExpand = !payload.is_expanded;
+      const { userExpanded: oldExpanded, userCollapsed: oldCollapsed } = state;
 
-      let nodes = [];
-      let search_nodes = (source) => {
-        if (source.children.length === 0) {
-          return;
-        }
-        _.forEach(source.children, (node) => {
-          if (
-            (shouldExpand && !node.isExpanded) ||
-            (!shouldExpand && node.isExpanded)
-          ) {
-            nodes.push(node.id);
-          }
-          search_nodes(node);
-        });
-      };
-      search_nodes(root);
+      const get_ids_to_update = ({ isExpanded, id, children }) => [
+        (shouldExpand && !isExpanded) || (!shouldExpand && isExpanded)
+          ? id
+          : [],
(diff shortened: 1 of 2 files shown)


AmndaPlant:
Further thinking about it I'm not sure if there is an easy way to optimize the `toggle_all` action. 

From my understanding drill down nodes are stored in a tree of objects, starting at `root`. Each node has an array known as `children` which simply stores all the child nodes. `search_nodes()` is a preorder depth first search implementation which has a time complexity of O(n) where n is the number of nodes. To my knowledge there aren't any tree traversal algorithms that have a better time complexity, therefore without redoing the data structure nodes are stored in (which would be both complicated and I'm not quite sure any data structure besides a tree makes sense for this use case) it can't be optimized in any meaningful way.

It's not a *huge* issue, it just means it takes a second or two for all nodes to expand on pages with more nodes in their drill down (National Defense is an example of what I mean). It's entirely possible I'm missing something so if anyone has any ideas that'd be great!

Stephen-ONeil:
Nice theoretical analysis! To be honest though, algorithmic complexity is rarely the pain point in client-side performance (awesome that you considered it though, it's still totally something worth keeping in mind). 

With these things, it tends to be rendering and DOM interaction. My gut's that a good chunk of the delay is from all the enter/exit transitions on all the explorer nodes.

Try digging in to the performance and React profiler tabs of the dev tools (only available if you have the React Dev Tools extension installed). I took a look my self, and it seems my gut was only partially correct. The information available in those tabs are pretty dense, and also pretty vague in places, but can still give solid leads on the real issues. Let me know when you have so we can compare findings, I've got a couple ideas for next steps based on my own digging!

Note: 
  - Scripting performance will often be significantly worse in dev builds. Try running the performance checks in a local prod build or on the dev link for a less skewed view of scripting vs rendering times. You'll also want to run performance checks in the dev build though, so that the links back to offending source code are actually useful/readable.
    - Some of the performance difference is because the tree shaking and other bundle size optimizations for production reduce the amount of JS that needs to be parsed, but a lot of it is probably also the difference between dev-React and prod-React (for one thing, dev mode React runs way more correctness checks while prod mode React has all the warnings pulled out). It's possible some babel transformations inadvertently optimize as well (I know that the babel arrow function transformation originally produced code that performed better than native arrow functions, but I'm sure JS engines quickly improved their implementations).
  - Less dramatic, but just having the developer tools open also has an impact on performance. That's usually kind of good though, since developer machines tend to be over-powered. You might even want to use the Chrome dev tools emulation to artificially slow things down further, to see just how bad it can get!


I'll do a quick code review shortly!

client/src/explorer_common/state_and_memoizing.js
```diff
@@ -71,6 +71,34 @@ function root_reducer(state = initial_root_state, action) {
...
+          if (
+            (shouldExpand && !node.isExpanded) ||
+            (!shouldExpand && node.isExpanded)
+          ) {
+            nodes.push(node.id);
+          }
+          search_nodes(node);
+        });
+      };
+      search_nodes(root);
```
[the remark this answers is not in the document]
Stephen-ONeil: I left it in for the sake of it being a refactor, but the "search" part, i.e. `(shouldExpand && !isExpanded) || (!shouldExpand && isExpanded) ? id : []`, is actually an implementation bug.

If you expand any nodes before you click "expand all" then the expanded nodes end up collapsed, and vice versa. Guessing you interpreted `userExpanded`/`userCollapsed` as the ids of nodes that the current action wants to update, but they're actually become the new state rather than merging/applying to it.

Any confusion there's understandable though, the alternative interpretation does a better job of explaining the apparent redundancy between the two arrays. Track both user expanded and user collapsed nodes was Alex's solution to the problem that nodes a) can be open OR closed by default and b) the default can vary between nodes.

client/src/panels/panel_declarations/results/result_drilldown/result_drilldown.js
```diff
@@ -189,7 +189,7 @@ const get_col_defs = createSelector(
 class SingleSubjExplorer extends React.Component {
   constructor() {
     super();
-    this.state = { query: "" };
+    this.state = { query: "", is_expanded: false };
```
Stephen-ONeil, line 192: Having the expand/collapse button be stateful is a can of worms. Adds complexity to the code and the site, and actually creates a usability deadzone IMO.

For site complexity, it's an issue of the number of possible interacting states. Depending on your path, you can end up with the same drilldown state but either button state. That's tricky for users and is also the sort of interplay that can leave more room for future bugs.

For the "usability deadzone", imagine a user clicks the button to expand everything, then starts collapsing things as they read, but later decides they want every thing expanded again. They can go back to the button they remember clicking, but it's been replaced by a collapse all. There's a lot of branches their behaviour could take at this point, but none of them are ideal (best case they figure out to click the button twice, but that's a pain even if the performance can be improved).

A possible fix is to sync the buttons state to the drilldowns state (e.g. if ANY nodes are still collapsed, the button acts as expand all; only collapses all if all are open). Pros: would at least make the state consistent (less room for state related bugs, better for users who can intuit the hidden rules). Cons: more code (and thus more room for bugs in general) and still a overly complex UX. 

My suggested fix is to split the behaviour in to two stateless, always present, buttons. One to expand all, one to collapse all. Sure, more controls on screen, but much less complexity for both users and developers (and the users always have a clear path to either option).

client/src/explorer_common/state_and_memoizing.js
```diff
@@ -71,6 +71,40 @@ function root_reducer(state = initial_root_state, action) {
...
+        _.map(children, get_ids_to_update),
+      ];
+      const ids_to_update = _.flattenDeep(get_ids_to_update(root));
+
+      return {
+        ...state,
+        userExpanded: [],
+        userCollapsed: oldCollapsed.concat(ids_to_update),
+      };
+    }
```
AmndaPlant, line 106: I'm thinking this block can be cleaned up with a helper function to avoid repeating a lot of shared code between expanding and collapsing all.

Stephen-ONeil: Sounds good to me!

client/src/panels/panel_declarations/results/result_drilldown/result_drilldown.js
```diff
@@ -309,6 +311,38 @@ class SingleSubjExplorer extends React.Component {
...
+                className="btn btn-ib-light"
+                style={{
+                  height: "40px",
+                  width: "50%",
+                  margin: "5px 0px",
+                }}
+                onClick={() => {
+                  this.setState({ loading_query: true });
+                  expand_all(root);
+                  this.setState({ loading_query: false });
```
Stephen-ONeil, line 325: First, if you're not familiar with the JS event loop architecture, I recommend googling about it and reading an article or two. Plenty of well written resources to explain it better than I can, and enough that I won't try to recommend a specific one. Just pick any that seem suited to your learning style and let me know if you have questions. It's not the direct culprit here, but it's an important player in all this. Going to be categorizing a lot of code as synchronous or asynchronous in both my explanation of what's gone wrong and in my suggestions for a fix.

Long story short, you're getting bit by a relatively new (as of React 16, the current major version number) feature, so it's definitely possible that myself or any of the tutorials you followed are to blame for glossing over this special case! You've probably been taught (by me?) that a `setState` immediately prompts a rerender... but there's currently ONE case where that's not true: inside of a React event handler!

State/prop ipdates occurring in the call stack of a React event handler are instead asynchronously batched for later, and if multiple prop/state changes for a single component fall in to the same batch queue than they're merged and applied as a single update. So, an event callback can only result in one rerender. Here, not only does the first `setState` end up clobbered by the second, but all the synchronous work done by `expand_all(root)` still needs to execute on the JS thread before `onClick` gets off the call stack and the asynchronous batched-render even has a chance to start!

For more nitty gritty on batched renders: https://stackoverflow.com/questions/48563650/does-react-keep-the-order-for-state-updates/48610973#48610973

The plan is, I believe, to eventually make ALL React rendering batch and occur asynchronously by default. Good for performance, but probably also a steeper learning curve for the whole framework... but it will be better than the current position, where it's super easy to be unaware of which `setState`s will synchronously re-render and which are going to be batched in to an asynchronous render.

So, options to subvert this "helpful" batching? 

One option: fight async with async
```
onClick={() => {
  this.setState({ loading_query: true });
  setTimeout(() => {
    expand_all(root);
    this.setState({ loading_query: false });
  }, 0);
}}
```
Moves `expand_all(root)` and `this.setState({ loading_query: false })` out of the `onClick` call stack and throws them on the JS message queue (even a timeout of 0 creates async code). Lets `onClick` resolve and the immediate async render for `this.setState({ loading_query: true })` occur before the the JS thread gets around to expanding nodes and removing the spinner. 

Complication 1: it's complicated. Requires all developers touching this code to understand the JS event loop and know what React's up to with batching renders caused by event handlers. There are enough timeouts flying around this file because of the debounced search input. If you don't dig up another solution, use this but add comments on the "why" of it.

Complication 2: need to do some book keeping to make sure timeouts clear when the component dies. See the `componentWillUnmount` lifecycle method, it's already tracking and killing timeouts related to the debounced search input. This is mostly because if `setState` is called on a dead component you get a lot of annoying console warnings about potential memory leaks, but it's also important because sometimes those warnings are right and there really is a memory leak.

Stephen-ONeil:
I mentioned this in the general comment of an earlier review, but this bug still needs fixing:

> Code comments aside, found a bug while testing the site. Might be a bug that already existed in the explorer code but was only surfaced by this new feature.
> 
> Steps to reproduce:
> 
> 1. press "expand all"
> 2. switch to a different results document/tab of the panel
> 3. press collapse all

I'll do another code review in a moment though!

client/src/panels/panel_declarations/results/result_drilldown/result_drilldown.js
```diff
@@ -309,6 +314,42 @@ class SingleSubjExplorer extends React.Component {
...
+                type="button"
+                className="btn btn-ib-light"
+                style={{
+                  height: "40px",
+                  width: "50%",
+                  margin: "5px 0px",
+                }}
+                onClick={() => {
+                  this.setState({ loading_query: true });
+                  this.expandTimout = setTimeout(() => {
```
Stephen-ONeil, line 327: Please add a comment here and below summarizing why this `setTimeout` is required.

I think this was in the wiki somewhere, but the rule of thumb for comments in the InfoBase is to avoid "what" comments, but to always include "why" comments as necessary. 

When code requires a comments to describe **what** the code is doing, then the code wasn't written clearly enough. Well named variables and functions (especially when you're writing side effect free functional-style code) should "self document". A) it forces us to write clearer, more maintainable code, and B) it avoids the (HUGE\*) problem of documentation comments and code getting out of sync. 

\* trust me, out of date comments are far worse than no comments, the InfoBase massively suffered from this problem back when I started.

The one defensible use of "what" comments are for APIs, but even then I'm personally not a huge fan. Inaccurate API documentation is a bug, and it's ambiguous whether the code or the comments are wrong (not to mention the consumers).

**Why** comments are for cases like this. For the sake of maintainability, we try not to write hacky/unobvious/esoteric code. When there's no avoiding it, usually for equally unobvious/esoteric reasons, and we _don't_ document those reasons, then (best case) we're setting up future developers to duplicate the entire discovery process we went through to determine that this was the only/best option. Worst case, they think past developers were just writing bad code, they reimplement this the straight forward way, and then the original bug regresses (they may not even notice). Basically, if the "why" behind an implementation doesn't boil down to something like "I wrote it this way because it seemed obvious" then it likely needs a comment.

Stephen-ONeil: ... like the `timedOutStateChange` in the query of this component. Vaguely feel like that was originally something I wrote. Don't remember _why_, and I'm wishing past me had document the importance of it. Not clear it's necessary, but I don't want to waste the time right now pulling it apart just to rediscover what it was there for.

AmndaPlant:
I just have a few confusions about the bugfix that I'd like clarity on.

Firstly I'm not sure what this entire block of code is doing
 ```js
//union the SYMMETRIC differences
const potential_to_toggle = _.union(
  _.difference(oldState.root.userCollapsed, state.root.userCollapsed),
  _.difference(state.root.userCollapsed, oldState.root.userCollapsed),
  _.difference(oldState.root.userExpanded, state.root.userExpanded),
  _.difference(state.root.userExpanded, oldState.root.userExpanded)
);

// IMPORTANT: the current set of nodes, oldFlatNodes, depends on the explorer scheme,
// which changes independantly of the explorer root. This means that userCollapsed and
// userExpanded state stored in root is not guaranteed to reflect the current set of nodes...
// This is a pretty fragile point in the current explorer design, only surfaced because of
// the expand/collapse all buttons (previously, there was no way to add or remove old scheme
// ids from the userCollapsed/userExpanded sets, so the symmetric difference always removed them).
// Something to avoid in an explorer refactor, but not something that could properly be fixed with out,
// essentially, a whole explorer refactor (or manual clearing implemented by each explorer at whatever
// point they update thier internal schemes... but the explorer has too much boiler plate already)
const safe_to_toggle = _.intersection(
  potential_to_toggle,
  _.map(oldFlatNodes, "id")
);
```

I'm also not really sure what this line is doing either
```js
set_doc !== doc && clear_expanded_collapsed() && set_doc(doc, subject);
```

Stephen-ONeil:
Going to restate some things first, just so I can refer to stuff with confidence that we're on the same page.

The root of the problem is that each explorer has two chunks of redux state, the state from `explorer_common`, called (...confusingly) `root`, and the state from the specific explorer implementation, called `scheme`. `userExpanded` and `userCollapsed` are stored in the `root` state and, largely, managed by internal `explorer_common` code. The `scheme` state is entirely defined and managed by the explorer implementation and, aside from a few required properties that `explorer_common` code expects to exist in any `scheme` state, it can be whatever it wants. The sticky point is that the set of nodes displayed in the explorer is totally based off of `scheme` and implementation-specific logic and there's nothing to keep `scheme` and `root` in sync. As a result, `userExpanded` and `userCollapsed` frequently contain ids for nodes that no longer exist.

Marrying these two distinct states would probably turn in to a huge rewrite of all explorer code. Not the time to do that, want to move this PR along.

---
RE: first block of code

`potential_to_toggle` used to be `toToggle`. I didn't change the right hand side of that expression, just the name. The first `_.difference` returns ids that have been removed from `userCollapsed` by the current state update, the second line returns ids that have been added to `userCollapsed` by the current state update. Repeat for `userExpanded`.

`oldFlatNodes`, at this point in the code, is a flat array of all nodes currently rendered by the explorer. `safe_to_toggle` is `potential_to_toggle` filtered to only include ids that correspond to existing nodes. This way, the code that attempts to toggle the state of individual nodes won't receive invalid ids. 

Filtering the ids prevents this whole class of explorer bug, but **doesn't** guarantee correct behaviour. E.g. if you expanded all nodes, switched to a new set of nodes, and then hit collapse all you wouldn't crash anymore but you _would_ potentially see some nodes _expand_ as a result. Specifically, any nodes with ids that were used for both the previous set of nodes and the current set of nodes could still slip through with unexpected results.  

---
RE: second block of code

The addition of `clear_expanded_collapsed()` to the `setDoc` line in `results_drilldown.js` represented an alternate fix. Simply put, it's a manual step to clear the parts of the `root` state that really should be clearing any time the underlying set of nodes changes.

This fix results in correct behaviour, crash-free. I was unsatisfied with relying on this alone though. Each explorer implementation is totally free to have code that can result in a change to the set of displayed nodes anywhere it wants, so this fix relies on developers writing and maintaining each explorer to always remember to use the `clear_expanded_collapsed` dispatcher as needed. The explorers already have way too many rules and obscure pieces of boilerplate to re-apply in each implementation.

Stephen-ONeil:
All that said, I'm going to push up another commit to clarify (out at least spell check) the comment I left above the `safe_to_toggle` line. Was re-reading it now and it's a bit of a jumble. Will also throw a warning when that code filters out dead ids, so that it isn't silently hiding bad behaviour.

PR #628, opened 2020-05-22
Improve dev link death

### LINKED ISSUE #572: Improve dev link death
Dev links lose their DB once their corresponding branch is merged/deleted, but their static content otherwise keeps existing (they become zombie sits that work up until you visit a page that needs GraphQL API data). That's not helpful to anyone!

The static content should also be killed, and maybe for maximum helpfulness the index html files should be replaced with links explaining that dev links are regularly cleaned (and maybe directing visitors to the master dev link, as presumably the features they were looking for are now merged and available there).

Stephen-ONeil:
Closes #572


ktw1016:
Hmm, have you thought of just re-directing them to master dev link?

Stephen-ONeil:
There was a few things I wanted that led to this approach:
  1) I wanted to know right away if I miss-typed a dev link URL, so I didn't want to take the easy route of totally deleting dead dev links and just setting any specific index file as the bucket's default html content (which is an option, although technically I'd be setting it as the 404 error page)
  2) I specifically wanted anyone visiting an old dev link to know that, whatever they thought they were doing, they'd missed their chance. Being put in a panic because someone vaguely tells me the site's broken, only to learn they were just using a stale dev link was a big reason for setting this up

I didn't think about redirecting to master directly at any point. Any easy way of doing it would have broken with #1 above though, but I guess I could do it now by sticking some JS in the stand in html I've rolled out. Any automatic redirects interfere with #2 though, for the sake of not wasting anyone's time I really want anyone visiting a stale link to know that they're late to the party.

If you haven't seen it, the page looks like this:
image
So it does at least include a link to take them to the master dev build, if that's what they want.

Stephen-ONeil:
Ah damn, just noticed the "it's has" typo... Updating this is a pain, because if nothing else pushing all the files up to all the old dev link directories takes like 30 min, haha.

PR #637, opened 2020-06-05
Align dollar columns to the right

GarvinH:
closes #633 
Added style const.
Add right text align to table data with numbers only.

image
image


GarvinH:
I wasn't able to test it on IE. Sorry!

PR #656, opened 2020-06-12
Lighthouse CI

ktw1016:
Check out LHCI server here: https://infobase-lhci.herokuapp.com/app/projects

Will need to work on #650 after this merges in


scripts/ci_scripts/lighthouserc.js
```diff
@@ -0,0 +1,23 @@
...
+      url: _.map(
+        route_load_tests_config,
+        ({ route }) =>
+          `https://dev.ea-ad.ca/${process.env.CIRCLE_BRANCH}/index-eng.html#${route}`
+      ),
+    },
+    upload: {
+      target: "lhci",
+      serverBaseUrl: "https://infobase-lhci.herokuapp.com/",
+      token: "554e44e8-3fcb-46ba-8444-9cb54850bc72", // could also use LHCI_TOKEN variable instead
```
ktw1016, line 20: `The build token is used by CI to upload new data to your server. It's not really very secret, especially if you'd like to evaluate the performance of external contributor contributions.`

Stephen-ONeil: We don't run CI for external contributor fork-PRs though, because of our other more-secret CI env vars. Might as well reissue this token and move it to an CircleCI env var. Worst case, someone could use it to DOS our heroku, or even just feed our lighthouse bad data, right? Would be a petty prank, and I don't expect anyone would, but keeping it secret's not that much more work.

Stephen-ONeil:
Oh! You could also add a link to the dashboard in either the root or client readme file, so that there's an easy place to find the link when we need it. If there's some API that would give us an overall performance metric, maybe we could look at adding a status badge for it too, that would be a handy reminder to work on improving things.

ktw1016:
Yeah just `Procfile` and `server.js`, well also the new dependencies in root folder

So just to clarify, do you mean moving those two files from root into sub-dir, then just move them out to root folder locally only when we re-deploy, then move them back into sub-dir?

ktw1016:
Pushed the changes as how I understand it !

.circleci/config.yml
```diff
@@ -370,4 +383,7 @@ workflows:
           requires:
             - test_end_to_end
             - deploy_server
-          <<: *deploy_filter
\ No newline at end of file
+          <<: *deploy_filter
+      - lighthouse:
+          requires:
+            - deploy_client
```
Stephen-ONeil, line 389: Oh wait, one more thought, would it be fine to only run this job on the master branch? It's still 10+ minutes, right? Don't want it clogging up CI.

PR #666, opened 2020-06-18
Fix CI dev link cleanup

Stephen-ONeil:
That was dumb, I'd renamed a variable but missed an instance.


Stephen-ONeil:
Self-approving, this was just a trivial one line fix in the end.

PR #667, opened 2020-06-18
hotfix for results previous year error

### LINKED ISSUE #663: Some 2020-21 indicators not showing last year's target properly
This is affecting an indicator for Statistics Canada named "Percentage of users satisfied with statistical information". They use this indicator at the CR level, as well as multiple times at the program level. The indicator(s) also existed in DP 2019-20, with a minimum target of 80%. The 2019 indicators show up in InfoBase as well as the CSVs. For example, at the CR level both the 2019 and 2020 versions of the indicator have a `stable_id` of  "DR-2027" in the CSV. 

image

ktw1016:
fixes #663

### DIFF
--- a/server/src/models/results/populate.js
+++ b/server/src/models/results/populate.js
@@ -53,14 +53,12 @@ export default async function ({ models }) {
       const doc_index = _.indexOf(time_period_ordered_doc_keys, doc);
       if (doc_index > 0) {
         // Look for corresponding indicator, by stable_id, from the previous documents. Pick most recent previous version
-        const previous_year_indicator_instance = _.chain(indicator_records)
-          .filter(
-            (indicator) =>
-              indicator.stable_id === stable_id &&
-              _.indexOf(time_period_ordered_doc_keys, indicator.doc) < doc_index
-          )
-          .first()
-          .value();
+        const previous_year_indicator_instance = _.find(
+          indicator_records,
+          (indicator) =>
+            indicator.stable_id === stable_id &&
+            indicator.doc === time_period_ordered_doc_keys[doc_index - 1]
+        );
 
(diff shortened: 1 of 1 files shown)


server/src/models/results/populate.js
```diff
@@ -59,7 +59,7 @@ export default async function ({ models }) {
               indicator.stable_id === stable_id &&
               _.indexOf(time_period_ordered_doc_keys, indicator.doc) < doc_index
           )
-          .last()
+          .first()
```
Stephen-ONeil, line 62: Ugh, wrote a nice long comment, but lost it.

Looking at this file, it was always relying on the order of the indicators within the csv. That was sloppy on my part, the sorting coming out of the pipeline must have reversed at some point (or this code was always wrong, doesn't mean it wasn't sloppy though).

If the sort ordering within the csv changes again, this will go back to being wrong. Shouldn't assume that either `first` or `last` is meaningful unless we've sorted the data ourselves first.

Code aside, the logic here is also wrong (unrelated to this bug, but worth correcting). If an indicator was dropped one year and revived later, this would still pick out the most recent historical copy. I may have figured that might be desired, but we call the fields "previous_year..." in the DB and label it as "Last year's ..." on the client, so this current logic is incorrect.

This is untested, but we probably want something like this
```
const previous_year_indicator_instance = _.find(
  indicator_records,
  (indicator) =>
    indicator.stable_id === stable_id &&
    indicator.doc === time_period_ordered_doc_keys[doc_index - 1]
)
```

`time_period_ordered_doc_keys[doc_index - 1]` is safe because this code's only reachable if `doc_index > 0`

ktw1016: Ah actually this was the first solution I thought of, I guess it's safe

PR #671, opened 2020-06-22
Remove reduce population by from titles

### LINKED ISSUE #655: Reduce use of "Population by ____" in people graph titles
UX testers recommended that recommended that the titles be front loaded with the content of what each particular graph is displaying. In other words, titles should be reworded to avoid all titles beginning with "Population by ___".

AmndaPlant:
Closes #655 

Removes the term "population by" from panel titles.

Haven't made the change for the french version yet. I did also notice a few other yaml files but I couldn't figure out exactly where they were being used so for now I didn't change them.

### DIFF
--- a/client/src/tables/orgEmployeeGender.js
+++ b/client/src/tables/orgEmployeeGender.js
@@ -27,13 +27,13 @@ export default {
   },
 
   name: {
-    en: "Population by Employee Gender",
-    fr: "Population selon le sexe",
+    en: "Employee Gender",
+    fr: "Sexe",
   },
 
   title: {
-    en: "Population by Employee Gender",
-    fr: "Population selon le sexe",
+    en: "Employee Gender",
+    fr: "Sexe",
   },
 
   add_cols: function () {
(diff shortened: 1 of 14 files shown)


Stephen-ONeil:
RE: `Table` class stuff. These are fairly legacy at this point. A full switch to GraphQL will almost certainly mean stripping out all the ancient `Table` related code. A fair bit of the GraphQL work will be rewriting things like the datasets page, the RPB, parts of the panel infrastructure, etc. because they're all heavily based off of `Table` code at the moment.

AmndaPlant:
The other instances are found in the `Table` data, I think changing the text *could* be beneficial to ensure consistency however I don't have a super strong opinion either way

client/src/tables/orgEmployeeGender.js
```diff
@@ -27,13 +27,13 @@ export default {
   },
 
   name: {
-    en: "Population by Employee Gender",
-    fr: "Population selon le sexe",
+    en: "Employee Gender",
+    fr: "Sexe",
```
Stephen-ONeil, line 31: Double check french with someone, should it also mention "employee"? 

English should be "sex" as well, in my opinion, but we don't have to change that immediately.

PR #672, opened 2020-06-24
Service Inventory

ktw1016:


GarvinH:
#673 I've made the service list using the FancyUL for the time being

GarvinH:
#680 Have the service filter/legend done

GarvinH:
#681 the new service list

AlexCLeduc:
The department infographic is super sharp, and we’ve finally got real data! 

A couple of comments


## service overview panel

Screen Shot 2020-07-10 at 5 07 56 PM

This is super busy. There are aesthetic issues as well as readability. For example, the minus signs mean no SIN/CRA BN ?  That's not really clear. 

I think most of the information (aside from the half-pie) can be presented in one of these tables (or is it a `<dl>`? I don't remember). I totally realize the irony, it was my idea to copy their dashboards 😆 Might be nice to make a whole new panel so we can show the service folks side-by-side what we changed about their original vision. 
Screen Shot 2020-07-10 at 5 13 01 PM



## Digital status panel

Is there a way to disable the search bar? I feel like it should be turned off by default when we have less than 10 rows or something like that. 
Would also be nice if we had a single word to summarize all those sentences. The sentences are absolutely necessary to clarify things, but power users are gonna have a hard time scanning through this table 


Screen Shot 2020-07-10 at 5 49 20 PM




## Service standards 

The results-style filter is kind of giant for what it does. I think we only have met/not-met, and maybe not-available. Maybe a simple `<select>` dropdown would work and take up less space? 
Screen Shot 2020-07-10 at 5 21 32 PM

ktw1016:
Near scope:
- [ ] (Alex) Make a separate overview panel imitating our `About` bubble
- [ ] External Link icon
- [ ] Overview panel consistency
- [ ] Different viz for standards performance gauge
- [ ] (Stephen) Service to move in between People and Results bubble
- [ ]  Single service communication to bar chart without percentages, also comments into footnotes
- [ ] program ids bug
Later scope:
- [ ] (Ross) Bubbles getting cluttered
- [ ] (Ross) Separate control for filtering by service types in provided services list

client/src/panels/panel_declarations/services/single_service_panels/service_standards.js
```diff
@@ -0,0 +1,167 @@
...
+    };
+  }
+
+  render() {
+    const { service } = this.props;
+    const { active_statuses } = this.state;
+    const { result_simple_statuses } = businessConstants;
+    const standards = service.standards;
+
+    const footnotes = _.chain(standards)
```
ktw1016, line 35: Not sure if this would be enough for Footnotes (same for service_digital_status)

server/src/models/services/populate.js
```diff
@@ -2,56 +2,100 @@ import _ from "lodash";
...
+      ),
+      sin_collected: convert_to_bool_or_null(sin_collected, "Yes", "No"),
+      service_report_comment,
+      ...other_fields,
+    })
+  );
+  const standard_report_rows = _.map(
+    get_standard_csv_file_rows("standard-report.csv"),
+    ({ is_target_met, comment: standard_report_comment, ...other_fields }) => ({
+      is_target_met: convert_to_bool_or_null(_.toInteger(is_target_met), 1, 0),
```
ktw1016, line 43: not sure if we should just ask Ross to put 1/0 or true/false for these. Also specifically this one might make sense to put as `met` or `not_met` since we use those keys to index icons (`services/shared.js`)

client/src/panels/panel_routes/SingleServiceRoute.js
```diff
@@ -0,0 +1,77 @@
...
+import { Service } from "../../models/services";
+import { Subject } from "../../models/subject";
+import { infograph_href_template } from "../../link_utils.js";
+import {
+  ServiceOverview,
+  ServiceOverviewV2,
+  ServiceChannels,
+  ServiceStandards,
+  ServiceDigitalStatus,
+} from "../panel_declarations/services";
```
ktw1016, line 14: Not sure if this current structure is okay, or can be structured better?

Stephen-ONeil:
Dev link broken? Seems the dev API function doesn't understand the `has_services` query on an org.

PR #673, opened 2020-07-02
Service list

GarvinH:
Formats service inventory services into lists with description and service type

### DIFF
--- a/client/src/panels/panel_declarations/services/provided_services_list.js
+++ b/client/src/panels/panel_declarations/services/provided_services_list.js
@@ -90,6 +90,14 @@ const ProvidedServicesListPanel = ({ panel_args }) => (
         num_of_services: panel_args.services.length,
       }}
     />
+    <input
+      aria-label={text_maker("explorer_search_is_optional")}
+      className="form-control input-lg"
+      type="text"
+      style={{ width: "100%", marginBottom: "1rem" }}
+      placeholder={text_maker("filter_results")}
+      onChange={(evt) => null}
+    />
     <FancyUL>
       {_.map(
         panel_args.services,
@@ -106,7 +114,7 @@ const ProvidedServicesListPanel = ({ panel_args }) => (
             >
               <div className="tag-badge">{service_type}</div>
               <a href={`#dept/${org_id}/service-panels/${id}`}>
-                See this service.
+                See this service
               </a>
             </div>
           </React.Fragment>
--- a/client/src/panels/panel_declarations/services/provided_services_list.js
+++ b/client/src/panels/panel_declarations/services/provided_services_list.js
@@ -6,7 +6,6 @@ import {
   declare_panel,
 } from "../shared.js";
 import { FancyUL } from "../../../components";
-import { filter } from "lodash";
 
 import { FancyUL } from "../../../components";
 
@@ -135,7 +134,6 @@ const ProvidedServicesListPanel = ({ panel_args }) => (
 >>>>>>> 9fa26043... Display service as a tag
 =======
 
-    console.log(panel_args.services);
     const filtered_services = _.filter(panel_args.services, (service) => {
       if (
         _.includes(service.name.toLowerCase(), this.state.query.toLowerCase())
--- a/client/src/panels/panel_declarations/services/provided_services_list.js
+++ b/client/src/panels/panel_declarations/services/provided_services_list.js
@@ -6,6 +6,7 @@ import {
   declare_panel,
 } from "../shared.js";
 import { FancyUL } from "../../../components";
+import { filter } from "lodash";
 
 const { text_maker, TM } = create_text_maker_component(text);
 
--- a/client/src/panels/panel_declarations/services/provided_services_list.js
+++ b/client/src/panels/panel_declarations/services/provided_services_list.js
@@ -5,6 +5,7 @@ import {
   InfographicPanel,
   declare_panel,
 } from "../shared.js";
+import { FancyUL } from "../../../components";
 
 import { FancyUL } from "../../../components";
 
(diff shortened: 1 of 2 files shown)


GarvinH:
image

ktw1016:
Hey Garvin, good start!! I was talking about this in Slack but I thought I should probably leave the comments here so it's more trackable.

1. Service types into something like `tag-badge`. I think service name should be more highlighted than service type. So it'd be nice if service type is a bit more subtle
2. I want it to be a bit more obvious that you can click to see more details on individual services. Possibly another button like `See this service` at the bottom right corner, something like this would be nice!
3. We'll need to be able to search through these services! Search input to filter by service name.
4. Speaking of filtering, filter by service types as well? :) I think this one can go into issues and be implemented later

GarvinH:
image

image

GarvinH:
I need to make "see this service" in french too

client/src/panels/panel_declarations/services/provided_services_list.js
```diff
@@ -6,25 +6,80 @@ import {
...
+    super(props);
+    this.state = { query: "" };
+  }
+
+  render() {
+    const { panel_args } = this.props;
+
+    const filtered_services = _.filter(panel_args.services, (service) => {
+      if (
+        _.includes(service.name.toLowerCase(), this.state.query.toLowerCase())
```
ktw1016, line 22: I personally prefer to declare a variable if object needs to be referenced multiple times like
`const { query } = this.state` or `const { services, subject } = panel_args`. This is optional personal preference though

I think variable name `query` can be a bit more specific also, like `service_query`?

ktw1016: Also it'd be a bit cleaner to my taste without any if statements
`_.filter(services, service => _.includes(name) || _.includes(type)`

client/src/panels/panel_declarations/services/provided_services_list.js
```diff
@@ -6,25 +6,80 @@ import {
...
+            subject_name: panel_args.subject.name,
+            num_of_services: panel_args.services.length,
+          }}
+        />
+        <input
+          aria-label={text_maker("explorer_search_is_optional")}
+          className="form-control input-lg"
+          type="text"
+          style={{ width: "100%", marginBottom: "1rem" }}
+          placeholder={text_maker("filter_results")}
```
ktw1016, line 51: I think placeholder text can be a bit more specific, like `Filter services by name or type`?

GarvinH:
image

French service type has bad input

client/src/panels/panel_declarations/services/provided_services_list.js
```diff
@@ -6,25 +6,80 @@ import {
...
+                <div
+                  style={{
+                    display: "flex",
+                    fontSize: "14px",
+                    justifyContent: "space-between",
+                  }}
+                >
+                  <div className="tag-badge">{service_type}</div>
+                  <a href={`#dept/${org_id}/service-panels/${id}`}>
+                    See this service
```
ktw1016: Should go into yaml file!

client/src/panels/panel_declarations/services/provided_services_list.js
```diff
@@ -6,25 +6,80 @@ import {
...
+                <p>{description}</p>
+                <div
+                  style={{
+                    display: "flex",
+                    fontSize: "14px",
+                    justifyContent: "space-between",
+                  }}
+                >
+                  <div className="tag-badge">{service_type}</div>
+                  <a href={`#dept/${org_id}/service-panels/${id}`}>
```
ktw1016, line 70: You should try styling this and see how it looks, I was originally thinking of styling it with something like `btn-ib-primary`

client/src/panels/panel_declarations/services/provided_services_list.js
```diff
@@ -6,25 +6,80 @@ import {
...
+
+    return (
+      <div>
+        <TM
+          k="list_of_provided_services_desc"
+          args={{
+            subject_name: panel_args.subject.name,
+            num_of_services: panel_args.services.length,
+          }}
+        />
```
ktw1016, line 45: A bit more separation between this text and input would be nice

ktw1016:
> image
> 
> French service type has bad input

Ah this seems like data issue, don't mind these, I'll look into it with real data once I get that set up

client/src/panels/panel_declarations/services/provided_services_list.js
```diff
@@ -1,30 +1,80 @@
...
-import {
-  create_text_maker_component,
-  InfographicPanel,
-  declare_panel,
-} from "../shared.js";
+import { InfographicPanel, declare_panel } from "../shared.js";
+import { FancyUL } from "../../../components";
 
-const { text_maker, TM } = create_text_maker_component(text);
+const text_maker = create_text_maker(text);
```
ktw1016, line 9: any reasons why you needed to import `TM` separately?

client/src/panels/panel_declarations/services/provided_services_list.js
```diff
@@ -1,30 +1,80 @@
...
+                <a href={`#dept/${org_id}/service-panels/${id}`}>{name}</a>
+                <p>{description}</p>
+                <div
+                  style={{
+                    display: "flex",
+                    fontSize: "14px",
+                    justifyContent: "space-between",
+                  }}
+                >
+                  <div className="tag-badge btn-ib-primary">{service_type}</div>
```
ktw1016, line 65: ah I actually meant `btn-ib-primary` style for `See the service`. My bad on not being too clear on that haha

client/src/panels/panel_declarations/services/provided_services_list.js
GarvinH: Whoops. I guess I had a different idea of what the method did earlier haha

PR #675, opened 2020-07-03
Searching by other language acronym

### LINKED ISSUE #592: Allow searching by other language dept acronym
For example, a French user might want to search by DND because it's a very common short hand, or a bilingual user might use one or the other out of habit.

This is slightly complicated because you actually *can* already search for DND and get the expected result in the French version. I believe (unconfirmed) that's because the PAS code is DND, and I think we allow matching on PAS codes. Not super common to have the PAS code match the english acronym, but happening even just for DND is enough for it too feel like a bug when a different english acronym *doesn't* work.

Just need to get `other_lang_acronym` in the unilingual pipeline outputs (and as a field in the API for consistency? sure) and add it to the search matching fields.

... well we're at it, the matched items should include the acronym you matched on if the match was exact, like I currently do for old names.

AmndaPlant:
Closes #592 

Allows for searching by other language acronym, applied title and legal title.

Originally the plan was just to allow searching by acronym and applied title although a good chunk of departments have the applied title blank as they only use their legal title so the decision was made to add `other_lang_legal_title` as well.


ktw1016:
I tried searching few with `other_legal_title` and it doesn't seem to work. Am I missing something, or have you tested it works?

AmndaPlant:
Nice catch, I forgot to update `populate_stores` with the new field 😅

PR #677, opened 2020-07-06
Fix Gross and Net Planned Spending

### LINKED ISSUE #674: Gross Planned Spending improperly calculated
In the "Gross and Net Planned Spending" panel(?), gross planned spending, planned revenues, planned spending (net) and perhaps Planned Specified Purpose Account Spending are being improperly calculated. The values that appear in InfoBase should reflect the values that appear in the CSV input (program_spending.csv)

|  CSV variable | InfoBase lablel  |
|---|---|
planning_year_X | Planned Spending (net)
planning_year_X_rev |  Planned Revenues
planning_year_X_spa | Planned Specified Purpose Account Spending
planning_year_X_gross | Gross planned spending

So in below, 2020-21 planned revenues should be 16599536

image

GarvinH:
closes #674 

image

Taken from JUS. I've verified the numbers for year 1 of spa, gross, and rev and now seem to be correct (that is, it now matches program_spending.csv).

It seems like the header was put in the wrong place rather than there being a miscalculation.

### DIFF
--- a/client/src/panels/panel_declarations/finances/drr_dp_resources/dp_rev_split.js
+++ b/client/src/panels/panel_declarations/finances/drr_dp_resources/dp_rev_split.js
@@ -73,13 +73,13 @@ export const declare_dp_rev_split_panel = () =>
           },
           net: {
             index: 1,
-            header: text_maker("dp_gross"),
+            header: text_maker("dp_net"),
             is_summable: true,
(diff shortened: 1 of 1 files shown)


Stephen-ONeil:
Hmm, this table needs some polish, going to open an issue

PR #679, opened 2020-07-07
Gross net spending polish

### LINKED ISSUE #678: Gross and Net Planned Spending table needs polish
image

 - Inconsistency between "Planned Spending (net)" and "Gross planned spending", both in structure and capitalization
    - this may have been partially on purpose, because net is what we usually mean when we say planned spending and this may have been done to try and communicate it that way (but it just looks sloppy)
  - Ordering feels off. With this ordering, column 1 = columns 2 + 3 - 4, but I think you'd expect the final column to be the product, not the first. Maybe order it either `gross | special | revenue | net` or `revenue | gross | special | net`?
    - could be good to some how further communicate the formula happening in this table visually (make revenue numbers red, net or gross some third colour?)
  - "Planned Specified Purpose Account Spending" should be hidden if it's zero in all years, it's a weird special case that very few departments have if I remember right (someone look at the data and check me on that). Annoying to have one super wide column in the middle of the table if it's going to be a bunch of zeros more often than not

GarvinH:
closes #678 

CBSA
image

JUS
image

Fixed capitalization
Rearranged columns to be cleaner
Hide SPA if it is all 0

### DIFF
--- a/client/src/components/DisplayTable/DisplayTable.js
+++ b/client/src/components/DisplayTable/DisplayTable.js
@@ -117,7 +117,7 @@ export class DisplayTable extends React.Component {
       const current_col_formatter = col_configs_with_defaults[col].formatter;
       const current_col_raw_formatter =
         col_configs_with_defaults[col].raw_formatter;
-      if (current_col_formatter) {
+      if (current_col_formatter && !_.isFunction(current_col_formatter)) {
         if (_.isString(current_col_formatter) && _.isNumber(row[col])) {
           return "right";
         }
--- a/client/src/components/DisplayTable/DisplayTable.js
+++ b/client/src/components/DisplayTable/DisplayTable.js
@@ -92,7 +92,6 @@ export class DisplayTable extends React.Component {
           sum_func: (sum, value) => ... <- (function) Custom sum func. Default to sum + value
           sort_func: (a, b) => ... <- (function) Custom sort func. Default to _.sortBy
           sum_initial_value: 0 <- (number) Default to 0
-          color: "class name"
         },
       }
       */,
--- a/client/src/components/DisplayTable/DisplayTable.js
+++ b/client/src/components/DisplayTable/DisplayTable.js
@@ -400,7 +400,7 @@ export class DisplayTable extends React.Component {
                               } else {
                                 return col_configs_with_defaults[
                                   col_key
-                                ].formatter(total_row[col_key], "white");
+                                ].formatter(total_row[col_key]);
                               }
                             } else {
                               return total_row[col_key];
@@ -410,7 +410,8 @@ export class DisplayTable extends React.Component {
                           }
                         })()}
                       </td>
-                    )).value()}
+                    ))
+                    .value()}
                 </tr>
               )}
             </tbody>
--- a/client/src/components/DisplayTable/DisplayTable.js
+++ b/client/src/components/DisplayTable/DisplayTable.js
@@ -31,6 +31,7 @@ const column_config_defaults = {
 /* Assumption: DisplayTable assumes 1st column to be string that describes its row
   - If total row exists, 1st column will have the word "Total"
   - 1st column cannot be toggled off by user
+  - Total row color is set to $textLightColor, see total_color
 */
 export class DisplayTable extends React.Component {
   constructor(props) {
@@ -368,11 +369,12 @@ export class DisplayTable extends React.Component {
               ))}
               {is_total_exist && (
                 <tr key="total_row">
-                  <td>{text_maker("total")}</td>
+                  <td className="total_color">{text_maker("total")}</td>
                   {_.chain(visible_ordered_col_keys)
                     .tail()
                     .map((col_key) => (
                       <td
+                        className="total_color"
                         style={{
                           textAlign: determine_text_align(total_row, col_key),
                         }}
--- a/client/src/components/DisplayTable/DisplayTable.js
+++ b/client/src/components/DisplayTable/DisplayTable.js
@@ -92,6 +92,7 @@ export class DisplayTable extends React.Component {
           sum_func: (sum, value) => ... <- (function) Custom sum func. Default to sum + value
           sort_func: (a, b) => ... <- (function) Custom sort func. Default to _.sortBy
           sum_initial_value: 0 <- (number) Default to 0
+          color: "class name"
         },
       }
       */,
@@ -343,6 +344,7 @@ export class DisplayTable extends React.Component {
                 <tr key={i}>
                   {_.map(visible_ordered_col_keys, (col_key) => (
                     <td
+                      className={col_configs_with_defaults[col_key].color}
                       style={{
                         fontSize: "14px",
                         textAlign: determine_text_align(row, col_key),
--- a/client/src/components/DisplayTable/DisplayTable.js
+++ b/client/src/components/DisplayTable/DisplayTable.js
@@ -116,13 +116,10 @@ export class DisplayTable extends React.Component {
       const current_col_formatter = col_configs_with_defaults[col].formatter;
       const current_col_raw_formatter =
         col_configs_with_defaults[col].raw_formatter;
-      if (current_col_formatter && !_.isFunction(current_col_formatter)) {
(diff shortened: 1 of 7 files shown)


Stephen-ONeil:
Just mentioning this in passing, haven't checked my self, but I'm not sure that shade of green is colour-blind accessible. Check the style applying that colour in the chrome dev tools, it will tell you the contrast ratio (and point out if it isn't a high enough contrast).

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -92,6 +92,7 @@ export class DisplayTable extends React.Component {
           sum_func: (sum, value) => ... <- (function) Custom sum func. Default to sum + value
           sort_func: (a, b) => ... <- (function) Custom sort func. Default to _.sortBy
           sum_initial_value: 0 <- (number) Default to 0
+          color: "class name"
```
ktw1016, line 95: This config can be a lot more flexible. With how it's written it is:
1. only limited to style color of td
2. only styles specific color
3. What if in the future, I want to use textGreen but it has nothing to do with `spending`?

You probably get my point. so this can be written with a lot more flexibility.

If you want to have additional config dedicated to `td` style, we can have, `td_class_name` (maybe think of a better name if that sounds bad). It would be a string, which is a css classname defined in where DisplayTable is called. That way, whoever is calling it will need to specify the color. Furthermore, it can be any css rule, not limited to only the color.

ktw1016: BUT, I'd personally prefer to actually even not do that haha. For a reason where it's kind of a duplicate of `formatter` config, and I want to avoid cluttering up the configs and keep it cleaner as much as possible.

so with formatter, you can do something like:
`formatter: (value) => <Format style={{color: 'red'}} type={"dollar"} content={value} />`
okay but then `Format` probably won't take style attribute so maybe you can add that, or do some workaround!

client/src/panels/panel_declarations/finances/drr_dp_resources/dp_rev_split.js
```diff
@@ -71,27 +72,39 @@ export const declare_dp_rev_split_panel = () =>
...
+            color: "spending",
+            initial_visible: (() => {
+              for (const data of table_data) {
+                if (data.spa != 0) {
+                  return true;
+                }
+              }
+
+              return false;
+            })(),
```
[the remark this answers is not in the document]
Stephen-ONeil: Or `_.every` https://lodash.com/docs/4.17.15#every

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -117,7 +117,7 @@ export class DisplayTable extends React.Component {
       const current_col_formatter = col_configs_with_defaults[col].formatter;
       const current_col_raw_formatter =
         col_configs_with_defaults[col].raw_formatter;
-      if (current_col_formatter) {
+      if (current_col_formatter && !_.isFunction(current_col_formatter)) {
```
ktw1016, line 120: Just wondering why did you need this one?

client/src/panels/panel_declarations/finances/drr_dp_resources/dp_rev_split.js
```diff
@@ -8,6 +8,8 @@ import {
   get_source_links,
   declare_panel,
 } from "../../shared.js";
+import { Format } from "../../../../components";
+import styles from "../../../../common_css/_common-variables.scss";
```
ktw1016, line 12: Doesn't seem used

client/src/panels/panel_declarations/finances/drr_dp_resources/dp_rev_split.js
```diff
@@ -22,6 +24,14 @@ const special_cols = _.flatMap(planning_years, (year) => [
 ]);
 const dp_cols = [...planning_years, ...special_cols];
 
+const spending = (value) => (
+  <Format style={{ color: "#008000" }} type={"dollar"} content={value} />
```
ktw1016, line 28: you can throw these in `infobase_color_constants`

client/src/components/DisplayTable/DisplayTable.js
ktw1016, line 95: You didn't delete this one

client/src/components/DisplayTable/DisplayTable.scss
```diff
@@ -89,6 +93,14 @@ table.display-table:not(.no-total-row) > tbody > tr:last-child > td:not(:last-ch
...
 }
 
+td.spending {
+  color: $textGreen;
+}
+
+td.revenue {
+  color: $textRed;
+}
+
```
ktw1016, line 103: You didn't delete these

client/src/panels/panel_declarations/finances/drr_dp_resources/dp_rev_split.js
ktw1016: Actually now that I think of it, green and red colors actually are already defined. Check `color_defs.js`, you can get these from `window. infobase_color_constants`

client/src/components/DisplayTable/DisplayTable.js
GarvinH: Since I changed formatter to a function, it would float left since `_.isString(current_col_formatter)` returned false.

client/src/panels/panel_declarations/finances/drr_dp_resources/dp_rev_split.js
GarvinH: The green in color_defs.js were too low contrast, so I took the green and red from the estimates comparison

client/src/components/DisplayTable/DisplayTable.js
ktw1016: This seems poorly written originally. Try this:
```     
return (
        (_.isString(current_col_formatter) && _.isNumber(row[col])) ||
        _.isNumber(current_col_raw_formatter(row[col]))
  ? "right" : "left"
)
```

client/src/core/color_defs.js
```diff
@@ -20,6 +20,8 @@ const buttonSecondaryColor = "#76a4df";
 
 const textColor = "#333";
 const textLightColor = "#fff";
+const textGreen = "#008000";
+const textRed = "#cc0000";
```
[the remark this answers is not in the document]
GarvinH: I used the highlight red and the success green. the red works but success green doesn't. The issue has to do with contrast against the white background. Stephen told me to use these colors. I've made sure to add that to the scss though.

client/src/components/DisplayTable/DisplayTable.scss
```diff
@@ -1,5 +1,9 @@
 @import '../../common_css/common-variables';
 
+td {
```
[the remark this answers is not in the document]
GarvinH: That was there because chrome was detecting the the blue background instead of the white background causing incorrect contrast ratio being displayed.

client/src/panels/panel_declarations/finances/drr_dp_resources/dp_rev_split.js
```diff
@@ -23,17 +23,17 @@ const special_cols = _.flatMap(planning_years, (year) => [
 ]);
 const dp_cols = [...planning_years, ...special_cols];
 
-const spending = (value) => (
+const spending = (value, custom_color = null) => (
```
ktw1016, line 26: This can actually be better named like `spending_formatter`

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -400,7 +400,7 @@ export class DisplayTable extends React.Component {
                               } else {
                                 return col_configs_with_defaults[
                                   col_key
-                                ].formatter(total_row[col_key]);
+                                ].formatter(total_row[col_key], "white");
```
ktw1016, line 403: I think we can safely make the text white across all display tables

so try putting `!important` on line 89 in Displaytable.scss. You'll probably need to work around with inline css on your formatters

PR #680, opened 2020-07-08
Service standards legend

GarvinH:
Filter service standards by status.

image

image

### DIFF
--- a/client/src/panels/panel_declarations/services/single_service_panels/service_standards.js
+++ b/client/src/panels/panel_declarations/services/single_service_panels/service_standards.js
@@ -145,7 +145,7 @@ export class ServiceStandards extends React.Component {
               : { is_target_met: false }
           )
         : data;
-    console.log(data);
+
     return (
       <Panel title={text_maker("service_standards_title")}>
         <TM className="medium_panel_text" k="service_standards_text" />
--- a/client/src/panels/panel_declarations/services/single_service_panels/service_standards.js
+++ b/client/src/panels/panel_declarations/services/single_service_panels/service_standards.js
@@ -15,11 +15,13 @@ const color_scale = d3
   .domain(["standard_met", "standard_not_met"])
   .range(_.take(newIBCategoryColors, 2));
 
+const standard_statuses = ["met", "not_met"];
+
 export class ServiceStandards extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      active_list: [],
+      active_list: standard_statuses,
     };
   }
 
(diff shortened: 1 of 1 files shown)


client/src/panels/panel_declarations/services/single_service_panels/service_standards.js
```diff
@@ -85,10 +95,81 @@ export class ServiceStandards extends React.Component {
...
+              <span className="link-unstyled" tabIndex={0} aria-hidden="true">
+                {result_simple_statuses[status_key].text}
+              </span>
+            ) : (
+              result_simple_statuses[status_key].text
+            ),
+            icon: large_icons[status_key],
+          }))}
+          item_component_order={["count", "icon", "text"]}
+          click_callback={(status_key) => toggle_active_status_key(status_key)}
```
ktw1016, line 169: you should be able to directly call `setState` in here

client/src/panels/panel_declarations/services/single_service_panels/service_standards.js
```diff
@@ -85,10 +95,81 @@ export class ServiceStandards extends React.Component {
...
+          width={41}
+          vertical_align={"0em"}
+          alternate_color={false}
+          inline={false}
+        />
+      ),
+    };
+
+    const filtered_data =
+      active_list.length == 1
```
ktw1016, line 140: Just heads up, we want to use `===`

GarvinH:
image

The changes seem to make the eyes appear by default now, when on the results page it doesn't. Not that important but I thought I'd bring it up anyways.

client/src/panels/panel_declarations/services/single_service_panels/service_standards.js
```diff
@@ -85,10 +97,76 @@ export class ServiceStandards extends React.Component {
...
+    );
+
     return (
       <Panel title={text_maker("service_standards_title")}>
         <TM className="medium_panel_text" k="service_standards_text" />
-        <DisplayTable data={data} column_configs={column_configs} />
+        <FilterTable
+          items={_.map(standard_statuses, (status_key) => ({
+            key: status_key,
+            count: status_key === "met" ? num_met : data.length - num_met,
```
ktw1016, line 148: Just a super minor thing because I'm not a big fan of this ternary

How about making `num_met` to a function like `get_status_sum` where it takes an argument `is_status_met`, pass in `status_key==="met"`, and I think you should be able to compare `is_status_met` with `is_target_met` to calculate separate sum for "met" and "not_met"

it's really not much different but I think that's slightly more accurate way to do it without this ternary where you do `data.length - num_met`

PR #684, opened 2020-07-13
Show Chart axis when selected on none

### LINKED ISSUE #607: Try and show chart axis even if all items are disabled
We previously stopped users from de-selecting all items in a graph, but it's now possible with the new select all/none legend controls on certain graphs. Can lead to broken looking panels.

Big white space left where the graph would be,  example from a mobile layout (better than the space collapsing in, worse than there still being something there):
image

Probably a fair bit of work, potentially hacking around how nivo works. Might be able to make it more straightforward to achieve this in the process of doing #599, if I plan for it while standardizing the wrapped nivo component APIs.

GarvinH:
closes #607 

image

image

### DIFF
--- a/client/src/panels/panel_declarations/finances/sobj/spend_by_so_hist.js
+++ b/client/src/panels/panel_declarations/finances/sobj/spend_by_so_hist.js
@@ -114,6 +114,7 @@ class SobjLine extends React.Component {
             raw_data={[max_y]}
             enableDots={false}
             lineWidth={0}
+            isInteractive={false}
             margin={{
               top: 10,
               right: 30,
--- a/client/src/panels/panel_declarations/finances/sobj/spend_by_so_hist.js
+++ b/client/src/panels/panel_declarations/finances/sobj/spend_by_so_hist.js
@@ -135,23 +135,8 @@ class SobjLine extends React.Component {
       } else {
         return (
           <NivoResponsiveLine
-            data={
-              !_.isEmpty(spending_data.reverse())
-                ? spending_data.reverse()
-                : [
-                    {
-                      id: "none",
-                      data: [
-                        { x: "2014-2015", y: 7000000000 },
-                        { x: "2015-2016", y: 6000000000 },
-                        { x: "2016-2017", y: 7000000000 },
-                        { x: "2017-2018", y: 7000000000 },
-                        { x: "2018-2019", y: 0 },
-                      ],
-                    },
-                  ]
-            }
-            raw_data={!_.isEmpty(raw_data) ? raw_data : [7000000000]}
+            data={spending_data.reverse()}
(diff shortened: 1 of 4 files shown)


Stephen-ONeil:
Did you look in to disabling tooltips while in none mode? Right now, there's a tooltip for the hidden line, which will be pretty confusing for users (they'll think it is a bug, which makes it a bug).

GarvinH:
Whoops I forgot to remove that earlier. Thanks!

client/src/panels/panel_declarations/finances/sobj/spend_by_so_hist.js
```diff
@@ -87,6 +95,66 @@ class SobjLine extends React.Component {
         .value(),
     };
 
+    const get_line_graph = (() => {
+      if (_.isEmpty(spending_data) && _.isEmpty(raw_data)) {
+        const max_y = _.max(_.map(data, (obj) => _.max(obj.data)));
```
ktw1016, line 100: I think this would be cleaner in a chain like this:
```       
 _.chain(data)
  .map((row) => _.max(row.data))
  .max()
  .value()
```

client/src/panels/panel_declarations/finances/sobj/spend_by_so_hist.js
```diff
@@ -87,6 +95,66 @@ class SobjLine extends React.Component {
         .value(),
     };
 
+    const get_line_graph = (() => {
+      if (_.isEmpty(spending_data) && _.isEmpty(raw_data)) {
```
ktw1016, line 99: Hmm, how about having one variable as Nivo props so that you can get rid of redundant props passing and if statements? This would go for same on other files

```
const is_data_empty = _.isEmpty(spending_data) && _.isEmpty(raw_data);
const empty_data_nivo_props = is_data_empty && { // props that are only for empty data
data: ...,
raw_data: ....
enableDots: ...,
},
const nivo_props = { // props for both regular data and empty data
data: ...,
raw_data:...,
margin:...,
graph_height:...,
...empty_data_nivo_props,
}
```
and then straight in render function,
`<NivoResponsiveLine {...nivo_props} />`

client/src/panels/panel_declarations/finances/transfer_payments/historical_g_and_c.js
```diff
@@ -192,25 +192,39 @@ class DetailedHistTPItems extends React.Component {
       .fromPairs()
       .value();
 
+    const max_value = _.max(
+      _.flatten(_.map(prepped_rows, (row) => _.map(exp_years, (yr) => row[yr])))
```
ktw1016, line 196: This can probably be written in `chain` like above

PR #689, opened 2020-07-16
fix planning_years to include pa_last_year_planned if DPs are tabled …

ktw1016:
…but DRRs not tabled

Check out if these look okay: `crso_by_prog_exp` and `dp_rev_split`

### DIFF
--- a/client/src/panels/panel_declarations/finances/welcome_mat/welcome_mat_exp_program_spending.js
+++ b/client/src/panels/panel_declarations/finances/welcome_mat/welcome_mat_exp_program_spending.js
@@ -12,7 +12,7 @@ import {
   Table,
 } from "../../shared.js";
 
-const { std_years, planning_years } = year_templates;
+const { std_years, correct_planning_years } = year_templates;
 const exp_cols = _.map(std_years, (yr) => `${yr}exp`);
 
 const calculate = (type, subject) => {
@@ -30,7 +30,7 @@ const calculate = (type, subject) => {
     type != "hist" ? programSpending.q(query_subject) : null;
   const progSpending =
     qProgSpending && subject.has_planned_spending
-      ? qProgSpending.sum(planning_years, { as_object: false })
+      ? qProgSpending.sum(correct_planning_years, { as_object: false })
       : null;
 
   return { exp, progSpending };
@@ -59,7 +59,7 @@ export const format_and_get_exp_program_spending = (type, subject) => {
   ];
 
   const history_ticks = _.map(std_years, run_template);
-  const plan_ticks = _.map(planning_years, run_template);
+  const plan_ticks = _.map(correct_planning_years, run_template);
 
   const gap_year =
     (subject.has_planned_spending && actual_to_planned_gap_year) || null;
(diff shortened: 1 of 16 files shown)


server/src/models/finances/constants.js
```diff
@@ -1,18 +1,18 @@
 import _ from "lodash";
 import {
   public_account_years,
-  planning_years,
+  correct_planning_years,
```
ktw1016, line 4: Nothing in this file is used anyways. Is this safe to delete?

server/src/models/constants.js
```diff
@@ -8,7 +8,7 @@ export const public_account_years = [
   "pa_last_year",
 ];
 
-export const planning_years = [
+export const correct_planning_years = [
```
ktw1016, line 11: Same thing for this. It seems like this is not being used anywhere

server/src/models/constants.js
```diff
@@ -8,12 +8,6 @@ export const public_account_years = [
   "pa_last_year",
 ];
 
-export const planning_years = [
-  "planning_year_1",
-  "planning_year_2",
-  "planning_year_3",
-];
-
```
ktw1016, line 16: Not in use anywheres either

server/src/models/finances/constants.js
Stephen-ONeil: I think that's more because the finance section of the API is old/unfinished, since we haven't decided to move to it yet. In theory this stuff would be used eventually, better to keep it around/up to date maybe.

client/src/panels/panel_declarations/finances/welcome_mat/welcome_mat_exp_program_spending.js
```diff
@@ -134,7 +135,7 @@ export const format_and_get_exp_program_spending = (type, subject) => {
         return tick === _.first(plan_ticks) || tick === _.last(plan_ticks);
       } else {
         return (
-          tick === gap_year ||
+          tick === _.nth(all_ticks, _.round(all_ticks.length / 2)) ||
```
Stephen-ONeil, line 138: This'll break if the number of history/planned years changes. Why the change?

ktw1016: Because ticks look a bit bare empty with too few tick labels rendered. I'm simply rendering an additional tick label in the middle. How will this break?

ktw1016:
Not all `planning_years` were replaced because `pa_last_year_planned` doesn't have _rev, _spa, _gross in the data (maybe we should pull that data too?)

so cols need to be defined separately from just `planning_years` and also for that reason, `dp_rev_split` panel should use `planning_years`

Stephen-ONeil:
Hmm, that's annoying. Maybe `extended_planning_years` would be a good name then. `correct_planning_years` is off, because their correctness depends on context (correct for an FTE/spending panel, not correct for a planned revenue panel).

client/src/panels/panel_declarations/finances/welcome_mat/welcome_mat_exp_program_spending.js
Stephen-ONeil: Ah, I thought this was specifically trying to dynamically locate the gap year for some reason. So it's just making sure that the label for the middle tick always renders? Ok.

PR #692, opened 2020-07-17
Hotfix extended_planning_years calculations to be done correctly for …

ktw1016:
…Statistics and missed brackets for one of them

### DIFF
--- a/client/src/tables/programFtes.js
+++ b/client/src/tables/programFtes.js
@@ -277,11 +277,7 @@ Statistics.create_and_register({
 
     add("fte_prg_num", first_year_prg_num);
 
-    const min_planning_yr =
-      "{{planning_year_" +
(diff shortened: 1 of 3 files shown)


client/src/tables/programFtes.js
```diff
@@ -277,11 +277,7 @@ Statistics.create_and_register({
 
     add("fte_prg_num", first_year_prg_num);
 
-    const min_planning_yr =
-      "{{planning_year_" +
-      _.min(_.map(extended_planning_years, (XX) => Number(XX.match(/\d+/)))) +
-      "}}";
-
+    const min_planning_yr = extended_planning_years[0];
```
ktw1016, line 280: extended_planning_years is an array where order is guaranteed, so this should be safe

PR #693, opened 2020-07-17
Standardize one line description

### LINKED ISSUE #648: Standardize one-line GC InfoBase description
There are several different descriptions that are used to explain GC InfoBase to users. The suggestion is to decide on a single definition to be used, and to keep it short and clear, 

Additionally, it was thought that the decision to place the terms “Interactive infographics" and "data” at the start of the description served undermine clarity. We need a simple, plain language, description of GC InfoBase that prioritizes users' needs, over the technical aspects of the site.

AmndaPlant:
Closes #648

### DIFF
--- a/client/src/home/home.yaml
+++ b/client/src/home/home.yaml
@@ -15,8 +15,8 @@ home_results_title:
   fr: Résultats
 
 home_sub_title:
-  en: An interactive data-visualization tool, transforming complex federal data into simple visual stories for Canadians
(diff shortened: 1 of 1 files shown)


Stephen-ONeil:
I think this closes it as far as the InfoBase is concerned. Someone (Darren, likely) will have to follow up with Comms after to make sure all the stuff on their end is standardized to this description.

Is this ready for review? Just a reminder to add that label when it is.

PR #694, opened 2020-07-17
Email backend extraction

### LINKED ISSUE #661: Data management scripts for email_backend
Sent emails are now also logged to a mongoDB. Will eventually need some scripts to make it easy to get that data back out. Haven't written them yet because it's not important data getting stored at the moment, but this should be done before #488.

GarvinH:
closes #661 

Takes the report a problem info and converts into a CSV


GarvinH:
I'm not sure I can access the JSON from memory.
The issue is that the script being run is a mongo script which means that I'm only limited to methods defined in the mongo docs.
I've tried using xargs to collect the output and reuse it as arguments but it makes the situation a little more complicated than necessary. (string values lose their quotation marks for example)

GarvinH:
I'm looking into node child process with exec. I'll try that out and see if that works first.

Stephen-ONeil:
Easiest would be directly using the Node mongo driver to connect from within the node portion, but better would be using mongoose. 

Pros: consistent with the app itself, can leverage the util functions in the app (and thus stay in sync if the implementation details, such as the DB connection string, are changed internally).

Cons: you need the mongoose models defined before you can query through mongoose

... solution: leveraging the util functions in the app 👍 

Connect with `connect_db()`, get the JSON templates and their names from `get_templates()`, pass those in to `make_mongoose_model_from_original_template({template_name, original_template})`, and then pull data from the DB through the models. As long as you can write a generalized function for converting the JSON fetched from the DB to a CSV (which you may already have), then everything should be good to go, with nothing hard coded to any specific template.

email_backend/data_management/generate_csv.js
```diff
@@ -0,0 +1,80 @@
...
+      .filter((key) => {
+        return key !== "__v";
+      })
+      .map((key) => {
+        const key_split = key.split(".");
+        return key_split.length > 1 ? key_split[1] : key;
+      })
+      .value();
+
+    const reduced_json = _.map(submissions, function (sub) {
```
Stephen-ONeil, line 30: Variable names are good for the most part, but lose cohesion as they're re-mapped through the file.

I.e.
  - Get documents from DB, name them `submissions` (presumably an array of the logs for submitted emails, the name `submissions` could be clearer about that, e.g. `email_logs`).
  - map `submissions` with elements as `sub` here (can assume short for a singular "submission", but what a "submission" is still requires knowledge from outside this file)
  - reduce the `sub` with elements as `value` (not a useful var name, who knows what a `value` of a `submission` is? Well, we do, it's a generic field for a logged email, could capture that in a var name, e.g. `email_log_field`). Accumulate the reduced `sub` `value`s as `result` (I'll give a pass here, generally tricky to give the accumulator in a reduce a better name than that)
  - store the reduced `submissions` as `reduced_json`. This is extra misleading, because it's A) an array of "reduced json"s, not a single "reduced jsons", b) has no JSON, just JS objects. More like `flattened_email_logs`

Don't mean to be overly critical. Like I said, taken on their own none of them are necessarily bad names, I just wanted to illustrate how they could be chosen to provide more information to future devs. It's extra important to clearly name the inputs, outputs, and internal arguments of any functional pipeline-style code (like lodash), because the names of variables at each step are the best summary of what each step is intended to do.

With all the lodash we write, I think good naming is particularly valuable. If you know what a step takes and what it outputs (and have confidence that there are no side effects), then you can generally skip reading the implementation. Real time saver, and makes debugging other dev's code a whole lot easier.

email_backend/data_management/generate_csv.js
```diff
@@ -0,0 +1,92 @@
...
+import { Parser } from "json2csv";
+
+const json2csv_parser = new Parser();
+
+connect_db()
+  .then(() => {
+    const templates = get_templates();
+    let templates_left = _.keys(templates).length;
+    _.forEach(templates, (template_value, template_name) => {
+      const collection = make_mongoose_model_from_original_template({
```
Stephen-ONeil, line 16: Should note, this is a `Model` instance, not a mongoDB collection. The mongoose model is just, among other things, to interact with the associated DB collection.

email_backend/data_management/generate_csv.js
```diff
@@ -0,0 +1,92 @@
...
+        original_template: template_value,
+        template_name,
+      });
+
+      //if it happens that later on we add another key with a json value,
+      //change reduced json to handle "key.value". For the time being, the CSV
+      //will be cleaner if we flatten the "key.value" to "value"
+
+      collection.find({}, function (err, email_logs) {
```
Stephen-ONeil, line 25: Aha, no need for the `templates_left` counter, this is a problem to solve with promises! Cleanup like closing connections should go in a `.finally()` at the end of the promises chain (after the `.catch`, same as putting cleanup code after a try-catch block).

Stephen-ONeil: While I'm giving promise-advice, `Model.find` only takes a callback function for legacy reasons (callback hell, the pre-promises state of all async JS code), but if you call it as `Model.find(...).exec()` then the return value is a promise for the (eventually) resolved query. Much nicer to handle than callbacks!

You can keep a promise chain going if you return another promise from a `.then()`. The next `.then()` will receive the result of the previous promise as an argument, once it resolves. In this case, you'd probably want this first `.then` to return a promise for all the async `.find` calls resolving, and then handle the data processing in an additional `.then` call in the chain.

The `Promise` object itself has a couple methods for handling the resolution of multiple promises:`all`, `allSettled`, `any`, and `race`. The differences between some of these are subtle, read the MDN Promises docs for more details. The point here is that each step in a promise chain can ultimately only return one promise (or a non-promise value in the last step), so you'll need to wrap all of the promises from the data fetching step in a `Promises.all` to collapse them to a single promise.

I.e. (untested code, no guarantee I didn't typo something here or miss something)
```
connect_db()
  .then(() => _.chain(get_templates())
    .map((template_value, template_name) => make_mongoose_model_from_original_template( /* stuff */))
    .map((model) => model.find({}).exec())
    .thru(Promises.all)
    .value()
  )
  .then((all_email_logs) => {
    // ... process data
  })
  .catch( (error) => console.log(error) )
  .finally( () => mongoose.connection.close() )
```

Note that if any of the promises from the `model.find({}).exec()` calls reject/throw errors than the data processing step never happens. This may or may not be the desired behaviour. I think it makes sense though, keeps the whole script all-or-nothing which is generally a better way to be (even if it might be annoying if one broken template stops you from getting the data from the others).

PR #700, opened 2020-07-20
Change progress gauge to have 10 circles and be red under 50%

### LINKED ISSUE #699: Service Overview Gauge changes
Change circles to represent the percentage of standard reports met.
Change color when percentage is below 50%

GarvinH:
closes #699 

image

image


ktw1016:
How about showing the actual number of circles if it's 10 or less. And show maximum 10 circles?

PR #702, opened 2020-07-20
Nivo code cleanup

GarvinH:
another step in closing #599 

Any bits of code that I notice can be refactored will be done here

### DIFF
--- a/client/src/core/format.js
+++ b/client/src/core/format.js
@@ -106,14 +106,19 @@ const compact = (precision, val, lang, options) => {
     if (val === 0) {
       return ["", 0];
     } else {
-      for (let i = 0; i < 3; i++) {
-        //can't break out of a lodash loop
-        const breakpoint = 1000000000 / Math.pow(10, i * 3); //checking 1B, 1M, and 1K
-        if (abs >= breakpoint) {
-          return [abbrev[breakpoint][lang], val / breakpoint];
+      const compactted = _.filter(
+        _.times(3, (i) => {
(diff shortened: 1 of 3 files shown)


client/src/core/format.js
```diff
@@ -154,33 +155,35 @@ const compact_written = (precision, val, lang, options) => {
...
-    if (abs >= 1000000000) {
-      reduced_val = val / 1000000000;
-      abbrev = abbrevs[1000000000][lang];
-    } else if (abs >= 1000000) {
-      reduced_val = val / 1000000;
-      abbrev = abbrevs[1000000][lang];
+  const [rtn, abbrev] = (() => {
+    if (abs >= 50000) {
+      const compacted = _.filter(
+        _.times(3, (i) => {
```
ktw1016, line 162: I think chaining this or writing it separately would make it cleaner

client/src/charts/wrapped_nivo/nivo_bar.js
```diff
@@ -127,7 +126,7 @@ export class NivoResponsiveBar extends React.Component {
             motionStiffness,
           }}
           legends={fix_legend_symbols(legends)}
-          keys={_.union([""], keys)}
+          keys={keys}
```
ktw1016, line 129: Are negative bars covered now?

GarvinH: Seems like it! Locally and on Dev link the negative bars are appearing

PR #718, opened 2020-08-04
venn diagram example

AlexCLeduc:
(Don't bother looking at this hackathon-style code, I'm just using a PR here to document this project)

I've been suggesting venn diagrams for years, but they're not the panacea I was hoping they were. Here are a few examples, and you can play with a live example for tooltips' sake by visiting the `#venn-example` route. 

Some of the problems I've come accross are:

1. The library just bails on representing certain areas and prints a console message. It probably bails for a good reason, and it only happens when there are a lot of intersections
2. The library favors accuracy of circular proportions over accuracy of intersections. This is kind of the opposite of what we want. We're pretty loose with scale when we're doing non-standard visualizations like this, and my goal with the venn diagrams was to showcase tiny intersections. 
    * there's no perfect way around this, you can never completely fill a circle's space with 2 half-sized circles
    * A hacky workaround that sometimes works is artificially inflating some of the larger containing groups so that they can fit more smaller circles inside.
3. This one's on me, but I was hoping to visualize the intersection of many sets. Math gets in the way quickly, as there are `2^(number of sets)` possible subsets
4. When using tooltips, it's difficult, sometimes impossible, to hover over specific intersections.

I can only think of 3 patterns of data to use with this chart:

1. Really simple, 2 or 3 set data-driven diagrams
2. many small disjoint sets laid against a single set that joins to most of them (see example 3)
3. Static, hand-crafted (non-data-driven) diagrams 

### 1. Covering the infobase universe of data
ib_universe

In this case, I wrapped everything in a "universe" circle. I don't do it in the other examples. Not sure if it adds any value, but you can see that it looks like there is a subset of organizations that fall out the universe, this is false and is a defect of the library.  

Along this line of thinking, we can also do:

* Orgs with QFR vs DP vs Public Accounts vs budget-tracked vs others
* Orgs with FTEs vs Orgs with headcounts

<br><br>

### 2. Services: External, internal and internal-entreprise

expected_service_venn

This one isn't perfect, I don't think the chart is properly showing there is a non-zero # of services in (not external and internal and internal-entreprise). Also, I originally tried this with the 6 e-enablement groups but it got crazy because `2^6=64` 😄 

<br><br>

### 3. institutional forms and data availability
has_data_and_inst_form_parent

I also tried to throw in the most granular institutional forms, which was a little messier
has_data_and_inst_form

* All of the circles here have empty fills, but I think it might help to fill the "has data" a tiny bit (like 0.1 transparency) to highlight which organizations have data
* I also played around with including population groups, but that was even messier. 
* the chart actually works pretty well in cases like this, all sets are disjoint from each other except data-availability. It's not coincidence that the library's docs feature [a similar example](http://benfred.github.io/venn.js/examples/intersection_tooltip.html)

## Next steps

If we're happy enough with the visuals and the caveats, there's a bunch of work left to do:

### Adding a legend

I used short variable names to label circles above and the overlap was already pretty bad, I can't even imagine what full french titles would look like.

Using more color and adding legends may help. Maybe hovering on the legend can highlight a circle's border.

### improving tooltips

I believe the library tries to pick the smallest hovered inner-join (never left/anti-joins). To stay visually consistent, we can calculate what subsets are part of this join and include these additional breakdowns in the tooltip. Here's an example

Screen Shot 2020-08-04 at 9 13 55 AM

For instance, in this case, the highlighted area is an intersection of 2 sets (results AND financials), but the highlighted region can be split into PEOPLE and NOT people. So its tooltip can have 3 items: all, w/ people and w/ out people.

I haven't totally thought it through, though. In this case, this algorithm would create a tooltip with a ton of items if you hover over the "universe" circle. 

### text-alternative

We could do something like this

| External | Internal | Internal entreprise | Count |
| -------- | -------- | ------------------- | ----- |
| yes      | no       | no                  | 1300  |
| yes      | yes      | no                  | 20    |
| yes      | no       | yes                 | 3     |
| yes      | yes      | yes                 | 3     |
| no       | yes      | no                  | 38    |
| no       | no       | yes                 | 25    |
| no       | yes      | yes                 | 5     |


Stephen-ONeil:
Archived

Issue #139, opened 2019-08-13
Prefer min-width media queries over max-width


Stephen-ONeil:
After PR #6 gets merged, we'll finally have consistent breakpoints, including all of our grid systems being consistent on using min-width queries. Temporary trade off, anything using a max-width media query will transition styling 1px before the grid system/most other content. This only hits at three pixel-specific screen sizes, but it's there and sometimes fairly noticeable.

Solution is to, as much as possible, stick to min-width queries. This should be added to the project style guidelines, and we should make some effort to standardize the existing queries (there's a lot and it's not always trivial to invert them though).

Stephen-ONeil:
Not dead, but should be documented in wiki then closed.

ktw1016:
Also related to #639

GarvinH:
Some files still have max-width lying around. I'm gonna clean it up.

GarvinH:
There's a couple max-width lying around so I'll try to clean it up

GarvinH:
Now that the .98px max is being used, there isn't any problem with the 1px gap anymore. 

I've also updated the wiki point so that is just says preferably use min-widths instead.

If this is all good, I'll close this issue

PR #719, opened 2020-08-05
Clean up media max-width where possible

GarvinH:
related to #139


GarvinH:
Closing so that we use .98 px solution instead of this

PR #721, opened 2020-08-05
Change max-width breakpoints to .98

GarvinH:
Alternative solution to #719


client/src/common_css/_common-variables.scss
```diff
@@ -34,10 +34,10 @@ $minSmallDevice: 768px;
 $minMediumDevice: 992px;
 $minLargeDevice: 1200px;
 /*media breakpoints -1 for max-width queries*/
-$maxExtraSmallDevice: 575px;
-$maxSmallDevice: 767px;
-$maxMediumDevice: 991px;
-$maxLargeDevice: 1199px;
+$maxExtraSmallDevice: 575.98px;
```
ktw1016, line 37: Can you add quick documentation why we're doing .98 here, so we're reminded in the future?

PR #727, opened 2020-08-06
Outage banner

### LINKED ISSUE #549: Mini-deploy for pushing status banners
Right now the deploy requires all(?) IB services to be up, but as we found out today that's not always the case! It would be nice to be able to push a deploy just of the code in order to e.g. enable a "stuff's broken" banner.

GarvinH:
closes #549 

Adding a banner for when there's an outage.
 
image

Data is obtained from a google bucket where the data is manually changed.


server/package.json
```diff
@@ -65,7 +65,8 @@
     "populate_db:debug": "nodemon --watch ../data -e csv --exec babel-node --inspect src/populate_db.js",
     "prebuild": "rm -rf transpiled_build",
     "build": "babel src/ --out-dir transpiled_build --copy-files --ignore node_modules",
-    "mongod": "sh ../scripts/dev_scripts/local_mongod.sh -p 27017"
+    "mongod": "sh ../scripts/dev_scripts/local_mongod.sh -p 27017",
+    "update_outage_info": "gsutil cp ./outage_msg.json gs://ib-outage-bucket/"
```
ktw1016, line 69: Oh so if we're updating it using this, why did you decide to put `outage_msg.json` in gitignore?

client/src/InfoBase/App.js
```diff
@@ -138,6 +164,12 @@ export class App extends React.Component {
           <DevFip />
           <InsertRuntimeFooterLinks />
           <EasyAccess />
+          {this.state.showNotification ? (
```
ktw1016, line 167: This can be short circuited with &&

package.json
```diff
@@ -19,6 +19,7 @@
   "license": "MIT",
   "dependencies": {
     "@lhci/server": "^0.4.1",
+    "axios": "^0.19.2",
```
ktw1016, line 22: So this is being used in client right, then this should actually go into `client/package.json`. Our repo is divided into 2 sub-repos (client and server) so client directory itself is like a repo

client/src/InfoBase/App.js
```diff
@@ -125,6 +127,30 @@ export class App extends React.Component {
...
+    showNotification: false,
+    message: "",
+  };
+
+  hideNotification = () => {
+    this.setState({ showNotification: false });
+  };
+
+  componentDidMount() {
+    axios
```
ktw1016, line 141: Is axios superior than built in `fetch()`?

server/package.json
GarvinH: I was thinking that only the person making the update to the bucket would need to have the file on their system

package.json
GarvinH: Hmm. I could've sworn i did an npm uninstall there. Oh well, I'll fix it

client/src/InfoBase/App.js
GarvinH: I chose it because: 1. It's primarily what I use haha and 2. I saw it being used in the email_backend by someone else.

A little research says that axios is more compatible with older browsers and auto stringifies data. Should I use fetch instead?

ktw1016: I think it's fine!

client/src/InfoBase/App.js
```diff
@@ -125,6 +127,30 @@ export class App extends React.Component {
...
+      .get("https://storage.googleapis.com/ib-outage-bucket/outage_msg.json")
+      .then((res) => {
+        const data = res.data;
+        if (data.outage) {
+          this.setState({
+            showNotification: true,
+            message: window.lang === "en" ? data.en : data.fr,
```
ktw1016, line 148: This is very small minor nitpick but this can be `data[window.lang]`?

ktw1016:
+ Just a reminder to myself to disable CORS for dev-link and live site later!

GarvinH:
Also reminder to update the json so that outage: false

GarvinH:
self reminder to confirm that dev link has no popup when outage: false

PR #728, opened 2020-08-07
Adding lodash to eslint

### LINKED ISSUE #686: Add new lint rules
Refer to #591 

Inspiration of this issue is that i spent few hours trying to find a bug, where lodash chain was missing `.value()` at the end🙂
so linter that catches that would be great

GarvinH:
closes #686


GarvinH:
Content wise, financial, peoples, and results panels seem to be the same

GarvinH:
Seems like the new package-lock shows that react-bootstrap(-typeahead) need to be upgraded.

image

Planning on putting that into its own issue.

GarvinH:
- pipe: custom implementation of lodash -> replace with thru (manual fix)
- string replace: use lodash instead (manual)
- string split: use lodash instead (manual)
- do not use chain for one method: remove the chain() and value() (manual)
- filter with negative comparator: use reject instead; use lodash instead
- map: use lodash instead
- map then flatten: use flatMap() instead
- object assign: use lodash instead
- pullAll: use difference() instead (pullAll is mutable)
- Object.keys: use lodash instead
- toUpperCase/toLowerCase: use lodash instead
- es chaining: use lodash instead
- indexOf compared to -1: use includes() instead
- mapping to execute a function: use invokeMap() instead
- mapping to obtain object values: use values() instead
- function that always returns same value: use constant() instead
- filter to remove falsey values: use compact() instead
- mapping with identity to drop elements: use drop() only instead
- pull: use without() instead (pull is mutable)
- accessing unfinished chain: always finish chains before accessing again
- nested methods: use chain() instead
- flat map to get values: break into. values().flatten()
- uniq with arguments: use uniqBy() instead
- filter then get last: use findLast() instead
- range then mapping: use times() instead
- filter then get first: use find() instead


completely optional (and fixable)
- each -> forEach
- extend -> assignIn
- first -> head

these changes are to use the main function

PR #734, opened 2020-08-12
Filter out inactive votes

### LINKED ISSUE #537: Inactive votes in Estimates Comparision
There are a lot of inactive votes in the estimates comparison tool right now, mainly because of the dead budget votes from 2020-21. I think that either we should:

1) Filter out budget votes (they have a vs_type of 7)
2) Filter out all inactive votes
3) Add footnotes explaining why these votes are inactive

GarvinH:
closes #537 

Before:
image

After:
image

### DIFF
--- a/client/src/EstimatesComparison/EstimatesComparison.js
+++ b/client/src/EstimatesComparison/EstimatesComparison.js
@@ -248,6 +248,17 @@ class EstimatesExplorer extends React.Component {
       col_click,
     };
 
+    _.forEach(
+      root.children,
+      (org) =>
+        (org.children = _.reject(
+          org.children,
+          (estimate) =>
+            estimate.data.current_value === 0 &&
+            estimate.data.percent_value === -Infinity
+        ))
+    );
+
     return (
       <div>
         <div className="medium_panel_text mrgn-tp-lg">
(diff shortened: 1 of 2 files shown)


Stephen-ONeil:
I think they can be useful to see sometimes. Maybe add a checkbox control to turn them back on, and just filter them by default?

client/src/EstimatesComparison/EstimatesComparison.js
```diff
@@ -248,6 +248,18 @@ class EstimatesExplorer extends React.Component {
       col_click,
     };
 
+    !this.state.show_inactive &&
+      _.forEach(
+        root.children,
+        (org) =>
+          (org.children = _.reject(
```
ktw1016, line 255: This will cause mutations..! I suggest below

```
    const filtered_root = this.state.show_inactive
      ? root
      : {
          ...root,
          children: 
            _.map(root.children, (org) => ({
              ...org,
              children: _.reject(
                org.children,
                (estimate) =>
                  estimate.data.current_value === 0 &&
                  estimate.data.percent_value === -Infinity
              ),
            })),
        };
```

client/src/EstimatesComparison/EstimatesComparison.js
```diff
@@ -237,7 +237,7 @@ class EstimatesExplorer extends React.Component {
 
     const { loading } = this.state;
 
-    const root = get_root(flat_nodes);
+    const root = _.cloneDeep(get_root(flat_nodes));
```
ktw1016, line 240: I guess you won't need this?

PR #738, opened 2020-08-14
Show headers of table on fail search

### LINKED ISSUE #737: Display table problem on fail search
Trying to search for a non-existent row in display table completely disables the table.

image

GarvinH:
closes #737 

Now:
image

Before:
image

### DIFF
--- a/client/src/components/DisplayTable/DisplayTable.js
+++ b/client/src/components/DisplayTable/DisplayTable.js
@@ -110,7 +110,7 @@ export class DisplayTable extends React.Component {
     const NoDataMessage = () => (
       <tbody>
         <tr>
-          <td style={{ padding: 0 }} colSpan="3">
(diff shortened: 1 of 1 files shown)


client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -108,9 +108,18 @@ export class DisplayTable extends React.Component {
       })
     );
     const NoDataMessage = () => (
-      <div className="well large_panel_text no-data-msg">
-        {text_maker("no_data")}
-      </div>
+      <tbody>
+        <tr>
+          <td style={{ padding: 0 }} colSpan={_.size(data[0])}>
```
ktw1016, line 113: This doesn't actually matter but `_.size(data[0])` should be equivalent as `visible_col_keys.length`. Same thing but doesn't need to do data[0]. Upto you if you wanna make the change or not haha

Issue #606, opened 2020-05-06
Results panel/language review


Stephen-ONeil:
Things moved quick with this years cramped DRR/DP schedule, new results panels need some more QA (UI/UX in particular).

Some low-hanging fruit already addressed in #605.

Initial notes:
  - [x] `gov_drr_summary_org_table_text` needs a full rewrite, it's doing the 5th grader level thing all backwards
  - [x] sentence introducing the counts reads a bit awkwardly, is there a better way to word that? Example from the gov level: "In 2018-19, of the 2,756 indicators (out of 2,756 total reported):"
  - [x] in the dept and below "Actual Results by Indicator" panel, there's links to a funky single-panel indicator route. I don't want to maintain that, and it's bad UX anyway. Should just be a modal or something that doesn't take you away from the page
  - [ ] revisit the colours used by the results status icons, they run together a lot, might as well all be grey
  - [ ] ... further review needed still

Stephen-ONeil:
I'd still prefer to go back to actually meaningful colours, but I think I otherwise cleaned up the worst of the hatchet job. We can close this, reviewing and polishing this content in general should just happen in an ongoing manner. Don't need to force it especially if it's been sitting open for a whole year.

PR #742, opened 2020-08-17
Results review fixes

GarvinH:
step towards #606

### DIFF
--- a/client/src/panels/panel_declarations/results/IndicatorModalButton.js
+++ b/client/src/panels/panel_declarations/results/IndicatorModalButton.js
@@ -1,3 +1,4 @@
+import text from "./IndicatorModalButton.yaml";
 import gql from "graphql-tag";
 
 import { get_client } from "../../../graphql_utils/graphql_utils.js";
@@ -8,15 +9,18 @@ import {
   SpinnerWrapper,
   StatelessModal,
   WriteToClipboard,
+  create_text_maker_component,
 } from "../../../components/index.js";
 
 import { Indicator } from "../../../models/results.js";
 
 import { IndicatorDisplay } from "./result_components.js";
-import { text_maker } from "./result_text_provider.js";
+import { text_maker as general_text_maker } from "./result_text_provider.js";
 import { Fragment } from "react";
 import { withRouter } from "react-router";
 import { IconCopyLink } from "../../../icons/icons.js";
+
+const { text_maker } = create_text_maker_component(text);
 
 const indicators_fields_fragment = `  id
   stable_id
@@ -155,13 +159,13 @@ class IndicatorModalButton extends React.Component {
           ref={this.buttonRef}
           className="btn-link"
           onClick={() => this.setState({ show_modal: true })}
-          aria-label={`Discover more about ${indicator.name}`}
+          aria-label={`${text_maker("discover_indicator")} ${indicator.name}`}
         >
           {indicator.name}
         </button>
         <StatelessModal
           show={this.state.show_modal}
-          title={text_maker("indicator_display_title")}
+          title={general_text_maker("indicator_display_title")}
           body={
             loading ? (
               <SpinnerWrapper ref="spinner" config_name={"sub_route"} />
@@ -173,7 +177,7 @@ class IndicatorModalButton extends React.Component {
                     <WriteToClipboard
                       text_to_copy={modal_link}
                       button_class_name={"panel-heading-utils"}
-                      button_description={"test"}
+                      button_description={text_maker("copy_indicator_link")}
                       IconComponent={IconCopyLink}
                     />
                   </div>
--- a/client/src/panels/panel_declarations/results/IndicatorModalButton.js
+++ b/client/src/panels/panel_declarations/results/IndicatorModalButton.js
@@ -122,6 +122,11 @@ export default class IndicatorModalButton extends React.Component {
     }
   }
 
+  toggle_modal = () =>
+    this.setState((prev_state) => ({
+      show_modal: !prev_state.show_modal,
+    }));
+
   render() {
     const { id } = this.props;
 
@@ -133,7 +138,7 @@ export default class IndicatorModalButton extends React.Component {
       <Fragment>
         <button
           className="btn-link"
-          onClick={() => this.setState({ show_modal: true })}
+          onClick={this.toggle_modal}
           aria-label={`${text_maker("discover_indicator")} ${indicator.name}`}
         >
           {indicator.name}
@@ -147,6 +152,7 @@ export default class IndicatorModalButton extends React.Component {
             ) : (
               <Panel
                 title={indicator.name}
+                /* the below content will be used once linking to indicators has been added back*/
                 // otherHeaderContent={
                 //   <div style={{ marginLeft: "auto" }}>
                 //     <WriteToClipboard
@@ -162,7 +168,7 @@ export default class IndicatorModalButton extends React.Component {
               </Panel>
             )
           }
-          on_close_callback={() => this.setState({ show_modal: false })}
+          on_close_callback={this.toggle_modal}
           additional_dialog_class={"modal-responsive"}
         />
       </Fragment>
--- a/client/src/panels/panel_declarations/results/IndicatorModalButton.js
+++ b/client/src/panels/panel_declarations/results/IndicatorModalButton.js
@@ -14,13 +14,11 @@ import {
 import { log_standard_event } from "../../../core/analytics.js";
 import { get_client } from "../../../graphql_utils/graphql_utils.js";
 
-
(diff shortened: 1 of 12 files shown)


client/src/panels/panel_declarations/results/IndicatorPanel.js
```diff
@@ -102,51 +101,42 @@ const query_api = (id) => {
 export default class IndicatorPanel extends React.Component {
   constructor(props) {
     super(props);
-    this.state = { loading: true };
+    this.state = { loading: true, show_table: false };
```
ktw1016, line 104: I was confused what `show_table` was at first, maybe better to name `show_modal`

client/src/panels/panel_declarations/results/IndicatorPanel.js
```diff
@@ -102,51 +101,42 @@ const query_api = (id) => {
...
-        ) : (
-          <Panel title={indicator.name}>
-            <IndicatorDisplay indicator={indicator} show_doc={true} />
-          </Panel>
-        )}
-      </StandardRouteContainer>
+      <Fragment>
+        <a role="button" onClick={() => this.setState({ show_modal: true })}>
+          {indicator.name}
+        </a>
```
Stephen-ONeil, line 123: A11y notes on this:
  - anchors that don't have href's aren't focusable by default. Add `tabindex="0"` to let users focus on this when tab navigating
  -  the indicator name doesn't explain what interacting with the button will do. A screen reader user interacting with the button won't have the visual feedback of the modal opening when they interact with it, needs an `aria-label` attribute to tell them that interacting will move them in to a modal with further details on the indicator
  - ... bit of a stretch goal, but ideally focus would return to the anchor when the modal closes (right now it goes back to the top of the table). Could add id's to the anchors and move focus back in the `on_close_callback`

Stephen-ONeil: Hmm, I guess the behaviour is potentially covered by the panel text, which says clicking on an indicator name will give more information.

ktw1016:
I think service inventory needs it single service routes. It has decent amounts of visualizations that wouldn't be served well if it was in a modal

Stephen-ONeil:
Oh, is it multiple panels, or just one panel with a lot in it?

ktw1016:
It has multiple panels, you can check it out here as an example (https://dev.ea-ad.ca/service_inventory/index-eng.html#dept/1/service-panels/136)

client/src/panels/panel_declarations/results/IndicatorPanel.js
```diff
@@ -102,56 +101,46 @@ const query_api = (id) => {
...
-          <Panel title={indicator.name}>
-            <IndicatorDisplay indicator={indicator} show_doc={true} />
-          </Panel>
-        )}
-      </StandardRouteContainer>
+      <Fragment>
+        <button
+          className="btn-link"
+          onClick={() => this.setState({ show_modal: true })}
+          aria-label={`Discover more about ${indicator.name}`}
```
Stephen-ONeil, line 124: Use `text_maker`, needs French version.

client/src/panels/panel_declarations/results/IndicatorPanel.js
```diff
@@ -102,56 +101,46 @@ const query_api = (id) => {
 export default class IndicatorPanel extends React.Component {
```
Stephen-ONeil, line 101: The name `IndicatorPanel` has become misleading, this is something different now. Maybe `IndicatorModalButton`?

Stephen-ONeil: A panel's a specific thing, I mean.

client/src/panels/panel_declarations/results/IndicatorPanel.js
```diff
@@ -102,56 +101,46 @@ const query_api = (id) => {
...
 
   componentDidMount() {
-    const {
-      match: {
-        params: { id },
-      },
-    } = this.props;
+    const { id } = this.props;
     query_api(id).then(() => this.setState({ loading: false }));
   }
```
Stephen-ONeil, line 110: This is going to fire for every "`IndicatorPanel`" as soon as the button mounts in the table. That intentional?

In the best case, that means the loading happens in the background and the user never sees a spinner... but it might also mean that the one indicator they cared about is stuck in a long loading queue behind a bunch of other API queries that they don't actually care about and they get stuck with a spinner longer than they wanted. In the worst (and probably most likely case), we're burning up their network/device resources on a ton of network traffic and JS processing for resources they'll never see.

I'd say don't kick off the loading till the given modal is opened.

Stephen-ONeil (changes requested):
Feels like `IndicatorPanel` could be two components. A higher order, general purpose modal button component to control the modal state, taking an arbitrary component as it's content, and an `IndicatorDisplayPanel` that handles it's own loading and doesn't care where it renders (modal or otherwise). Take that advice if you want, otherwise I've also got some specific comments below (which may or may not still matter if you split the component up).

client/src/panels/panel_declarations/results/IndicatorModalButton.js
```diff
@@ -0,0 +1,202 @@
...
+  constructor(props) {
+    super(props);
+    this.state = { loading: true, show_modal: false };
+
+    const {
+      location: { search },
+      id,
+    } = this.props;
+
+    const query = new URLSearchParams(search).get("ind");
```
Stephen-ONeil, line 122: `URLSearchParams` doesn't have IE11 support, not sure if we're polyfilling for it so this might be broken in IE.

Stephen-ONeil (changes requested):
Hey, nice initiative on putting linking to open modals in this PR!

But... there was a reason I suggested getting this PR merged without it and following up with an issue/consecutive PR for that feature. Doing it properly is a bit more involved than using query strings. I meant to explain what it'd take at some later point, but I'll get in to it now.

The infographic route itself already has a param for arbitrary key value pairs. It's the param named `options` [here](https://github.com/TBS-EACPD/infobase/blob/edd6145555ce7093e6e1b40f1b03ee949e554970/client/src/infographic/Infographic.js#L359). Currently only used by `Infographic.js` itself [here](https://github.com/TBS-EACPD/infobase/blob/edd6145555ce7093e6e1b40f1b03ee949e554970/client/src/infographic/Infographic.js#L153), for the direct-link-to-panel feature. The `options` param uses a URL safe JSON encoding specifically so it could hold arbitrary\* optional route state, without needing to constantly be changing the set of route params (and possibly breaking old links).

The plan was to give the infographic as a whole a React context that would mirror to the options URL param. As needed, panels that want to store state in the URL could consume and update that context. The "plumbing" I referred to in my earlier comment was creating that context/hooking it up to the options param. It's remained unimplemented because we didn't actually have a solid use case for that (until now).

\* we wouldn't actually put _arbitrary_ state in there though, it's ugly and (rarely) gets miss-interpreted as an invalid URL by older/non-spec compliant programs. Only to be used as necessary

Up to you if you push on with the feature now, getting in to this infographic-wide context stuff, or roll back to before you added that feature. I wouldn't want to go out with the query parameter version just to switch later, always got to be careful about maintaining support for anything in the URL bar, since old links can float around for a long time in emails, etc.

Either way, please read my comments below. I wrote them as if I wasn't going to suggest a different approach in the end, because I figured there were some good learning experiences to be had. Thanks!

client/src/panels/panel_declarations/results/IndicatorModalButton.js
```diff
@@ -0,0 +1,171 @@
...
+        <StatelessModal
+          show={this.state.show_modal}
+          title={general_text_maker("indicator_display_title")}
+          body={
+            loading ? (
+              <SpinnerWrapper ref="spinner" config_name={"sub_route"} />
+            ) : (
+              <Panel
+                title={indicator.name}
+                // otherHeaderContent={
```
Stephen-ONeil, line 150: I get leaving this in for now if we hope to be able to use it sooner rather than later. Just in case, add a comment line over this explaining why these lines were commented out rather than deleted (in case we forget about this and someone is confused down the line about why there's a block of commented out code here since we tend no to do that).

client/src/panels/panel_declarations/results/IndicatorModalButton.js
```diff
@@ -0,0 +1,171 @@
...
+
+    const { loading } = this.state;
+
+    const indicator = Indicator.lookup(id);
+
+    return (
+      <Fragment>
+        <button
+          className="btn-link"
+          onClick={() => this.setState({ show_modal: true })}
```
Stephen-ONeil, line 136: Earlier comment still applies.

Stephen-ONeil (changes requested):
Just two things.

Stephen-ONeil (approved):
The aria label

PR #743, opened 2020-08-17
time slider for canada graph

### LINKED ISSUE #654: Improve Canada Graph
UX participants were able to find the numbers for employees working abroad for the most recent year, but struggled to find historical numbers. 

Consider adding the historical data as an inset line graph, or the addition of a time-scrubber under the graph to allow for the selection of the historical year.

ktw1016:
closes #654 

Pretty hacky, but oh well


client/src/charts/canada/CanadaD3Component.js
```diff
@@ -153,14 +150,11 @@ export class CanadaD3Component {
     };
 
     // Set province colours, attach event dispatchers
-    const province_is_active = (prov_key) =>
-      _.some(data, (year) => year[prov_key]);
+    const province_is_active = (prov_key) => data[prov_key];
```
GarvinH, line 153: Not sure if you want to change it back to how it was before but

image

Previously, the provinces/territories with historical data still had colour showing you could hover over it.

If you choose to keep it this way, you could remove the method altogether like this
`const get_color = (prov_key) =>
      data[prov_key] ? main_color : secondary_color;
    const get_opacity = (prov_key) =>
      data[prov_key] ? color_scale(data[prov_key] || 0) : 0.5;`

GarvinH, line 153: On a separate note

image

Ontario seems to have lost its label while nova scotia's hangs around

image

client/src/charts/canada/CanadaD3Component.js
```diff
@@ -237,9 +229,12 @@ export class CanadaD3Component {
 
         d3.select(this)
           .append("p")
-          .style("margin-bottom", "0px")
-          .html(formatter(last_year_data[prov_key] || 0));
+          .attr("class", "label-value")
+          .style("margin-bottom", "0px");
       });
+    html.selectAll("p.label-value").each(function (prov_key, i) {
```
GarvinH, line 235: Just a minor thing, eslint lodash has a preference to use .forEach since .each is an alias and using .forEach only will be consistent once the merge is completed. (If you don't like this preference, we can disable it. Just let me know!)

client/src/charts/canada/CanadaD3Component.js
```diff
@@ -248,10 +243,7 @@ export class CanadaD3Component {
       });
     const hide_optional_components = (prov_keys, selector_template) =>
       _.each(prov_keys, (prov_key) => {
-        const corresponding_province_has_data = _.some(
-          data,
-          (yearly_data) => yearly_data[prov_key]
-        );
+        const corresponding_province_has_data = data[prov_key];
```
GarvinH, line 246: We could probably remove this line and adjust the line under `if (!data[prov_key])`

client/src/charts/canada/CanadaD3Component.js
ktw1016: Ahh I think this is `each` from d3 so it should be irrelevant!

Stephen-ONeil (changes requested):
Quick research, looks like the "thumb" (the circle on the slider) can be overwritten with CSS. E.g. https://css-tricks.com/styling-cross-browser-compatible-range-inputs-css/#styling-the-thumb Could do a pure CSS triangle or possibly use an svg (via `background: url(...)`, [source](https://codepen.io/esalter-va/pen/prRKvM)). 

That wouldn't solve the label text... going back to hacky, but if the thumb is a solid colour you could pre-place the label text with the standard background colour, so that it's "invisible" unless the thumb is positioned under it? Maybe a bit more thought would present a less hacky solution though, haha.

Otherwise, the arrow gets mis-positioned on screen resizes right now. If you stick with the current approach then you might want to force it to rerender on (debounced) screen resizes.

ktw1016:
Well, I pre-placed the text for dates above slider track and tried using pseudo-element slider thumb:
1. map marker svg css inline. I couldn't get the size of the map marker to be big enough to fit all of the text
2. pure css map marker - this is apparently impossible using pseudo element
3. Simple pure css triangle - again, can't really get triangle to fit just right to show all of the text.

Biggest problem is that it's really hard to center the text to the shape on every tick

Stephen-ONeil:
Looking back at this, while it may have been fancy looking if implemented well, a "time scrubber" was probably a poor design choice. The time range here isn't near-continuous like a video, etc., it's just five discrete values. A scrubber takes up a lot of space, requires more effort to interact with from the user, and possibly some additional directions on how to use it... but the root of the original problem was that users didn't read the direction explaining how they could already access historical data through the chart!

Solved the same problem with a dirt simple, self explanatory, select element.

PR #748, opened 2020-08-19
Change effectif for footnotes and glossary CSV

GarvinH:
In #669, after the undos on all the CSVs, I forgot to re-update the CSVs that infobase has control over for the word "effectif"


ktw1016:
Was that all for the word effectif? BTW this needs to be updated from pipeline

Stephen-ONeil:
This was still a potentially contentious change. On one hand, it's clearly a change for the better in terms of readability, on the other hand there were vague concerns of technical accuracy. Still waiting to hear back on discussion with translation.

Stephen-ONeil:
In some footnotes, effectif is the more technically correct term. Otherwise, I'm not too concerned about making all footnotes plain-language, it's more permissible to use technical language there then in more up front content.

Besides, footnote and glossary changes have to happen up-stream in the pipeline, not here.

PR #749, opened 2020-08-19
Salary and motor car footnote

### LINKED ISSUE #536: Really prominent footnote about Motor Car Allowance
In the estimates comparison, there is a very prominent footnote at the department level for several departments that mentions that the Salary and Motor Car Allowance items have been combined. I don't think it needs to be so prominent. Can it be collapsed or or just moved to the actual stat item?

GarvinH:
closes #536 

I don't really see a way to avoid hardcoding this without changing the data that is coming in.
Also this solution unfortunately cannot address the irrelevant footnotes under other departments.

### DIFF
--- a/client/src/EstimatesComparison/EstimatesComparison.js
+++ b/client/src/EstimatesComparison/EstimatesComparison.js
@@ -14,6 +14,7 @@ import {
   RadioButtons,
   LabeledBox,
   CheckBox,
+  Details,
 } from "../components/index.js";
 import { get_root } from "../explorer_common/hierarchy_tools.js";
 import {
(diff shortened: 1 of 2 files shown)


ktw1016:
I'm actually thinking Bill already fixed this... Any thoughts?

GarvinH:
image

Doesn't seem to be

bill-lliu:
I remember reporting this but I think Ross mentioned he wanted to take a look after I showed him, so I just left it ahha

Stephen-ONeil:
Whoa, wait, why not just remove it from the footnote source? This is a bug-waiting-to-happen, when we all forget about this and a new footnote about the "salary and motor car allowance"gets added but never gets displayed here.

Also, this fix only works in English.

PR #754, opened 2020-08-24
Add option to disable UI elements of display table

### LINKED ISSUE #752: add ability to disable UI interactions in DisplayTable
In certain cases, the `DisplayTable`'s  several UI interactions are sort of useless, like searching or exporting a table that only has 2 rows. Can we allow opting out from showing some of those options?

* allow disabling the searching
* allow disabling the sort buttons (I'm not suggesting we don't sort, just the buttons)
* allow disabling the column selection
* allow disabling the export icons

These can all be simple boolean props passed to `DisplayTable`. To keep things simple for consumers, disabling sorting/searching could impact all columns. 

If these options are `undefined`, we could have sensible defaults:

* if # of rows < X, disable searching
  * I'd say X= 6 or 7 :shrug:
* if # of rows < Y disable sort buttons
  * I'd say Y = 3
* If # of _possible_ columns < 3, disable column selection
* If `num_rows x num_columns` in total < Z, disable export icons
  * This estimates the work involved in manually copying the table, but I'm not sure how people use this field. Pasting a small table in excel is pointless, but might make sense in an email. I'll give it a try in outlook, if it doesn't work, Z can be larger.

GarvinH:
closes #752

image

image

### DIFF
--- a/client/src/components/DisplayTable/DisplayTable.js
+++ b/client/src/components/DisplayTable/DisplayTable.js
@@ -26,6 +26,7 @@ const column_config_defaults = {
   sum_func: (sum, value) => sum + value,
   raw_formatter: _.identity,
   sum_initial_value: 0,
+  show_search: true,
 };
 
 /* Assumption: DisplayTable assumes 1st column to be string that describes its row
@@ -96,14 +97,11 @@ export class DisplayTable extends React.Component {
           sort_func: (a, b) => ... <- (function) Custom sort func. Default to _.sortBy
           sum_initial_value: 0 <- (number) Default to 0
           visibility_toggleable: true <- (boolean) Defaults to false for index 0, true for all other indexes.
+          show_search: (boolean) Default to true
         },
       }
       */,
       util_components,
-      show_search,
-      show_sort,
-      show_column_select,
-      show_export_icons,
     } = this.props;
     const { sort_by, descending, searches, visible_col_keys } = this.state;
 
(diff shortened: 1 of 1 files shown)


client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -214,18 +218,31 @@ export class DisplayTable extends React.Component {
...
-      downloadCsvUtil: (
+      downloadCsvUtil: showing_export_icons && (
         <DisplayTableDownloadCsv
           key="downloadCsvUtil"
           csv_string={csv_string}
           table_name={table_name}
         />
       ),
-      columnToggleUtil: (
+      columnToggleUtil: showing_column_select && (
```
ktw1016, line 245: So `_.size(all_ordered_col_keys) > 2` straight in here. But also, we'd want to check that `initial_visible` is true for all columns for this as well to make sure there are no hidden columns

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -324,7 +341,7 @@ export class DisplayTable extends React.Component {
                         />
                       </div>
                     )}
-                    {searchable && (
+                    {searchable && showing_search && (
```
ktw1016, line 344: I think we can do
```searchable = col_configs_with_defaults[column_key].is_searchable && col_configs_with_defaults[column_key].show_search```

PR #759, opened 2020-08-26
Graph alert when screen too small

### LINKED ISSUE #55: Two ideas for accommodating graphs that just plain don't fit well on phones
Specifically, this is about a case like an isolated graphic within an infographic, where just giving the whole route some large min-width and letting mobile users swipe/pinch-zoom would be the nuclear option.

Idea 1) a component that, under a minimum width, displays a transparent grey overlay with the standard "turn your phone to portrait mode" icon. Portrait mode on any phone should be enough to decently render any of our current problematic graphs, I think. It could probably also do with a dismiss button for users who don't want to switch to portrait mode.

Idea 2) a component that over-flow scrolls its children with some adequate min-width. Maybe give it an animated icon of a pointer (👆) swiping right to left to indicate that it's content can be scrolled, with that icon being removed by the first onScroll event.

GarvinH:
closes #55 

Mobile solution:
image

Desktop solution:
image


client/src/charts/wrapped_nivo/wrapped_nivo_bar.js
```diff
@@ -200,6 +200,7 @@ export class WrappedNivoHBar extends React.Component {
       isInteractive,
       motionDamping,
       motionStiffness,
+      mobile_overlay,
```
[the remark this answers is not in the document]
GarvinH: undo function ran out of memory :(

client/src/components/RotateLandscape/RotateLandscape.js
```diff
@@ -1,43 +1,66 @@
...
-                dangerouslySetInnerHTML={{ __html: svgs.expand_width_icon }}
-              />
-            </div>
-          </MediaQuery>
-          <MediaQuery minWidth={568}>{children}</MediaQuery>
-        </MediaQuery>
-      </Fragment>
+      <div style={{ position: "relative" }}>
+        {children}
+        {rotate_landscape ? (
```
ktw1016, line 22: Minor things but this can be short circuited `rotate_landscape &&`
and also
```
              <div className="overlay">
                <TM k="expand_width_text" el="h2" style={{ color: "white" }} />
                <IconExpandWindowWidth width="50%" />
                <button
                  className="btn btn-ib-primary"
                  onClick={rotate_landscape_off}
                >
                  {text_maker("show_anyway")}
                </button>
              </div>
```
this is redundant so we can declare it in variable and just call it

ktw1016: Also `rotate_landscape_off` name is a bit unintuitive for me. Something like `turn_landscape_overlay_off` would be better or `ignore_landscape_overlay`

ktw1016 (commented):
Minor comments and I love the gradient you put and functionality is great!
Alex or Stephen should take a look and see what they think first but it looks good to go for me

client/src/components/GraphOverlay/GraphOverlay.js
```diff
@@ -2,32 +2,32 @@ import { Fragment } from "react";
...
-import { rotate_landscape_off } from "../../InfoBase/AppState.js";
+import { hide_graph_overlay } from "../../InfoBase/AppState.js";
 import { create_text_maker_component } from "../misc_util_components";
 
-import text from "./RotateLandscape.yaml";
+import text from "./GraphOverlay.yaml";
 
 const { TM, text_maker } = create_text_maker_component(text);
 
 class LandscapeOverlay extends React.Component {
```
ktw1016, line 15: Kinda weird to see that everything changed to "GraphOverlay" and this didn't change. Not that it matters whole lot

client/src/components/GraphOverlay/GraphOverlay.js
```diff
@@ -2,32 +2,32 @@ import { Fragment } from "react";
...
   render() {
-    const { children, rotate_landscape, rotate_landscape_off } = this.props;
+    const { children, show_graph_overlay, hide_graph_overlay } = this.props;
```
ktw1016, line 17: `show_graph_overlay, hide_graph_overlay` is still a bit weird to me lol sorry I think leaving `hide_graph_overlay` as-is and `is_showing_graph_overlay` would be good!

ktw1016 (approved):
Two very minor comments and otherwise really nice work! I think this will solve a lot of problems now and also in the future

PR #761, opened 2020-08-28
Results indicator sort by target achieve date fix

### LINKED ISSUE #487: Results ordering
Sigh, perhaps this isn't solved, even after #346

Issues:
- Departments have the ability to order CRs, programs, results, and indicators however they wish (because some depts demanded the feature in Titan). It's not clear how much of this makes it through the pipeline, and what (if anything) InfoBase does to sort programs etc on its end.
- Right now indicators are sorted by target date, which probably has a lot of ties. People are reporting differing orderings of indicators--this is *probably* due to ties in the target date but it's worth double checking to make sure there aren't bugs happening. I maybe didn't check the hierarchy itself when I was doing #346 and used textdiff to check the sorting?

GarvinH:
closes #487 

image

Note: initial sort order remains as alphabetical on core responsibility/program

### DIFF
--- a/client/src/components/DisplayTable/DisplayTable.js
+++ b/client/src/components/DisplayTable/DisplayTable.js
@@ -172,7 +172,10 @@ export class DisplayTable extends React.Component {
               )
             : _.sortBy(unsorted_array, (row) =>
                 is_number_string_date(row[sort_by])
-                  ? row[sort_by]
+                  ? /*FIXME: sorting data containing BOTH string and date is not consistent 
+                 It's probably trying to cast string into date object and if that fails, it probably stops sorting.
+                 We'll have to find all locations of where date is used and update it.*/
+                    row[sort_by]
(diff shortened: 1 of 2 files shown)


client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -172,7 +172,10 @@ export class DisplayTable extends React.Component {
               )
             : _.sortBy(unsorted_array, (row) =>
                 is_number_string_date(row[sort_by])
-                  ? row[sort_by]
+                  ? /*FIXME: sorting data containing BOTH string and date is not consistent 
+                 It's probably trying to cast string into date object and if that fails, it probably stops sorting.
+                 We'll have to find all locations of where date is used and update it.*/
```
ktw1016, line 177: Could be wrong but I don't think there are anymore

PR #765, opened 2020-09-01
wrapper displaytable component for smart props

### LINKED ISSUE #760: Wrapper component for DisplayTable that picks props
"We could maybe have a wrapper component that picks props (e.g. remap columns’ is_searchable based on data), that way the API can stay low-level/non-redundant and panels don’t have to repeat the same calculations all the time.
It’s clear to me we’re gonna keep using this for a ton of stuff, and it’ll probably get more complicated. Dumping the responsibility of picking defaults onto wrappers can help prevent it from getting too smart"

For e.g. 
```
function SmartDisplayTable({columns, data}){
  cols_prop = _.map(columns, col => {...col, is_searchable: col.is_searchable && has_enough_data(col,data) )
 return <DisplayTable columns={cols_prop} data={data} />
}
```

from Alex

ktw1016:
closes #760


AlexCLeduc (approved):
LGTM, thanks!

PR #766, opened 2020-09-01
hide pagedetails on contact route

### LINKED ISSUE #757: Report a problem form on the contact us page, source of confusion?
A user used the report a problem form to ask a number of questions while on the contact route. It might be that they thought it was a contact form. Being a one-way form, with no response mechanism, it's not to anyone's benefit for this to happen. Perhaps we should hide the report a problem form on the contact page, or add a new contact us form on that page to draw attention from the report a problem form?

ktw1016:
closes #757

### DIFF
--- a/client/src/components/PageDetails.js
+++ b/client/src/components/PageDetails.js
@@ -1,6 +1,8 @@
 import "./PageDetails.scss";
 import text from "./PageDetails.yaml";
 
+import { withRouter } from "react-router";
+
 import { Details } from "./Details.js";
 import { ExternalLink } from "./misc_util_components.js";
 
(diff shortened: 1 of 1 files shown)


client/src/components/PageDetails.js
```diff
@@ -39,31 +41,42 @@ class VersionNumber extends React.Component {
...
-        />
-        <div className="pagedetails__version-number fcol-md-4 fcol-sm-6">
-          <VersionNumber />
-        </div>
-      </div>
-    );
+const PageDetails = withRouter(
+  class PageDetails extends React.Component {
+    render() {
+      const { location, toggleSurvey, showSurvey } = this.props;
```
AlexCLeduc, line 47: Total style nitpick, but I think we should prefer early returns as a convention e.g. putting

```javascript
if(location.pathname !== "/contact"){
  return null;
}
``` 

at the top.


Why? As far as I can tell, the return value is the same when `location.pathname !== "/contact"`, but it's a little tough to see that in the diffs because the indentation changed. 


Just something to keep in mind next time, merging PRs is more important than style nitpicks right now haha.

PR #768, opened 2020-09-02
add import linting and sorting

AlexCLeduc:
Added an eslint-plugin to get notified of undefined imports:

Screen Shot 2020-09-02 at 8 28 18 AM

The rule also warns about duplicate imports (two different lines from the same module), so I fixed those.

I tried to get this working in the server, but didn't succeed. We can revisit, it's not nearly as useful for a much smaller project.

### DIFF
--- a/client/.eslintrc.json
+++ b/client/.eslintrc.json
@@ -50,18 +50,19 @@
           },
           {
             "pattern": "src/components/**/*.js",
-            "group": "external",
-            "position": "after"
+            "group": "internal",
+            "position": "before"
           },
           {
             "pattern": "src/models/**/*.js",
-            "group": "external",
-            "position": "after"
+            "group": "internal",
+            "position": "before"
+
           },
           {
             "pattern": "src/core/**/*.js",
-            "group": "external",
-            "position": "after"
+            "group": "internal",
+            "position": "before"
           }
         ]          
       }
--- a/client/.eslintrc.json
+++ b/client/.eslintrc.json
@@ -49,17 +49,17 @@
             "position": "after"
           },
           {
-            "pattern": "components/**/*.js",
+            "pattern": "src/components/**/*.js",
             "group": "external",
             "position": "after"
           },
           {
-            "pattern": "models/**/*.js",
+            "pattern": "src/models/**/*.js",
             "group": "external",
             "position": "after"
           },
           {
-            "pattern": "core/**/*.js",
+            "pattern": "src/core/**/*.js",
             "group": "external",
             "position": "after"
           }
@@ -73,7 +73,7 @@
     },
     "import/resolver":{
       "node": {
-        "moduleDirectory": ["src/","node_modules/"]
+        "moduleDirectory": ["./","node_modules/"]
       }
     }
   },
--- a/client/.eslintrc.json
+++ b/client/.eslintrc.json
@@ -1,6 +1,7 @@
 {
   "plugins": [
-    "react"
+    "react",
+    "import"
   ],
   "parser": "babel-eslint",
   "parserOptions": {
--- a/client/.eslintrc.json
+++ b/client/.eslintrc.json
@@ -26,11 +26,55 @@
       "react/no-string-refs": [ "off" ],
       "react/display-name": [ "off" ],
       "react/no-find-dom-node": [ "off" ],
-      "react/no-children-prop": [ "off" ]
+      "react/no-children-prop": [ "off" ],
+    "import/order": [
+      "warn", { 
+        "newlines-between": "always-and-inside-groups",
+        "alphabetize": {
+          "order": "asc",
+          "caseInsensitive": true 
+        },
+        "groups": [
+          "builtin",
+          "external",
+          "internal",
+          "parent",
+          ["sibling", "index"]
+        ],
+        // We can group imports further by adding rules here, the order they're defined in breaks ties when group and position are equal
+        "pathGroups":[
+          {
+            "pattern": "./*.yaml",
+            "group": "index",
+            "position": "after"
+          },
+          {
+            "pattern": "components/**/*.js",
+            "group": "external",
+            "position": "after"
+          },
+          {
+            "pattern": "models/**/*.js",
+            "group": "external",
+            "position": "after"
+          },
+          {
+            "pattern": "core/**/*.js",
+            "group": "external",
+            "position": "after"
+          }
+        ]          
+      }
+    ]
   },
   "settings": {
     "react": {
       "version": "16"
+    },
+    "import/resolver":{
+      "node": {
+        "moduleDirectory": ["src/","node_modules/"]
+      }
     }
   },
   "globals": {
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
   "license": "MIT",
   "dependencies": {
     "@lhci/server": "^0.4.4",
+    "eslint-import-resolver-node": "^0.3.4",
     "pg": "^7.12.1",
     "pg-hstore": "^2.3.3"
   },
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -14,7 +14,7 @@
         "commonjs": true,
         "es6": true
     },
-    "extends": [ "eslint:recommended" ],
+    "extends": [ "eslint:recommended" , "plugin:import/warnings", "plugin:import/errors"],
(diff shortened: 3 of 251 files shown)


AlexCLeduc:
won't merge before #763 because I want to make sure the module-level imports still work, but I'm still planning on merging before trying to sort imports, as that's gonna have a much larger, distinct footprint

AlexCLeduc:
Ok, the scope of this PR has changed massively now. I've also changed the base branch to oo-explorer (which adds `src/` imports, apologies for not splitting that out into its own PR) until its get merged to master.



There's a way to autofix it for the project in-bulk, it kind of chokes on css/scss imports, though, because the plugin avoids moving imports it knows has side-effects. So when bulk-fixing, I'll just move all style/side-effect imports above non-side-effect imports.

Once all code complies with the rules, you may want to turn on eslint-autofixing. It can be a little annoying to manually sort imports if you don't know the rules super well. I've pasted this in my user settings.json, but we can consider putting it in the workspace too. 

```json
    "editor.codeActionsOnSave": {
        "source.fixAll.eslint": true
    }
```


Import sorting is pretty sweet, but I'd like input on what the rules should be,  here's what I picked:

```json
"newlines-between": "always-and-inside-groups", // allow optional newlines inside import groups
"alphabetize": { 
  "order": "asc",
  "caseInsensitive": true 
},
"groups": [
  "builtin", // things like fs, i don't think we have any in client/src/
  "external",  // anything from node_modules
  "internal", // imports that start from src/
  "parent", // import x from "../" 
  "sibling", //import x from "./x"
  "index" // import x from "./"
],
// We can group imports further by adding rules here, the order they're defined in breaks ties when group and position are equal
"pathGroups":[
  {
    "pattern": "./*.yaml", //force yaml to be imported last
    "group": "index",
    "position": "after"
  },
  {
    "pattern": "components/**/*.js",
    "group": "external",
    "position": "after"
  },
  {
    "pattern": "models/**/*.js",
    "group": "external",
    "position": "after"
  },
  {
    "pattern": "core/**/*.js",
    "group": "external",
    "position": "after"
  }
]          
}
```

a few notes:

* alphabetize sounds annoying but it's only _within_ groups, and it's the only way to prevent stuff like:
  * ```js
      import aa from a/a;
      import ba from b/a; 
      import ab from a/b;
    ```
  * it doesn't mean that imported members have to be sorted alphabetically
* I made extra groups to enforce spaces between models, components and core. These are very frequently imported, especially in large modules
  * this might feel a little overkill for small modules, but taming large files with tons of imports is a more important goal IMO  
* I felt it useful to group yaml imports separately. If you all want them imported _first_, so they're closer to css imports, I'm cool with that too. I just find it useful not to have them interleaved inside plain imports.

here's an example of sorted imports, which was done automatically

```js
import "./TreeMap.scss";

import _ from "lodash";

import {
  SpinnerWrapper,
  create_text_maker_component,
} from "components/index.js";

import { run_template } from "models/text.js";
import { actual_to_planned_gap_year } from "models/years.js";

import {
  sequentialPurples,
  sequentialReds,
  sequentialBlues,
  sequentialGreens,
} from "core/color_schemes.js";
import { formats } from "core/format.js";
import { StandardRouteContainer } from "core/NavComponents.js";

import a from "general_utils";

import { infograph_href_template } from "../infographic/infographic_link.js";

import { get_data, load_data } from "./data.js";
import { TreeMapInstructions } from "./TreeMapInstructions.js";
import { TreeMapLegend } from "./TreeMapLegend.js";
import { TreeMapSidebar } from "./TreeMapSidebar.js";
import { TreeMapTopbar } from "./TreeMapTopBar.js";
import { TreeMap } from "./TreeMapViz.js";

import treemap_text from "./TreeMap.yaml";


```

client/.eslintrc.json
```diff
@@ -25,11 +26,55 @@
...
+        // We can group imports further by adding rules here, the order they're defined in breaks ties when group and position are equal
+        "pathGroups":[
+          {
+            "pattern": "./*.yaml",
+            "group": "index",
+            "position": "after"
+          },
+          {
+            "pattern": "components/**/*.js",
+            "group": "external",
```
ktw1016, line 53: Not sure if I'm not understanding right, but wouldn't these groups be "internal"? 
https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md#importorder-enforce-a-convention-in-module-import-order

AlexCLeduc: They are, but this rule forces them to be placed in between external and internal. 

I think `group:internal, position: before` would achieve the same thing, if that's clearer we can do that 🤷

package.json
```diff
@@ -19,6 +19,7 @@
   "license": "MIT",
   "dependencies": {
     "@lhci/server": "^0.4.4",
+    "eslint-import-resolver-node": "^0.3.4",
```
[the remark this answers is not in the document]
AlexCLeduc: client/eslint doesn't have it own `extends` ... , and instead uses the top level extend, which now extends this plugin's recommended settings, so I think it's required to have it at top level.

We can force client to have its own `extends:` instead I suppose. 

The way eslint is set up is totally weird. I'm happy to change it here, or in an upstream PR, but I don't wanna mess with the lodash-eslint PR.

ktw1016:
I've actually been trying to guess how this would look and after seeing your example, it cleared up so fast. Thank god you wrote that comment lol

1. I think `group:internal, position: before` is clearer since they're technically internal group
2. Why does scss files go at the top? It's index, so it should be at the bottom?
3. I'd prefer yaml to be with scss at the top
4. Just trying to differentiate parent and internal.. So do you expect using both `../../x.js` and `src/...` (parent and internal) syntax?

ktw1016:
Also panel/panel_declaration/shared.js is massively used a lot as well

AlexCLeduc:
> I think group:internal, position: before is clearer since they're technically internal group

Sure!

> I'd prefer yaml to be with scss at the top

Sounds good! I'll put them together


> Why does scss files go at the top? It's index, so it should be at the bottom?

Good point, having it in the middle would jam up the plugin, but putting it at the bottom should be fine. This would also ensure that dependencies' CSS also get loaded before dependents, which is a little more intuitive. This would have to be manually-enforced, though, because the plugin can't sort imports without members. (This is true regardless of whether it's at the top/bottom)


> Just trying to differentiate parent and internal.. So do you expect using both ../../x.js and src/... (parent and internal) syntax?

This is just a silly example to illustrate them in different groups, I don't see any great reasons to use both kinds in the same module, unless we have crazy nesting.

> Also panel/panel_declaration/shared.js is massively used a lot as well. charts not as used as frequently but could be a candidate to be its own

Good point, but it feels a little strange to throw in a single file in its own group. I also think we should rely less on shared (code-splitting is probably suffering from evert single panel using `shared.js`), which will be easier with the new `src/` level imports. Putting shared in its own group would encourage the opposite. 

Also note that a panel importing stuff from shared would often be in its own group, unless it has other src/ imports that aren't core/components/models, and even if it does, you can add optional newlines within that `src/` group.

I would rather remove groupings than add new ones, and let module authors decide when their `src/` group get large enough to add newlines. The alphabetize rule will prevent interleaving core and components anyway. 


<hr/> 
I also just had another idea. I think I'll force `src/`-level imports to actually start with `src/` 

* it's probably easier to make this work in non-webpack projects like email-backend and server
* No one is left wondering whether something is external or internal. 
* It fits the python import style 
* It's only 4 extra characters
* It's easier to reverse this decision (replace "src/" with "") if we don't like the extra 4 characters

ktw1016 (approved):
This is good with me I guess we'll need to look out to sort scss and yamls manually going forward

AlexCLeduc:
Any open PRs will probably have conflicts now.  These are pretty simple to solve, 

1. enable eslint-autofixing in vscode
   - Add this to your user settings.json. I'm also open to adding this to the workspace settings if no one objects
   - ```json 
      "editor.codeActionsOnSave": {
        "source.fixAll.eslint": true
      }
     ``` 

2. on any conflicting file just remove the conflict markers around imports
3. You'll end up with duplicate imports, just remove these
4. If you have brand new files, just re-save them in vscode and let the autofixer handle the formating

Stephen-ONeil:
+1 to adding it to workspace settings, better to synchronize our dev environments on things like this IMO.

PR #775, opened 2020-09-08
Disable sending blank text field when textarea becomes enabled

### LINKED ISSUE #772: Remove simple survey "feedback" link from footer
image

Beside text-only, in the links along the bottom left of the footer.

No longer necessary now that we have the in-site survey. Double check that there aren't other links to the old external feedback survey kicking around (on the about route maybe?).

GarvinH:
closes #772 

image

VS

image

### DIFF
--- a/client/src/InfoLab/InfoLab.js
+++ b/client/src/InfoLab/InfoLab.js
@@ -20,6 +20,8 @@ export default class InfoLab extends React.Component {
   }
 
   render() {
+    const { toggleSurvey } = this.props;
+
     return (
       <StandardRouteContainer
         title={text_maker("lab_title")}
@@ -31,6 +33,13 @@ export default class InfoLab extends React.Component {
         <TM k="lab_title" el="h1" />
         <div className="medium_panel_text">
           <TM k="lab_intro_text" />
+          <button
+            style={{ marginBottom: "2rem" }}
+            className="btn btn-ib-primary"
+            onClick={() => toggleSurvey()}
+          >
+            {text_maker("lab_intro_survey_button")}
+          </button>
         </div>
         <div className="frow">
           <div className="fcol-md-6 fcol-sm-12">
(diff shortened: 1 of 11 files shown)


GarvinH:
A11y now looks like this if this is okay

image

client/src/InfoLab/InfoLab.js
```diff
@@ -28,6 +30,13 @@ export default class InfoLab extends React.Component {
         <TM k="lab_title" el="h1" />
         <div className="medium_panel_text">
           <TM k="lab_intro_text" />
+          <button
```
ktw1016, line 33: Maybe we should disable the survey button in PageDetails on InfoLab page
(I did the same for about route #766 )

client/src/components/EmailFrontend.js
```diff
@@ -167,8 +167,27 @@ class EmailFrontend extends React.Component {
       )
       .value();
 
+    const all_connected_user_fields_are_filled = _.chain(user_fields)
+      .toPairs()
+      .filter(([field_name, field_val]) =>
+        _.includes(_.keys(field_val), "connection")
```
ktw1016, line 173: I'm guessing here but is it necessary to put two filters here?
```
      .filter(
        ([field_name, field_val]) =>
          _.includes(_.keys(field_val), "connection") &&
          _.includes(_.keys(completed_template), field_val.connection.name) &&
          _.includes(
            completed_template[field_val.connection.name],
            field_val.connection.enable
          )
      )
```
You should check if that works !

ktw1016 (approved):
2 super minor comments, other than good to go

client/src/InfoLab/InfoLab.js
GarvinH: 👍 I changed it back so that it still shows the report a problem and github version banners

PR #780, opened 2020-09-14
a11y table to displaytable with nivo default

### LINKED ISSUE #769: Nivo to automatically display DisplayTable in a11y

ktw1016:
closes #769


client/src/charts/wrapped_nivo/CircleProportionChart.js
```diff
@@ -228,7 +228,7 @@ export class CircleProportionChart extends React.Component {
       { label: parent_name, value: parent_value },
       { label: child_name, value: child_value },
     ];
-    const table = !disable_table_view && (
+    const table = (window.is_a11y_mode || !disable_table_view) && (
```
Stephen-ONeil, line 231: I think `disable_table_view` would be used when the table doesn't make sense, right? Wouldn't there be cases where we want to disable the table even in a11y mode? 

More flexible to leave the option of no table even in a11y mode at least. I think it should be the responsibility of any specific panels that use `disable_table_view` to decide wether to set it to false when in a11y mode.

Stephen-ONeil (changes requested):
Just one thought. Let me know if you think I'm wrong though, convince me and I'll approve haha.

Stephen-ONeil (approved):
CI shows failing, but the dashboard shows it as having been manually canceled on the deploy server step. That accurate? If so 👍  for merging

PR #785, opened 2020-09-16
Minor a11y and displaytable

### LINKED ISSUE #782: panel expand/collapse is not a11y friendly
Screen Shot 2020-09-16 at 8 03 38 AM

Right now it just says button, and isn't obvious that it expands or collapses. 

The nodes in explorer have stateful aria text that announce "expand this node"/"collapse this node". Maybe we can copy that. 

Although most UIs always have a collapse before the collapsible content, it might still not be clear whether it collapses the previous or following panel.  Maybe the text should say "Expand the _following_ panel". (And the explorer should also also say _following_ node)

ktw1016:
closes #782 and comments from Alex on Slack
```
I can’t seem to toggle columns on display-tables, the checkbox widget just doesn’t check stuff. It’s not even a screen-reader thing, you can’t check it with a mouse click
https://dev.ea-ad.ca/a11y_table_to_displaytable/index-basic-eng.html#orgs/gov/gov/infograph/financial/.-.-(panel_key.-.-'auth_exp_planned_spending)

Hmm so the heading of a display-table are spread accross two rows
the titles
the sort buttons
It can be weird to have a blank row is when there are no sort buttons. Can we just exclude the sort row in that case?
```

### DIFF
--- a/client/src/components/CheckBox.js
+++ b/client/src/components/CheckBox.js
@@ -4,11 +4,12 @@ import { IconCheckmark } from "../icons/icons.js";
 import "./CheckBox.scss";
 
 export class CheckBox extends React.Component {
-  handleOnClick = (onClick, disabled, id) => {
+  handleOnClick() {
+    const { disabled, onClick, id } = this.props;
     if (!disabled && onClick) {
       onClick(id);
     }
-  };
+  }
   render() {
     const {
       id,
@@ -31,7 +32,7 @@ export class CheckBox extends React.Component {
             <input
               type="checkbox"
               checked={active}
-              onChange={this.handleOnClick.bind(this, onClick, disabled, id)}
+              onChange={this.handleOnClick}
               style={{ marginRight: 5 }}
             />
             {label}
@@ -56,7 +57,7 @@ export class CheckBox extends React.Component {
               ...checkbox_style,
             }}
             className={classNames("checkbox-span", onClick && "span-hover")}
-            onClick={this.handleOnClick.bind(this, onClick, disabled, id)}
+            onClick={this.handleOnClick}
           >
             {!isSolidBox && (
               <IconCheckmark
@@ -74,7 +75,7 @@ export class CheckBox extends React.Component {
               aria-checked={active}
               tabIndex={0}
               className="link-styled"
-              onClick={this.handleOnClick.bind(this, onClick, disabled, id)}
+              onClick={this.handleOnClick}
               onKeyDown={(e) =>
                 (e.keyCode === 13 || e.keyCode === 32) &&
                 !disabled &&
--- a/client/src/components/CheckBox.js
+++ b/client/src/components/CheckBox.js
@@ -4,12 +4,12 @@ import { IconCheckmark } from "../icons/icons.js";
 import "./CheckBox.scss";
 
 export class CheckBox extends React.Component {
-  handleOnClick() {
+  handleOnClick = () => {
     const { disabled, onClick, id } = this.props;
     if (!disabled && onClick) {
       onClick(id);
     }
-  }
+  };
   render() {
     const {
       id,
--- a/client/src/components/CheckBox.js
+++ b/client/src/components/CheckBox.js
@@ -26,7 +26,7 @@ export class CheckBox extends React.Component {
             <input
               type="checkbox"
               checked={active}
-              onChange={onClick}
+              onChange={() => !disabled && onClick && onClick(id)}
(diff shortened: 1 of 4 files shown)


client/src/components/CheckBox.js
```diff
@@ -26,7 +26,7 @@ export class CheckBox extends React.Component {
             <input
               type="checkbox"
               checked={active}
-              onChange={onClick}
+              onChange={() => !disabled && onClick && onClick(id)}
```
GarvinH, line 29: > Now that we have a nice syntax (using arrow functions) for bound methods, we should start respecting React best practices and use bound methods on the component for event handlers, instead of in-line methods.

> It's a performance thing. Every time render is called a new method instance is created for this event handler. JS-engine wise that's trivial, but the problem is that functions are compared by memory location. That means, even if nothing has effectively changed about the component with the event handler, React only knows that the event handler methods appear to have changed and that it thinks it needs to update the actual DOM (relative to the JS-engine, this can be a non-trivial performance hit). With a bound method, there's only one instance (and it, of course, compares as equal to itself), so React can properly do its virtual DOM magic.

GarvinH: Actually, I think it might be okay to leave it like this since the re-renders will be triggered by props which the function relies on anyways.

ktw1016: you're right, I always forget this. Thanks!

AlexCLeduc: I didn't know you can pass additional arguments to bind, I thought it was just for `this`! 

I still find this confusing, though. The intent isn't clear.

- binding `this` is the primary use of `.bind`. This function doesn't need that, as arrow-functions are automatically bound. 
- each render call will create a brand new function anyway, so there's no performance gain 

Since this onchange is just passing props onto the handler, couldn't we access those props directly in the handler?

```js
export class CheckBox extends React.Component {
  handleOnClick = () => {
    const { onClick, disabled, id} = this.props
    if (!disabled && onClick) {
      onClick(id);
    }
  };

<input 
  onChange={this.handleOnClick}
/>
```

GarvinH: Hmm... I did a little more reading and it apparently `.bind()` actually [creates a new function anyways](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind)

I think Alex is right in that it'd be best to access props inside the handler function

client/src/components/CheckBox.js
```diff
@@ -4,6 +4,11 @@ import { IconCheckmark } from "../icons/icons.js";
 import "./CheckBox.scss";
 
 export class CheckBox extends React.Component {
+  handleOnClick = (onClick, disabled, id) => {
+    if (!disabled && onClick) {
```
GarvinH, line 8: Hmm... I did a little more reading and it apparently `.bind()` actually [creates a new function anyways](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind)

I think Alex is right in that it'd be best to access props inside this `handleOnClick`function

Stephen-ONeil: You usually don't bind inline in the render, you'd do it in the constructor. Arrow function methods shouldn't need a `.bind`  though?

ktw1016: I was doing it inline to pass in parameters. I was too focused on trying to pass the parameter, and should've known that binding in render creates a new function every render (doing it in constructor would not) But yeah Alex is right because without binding and using props straight in the function, it'll be auto-bound function

Stephen-ONeil (approved):

client/src/components/Panel.js
```diff
@@ -91,6 +91,11 @@ export class Panel extends React.Component {
                 className={classNames("panel-heading-utils")}
                 onClick={() => this.setState({ isOpen: !isOpen })}
                 aria-labelledby={label_id}
+                aria-label={
```
Stephen-ONeil, line 94: Oh, wait, shouldn't have both a labelled by and a label. What element is `label_id` supposed to refer to?

Stephen-ONeil (changes requested):

client/src/components/Panel.js
ktw1016: oh yeah this is actually fixed in a11y_review branch #800 ! I have to admit It's getting a bit confusing because I have branches unnecessarily merging into another (unintentionally because I was waiting for PR reviews!)

Stephen-ONeil (approved):
👍

PR #787, opened 2020-09-17
add edc and ppp canada to route tests

### LINKED ISSUE #786: Expand browser test scope
Right now only gov and TBS infographics are tested (TBS as a dept plus some of its CR and program infographic routes). Should also visit special case infographics (a crown corp, dead org, etc.), to expand coverage.

ktw1016:
closes #786 

Not actually sure if Stephen meant this as browser tests


Stephen-ONeil (approved):
Yup, that's what I meant!

PR #788, opened 2020-09-18
React-responsive update for MediaQuery

### LINKED ISSUE #589: MediaQuery components with multiple direct descendants
Seems a MediaQuery component with multiple direct children will render them wrapped in a div. When not desired (why would it be...) you can manually wrap the children in a `Fragment`.

At some point we should double check that this little gotcha isn't messing us up anywhere.

Maybe write a linter rule to warn about it, or at least add it to the project wiki as something to be aware of.

GarvinH:
closes #589 

Updated to 6.1.2 (not the latest version) - probably need to update react to get latest to work

but now multiple descendants of MediaQuery no longer produces a div


ktw1016 (approved):
👍

PR #789, opened 2020-09-21
Add linking to results indicators

### LINKED ISSUE #784: Add links to reach indicators
I'm going to undo what i currently have for linking to indicators in #742.

Once that has been merged, I will begin doing this.

GarvinH:
closes #784

### DIFF
--- a/client/src/components/modals_and_popovers/ModalButton.js
+++ b/client/src/components/modals_and_popovers/ModalButton.js
@@ -21,6 +21,12 @@ class ModalButton_ extends React.Component {
     this.state = {
       show_modal:
         show_condition &&
+        /*
+          show_condition: {
+            name: string
+            value: any
+          }
+         */
         options_object[show_condition.name] &&
         show_condition.value === options_object[show_condition.name],
     };
--- a/client/src/components/modals_and_popovers/ModalButton.js
+++ b/client/src/components/modals_and_popovers/ModalButton.js
@@ -5,7 +5,6 @@ import { StatelessModal } from "./StatelessModal";
 import { SafeJSURL } from "src/general_utils";
 
 class ModalButton_ extends React.Component {
-  state = { show_modal: false };
   constructor(props) {
     super(props);
 
@@ -18,12 +17,11 @@ class ModalButton_ extends React.Component {
 
     const options_object = SafeJSURL.parse(options);
 
-    this.state =
-      show_condition &&
-      options_object[show_condition.name] &&
-      show_condition.value === options_object[show_condition.name]
-        ? { show_modal: true }
-        : { show_modal: false };
+    this.state = (show_condition &&
+      options_object[show_condition.name] && {
+        show_modal:
+          show_condition.value === options_object[show_condition.name],
+      }) || { show_modal: false };
   }
 
   toggle_modal = () =>
--- a/client/src/components/modals_and_popovers/ModalButton.js
+++ b/client/src/components/modals_and_popovers/ModalButton.js
@@ -18,11 +18,12 @@ class ModalButton_ extends React.Component {
 
     const options_object = SafeJSURL.parse(options);
 
-    this.state = (show_condition &&
-      options_object[show_condition.name] && {
-        show_modal:
-          show_condition.value === options_object[show_condition.name],
-      }) || { show_modal: false };
+    this.state = {
+      show_modal:
+        show_condition &&
+        options_object[show_condition.name] &&
+        show_condition.value === options_object[show_condition.name],
+    };
   }
 
   toggle_modal = () =>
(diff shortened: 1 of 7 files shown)


client/src/infographic/Infographic.js
```diff
@@ -218,84 +229,91 @@ class InfoGraph_ extends React.Component {
...
     return (
-      <div>
-        <AnalyticsSynchronizer {...this.props} />
-        {window.is_a11y_mode && (
+      <InfographicContext.Provider
+        value={{
+          get_copy_link: this.get_copy_link,
+          get_link_state: this.get_link_state,
+        }}
+      >
```
ktw1016, line 237: I just wanna double check, none of the code below was actually changed right?

client/src/infographic/infographic_link.js
```diff
@@ -15,4 +15,18 @@ const panel_href_template = (
     data_area ? data_area : ""
   }/${panel_key ? SafeJSURL.stringify({ panel_key: panel_key }) : ""}`;
 
-export { infograph_href_template, panel_href_template };
+const infograph_options_href_template = (
```
ktw1016, line 18: I think you should replace `panel_href_template` with this, they're pretty redundant

ktw1016 (changes requested):
I keep thinking if context should be used, you could pass in subject id from instance (result_flat_table this case), and just call `infograph_options_href_template` on IndicatorDisplayPanel level. Reading [this](https://reactjs.org/docs/context.html#before-you-use-context), do you think there are other benefits to using context in this context? 😏

client/src/infographic/Infographic.js
GarvinH: I don't think i changed anything other than the context

client/src/infographic/Infographic.js
```diff
@@ -166,6 +168,15 @@ class InfoGraph_ extends React.Component {
...
     }
   }
+
+  get_copy_link = (options) =>
+    infograph_options_href_template(
+      this.state.subject,
+      this.state.active_bubble_id,
+      options
+    );
+  get_link_state = () => SafeJSURL.parse(this.props.options);
```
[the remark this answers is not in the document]
GarvinH: Can be used in the future, but I guess I'll remove it for now since it's not being used

client/src/components/modals_and_popovers/ModalButton.js
```diff
@@ -18,12 +18,11 @@ class ModalButton_ extends React.Component {
...
-      show_condition &&
-      options_object[show_condition.name] &&
-      show_condition.value === options_object[show_condition.name]
-        ? { show_modal: true }
-        : { show_modal: false };
+    this.state = (show_condition &&
+      options_object[show_condition.name] && {
+        show_modal:
+          show_condition.value === options_object[show_condition.name],
+      }) || { show_modal: false };
```
ktw1016, line 25: I actually meant this:
```
    this.state = {
      show_modal:
        show_condition &&
        options_object[show_condition.name] &&
        show_condition.value === options_object[show_condition.name],
    };
```
just cleaner imo

ktw1016 (changes requested):
I thought more about using context, so this new context is carrying sort of global variables of subject, and bubble id. Well, that's actually already done in `PanelRenderer` and as I mentioned, in the other comment we actually wouldn't even need to use that context (you could pass in subject id from instance (result_flat_table this case), and just call `infograph_options_href_template` on IndicatorDisplayPanel 
level)

To me this adds unnecessary complexity in the code and doesn't add much, so I'd suggest removing context!

client/src/panels/panel_declarations/results/IndicatorDisplayPanel.js
```diff
@@ -117,28 +119,34 @@ export default class IndicatorDisplayPanel extends React.Component {
...
-    const { id } = this.props;
+    const { id, subject, active_bubble_id } = this.props;
 
     const { loading } = this.state;
 
     const indicator = Indicator.lookup(id);
 
+    const panel_link = window.location.href.replace(
+      window.location.hash,
+      infograph_options_href_template(subject, active_bubble_id, {
```
ktw1016, line 130: I think it's safe to just put "result" straight in, without the need of props. Sorry for not being clear on that and keep asking for changes!

client/src/components/modals_and_popovers/ModalButton.js
```diff
@@ -1,11 +1,36 @@
...
+
+    this.state = {
+      show_modal:
+        show_condition &&
+        /*
+          show_condition: {
+            name: string
+            value: any
+          }
+         */
```
ktw1016, line 29: Is this dead code or documentation?

GarvinH: Yup thats documentation

ktw1016 (approved):

PR #790, opened 2020-09-25
Advanced search design

### LINKED ISSUE #496: Advanced Search design change

GarvinH:
closes #496


GarvinH:
Tried using bounding rect but it's too laggy... Gonna close this

PR #792, opened 2020-09-28
Some things to keep in mind

### LINKED ISSUE #653: Reconsider or improve the  "Some things to keep in mind..." banner
The UX participants seemed to miss or ignore the “Some things to keep in mind as you explore this data...” banner entirely. 

Consider ways to increase its prominence, or relocate this information and find ways to direct users to it. Alternately, reconsider the need for the banner/info in general.

GarvinH:
closes #653


ktw1016:
I've been trying to trigger the modal.. I see you set a timer, is it supposed to just pop up?

GarvinH:
Yup it just pops up at 30 seconds if the "some things to keep in mind" accordion isn't pressed or if you haven't seen the popup/accordion in the past

AlexCLeduc:
Triggering a modal seems excessive, reminds me of medium 🤢  

I like the idea of using localStorage, though. A manually dismissible tooltip next to the accordion might grab attention less annoyingly. Something this subtle could probably be triggered 
- just a couple seconds in instead of 30, or 
- based on scroll position, 
- immediately and potentially off screen (sounds way simpler)


I'm not sure we have any examples of tooltips that are programmatically triggered at the moment, all the existing tooltip boxes are hover-triggered. 

`client/src/glossary/TooltipActivator.js` suggests that some code that might work:

```javascript
import Tooltip from "tooltip.js";

new Tooltip(node,{
  placement,
  container,
  html,
  arrowSelector,
  innerSelector,
  delay,
})
```

I'm guessing we could just use a `ref` and `componentDidMount()` to call `new Tooltip(...` 

I had a quick look around and I can't really find any docs on the [tooltip.js package](https://www.npmjs.com/package/tooltip.js/v/1.3.1), it looks like it's been deprecated and you're supposed to use popper now?

GarvinH:
Ok so I replaced it with a bootstrap tooltip that says "click here!/cliquer ci-dessous" which triggers 1 second after loading.

For now, I've left out using a calculation to detect scroll position (getboundingclientbox is decently resource intensive for scrolling if i remember correctly)

Let me know how it feels now! and let me know if the french is bad 😂

ktw1016:
I'm not seeing any tooltip popping up at all, am I missing something?

AlexCLeduc:
Cool! I didn't realize we had popovers in react-bootstrap

Screen Shot 2020-10-02 at 10 12 12 AM


I kind of pictured it sticking out from the right, breaking-out from the right-alignment of the entire infographic. This might not really work on mobile, but maybe we can just turn it off for mobile 🤷 



Also, instead of "click me", we could try

> Recommended reading 

> lecture recommandée

AlexCLeduc:
> I'm not seeing any tooltip popping up at all, am I missing something?

try `localStorage.clear()` , the last dev link used the same keys

GarvinH:
I'm not sure but I think the right align popup doesn't scale too well since the white space on the left and right are inconsistent until reaching the breakpoints

ktw1016:
yeah I can't say I'm a fan of it sticking out from the right. Well the whitespace isn't consistent and it'll have to be debounced on screen resize. Also the triangle shape seem like it needs a fix.

Although it does grab more attention so to use it, it should only be used if the screen size is wide enough. I'm fine with sticking from the right or the top

Stephen-ONeil:
Coming in pretty late here, sorry about that, but what about using position sticky? If the widget was thinner (at least on mobile, maybe only when "stuck"), I think that would look fine. Feels like keeping it on-screen fits well with the idea of it being "things to keep in mind *as you explore*" better than it being something we show once at the top.

Could keep using local storage and all that, have an "X" that removes the position sticky (sending it back to its original spot at the top of the screen). Maybe a "pin" button to turn position sticky back on, in case the user wants that.

Probably turn all that off in text-only mode.

GarvinH:
Hmm, I can't get position sticky to work...
And it seems like all the available react packages for implementing sticky seem to be either no longer supported or haven't been updated in a while.

I'm gonna do an in-house solution for the sticky.
But the width when in sticky mode doesn't seem to want to obey...

ktw1016 (changes requested):
Hey this is awesome! It'll need some tweaks but super cool ! Just some comments from using:
- On mobile, I can't scroll on the accordion, it just scrolls the page (If it can be fixed, should be fine on mobile. If not, just disable it on mobile!)
- I think unpin should also close (collapse) the accordion
- I think pin and unpin icon should be the other way around. So that it's showing the current state if it's pinned or not
- The accordion width expands on clicking on accordion item when it's pinned. Can you make accordion width constant?

GarvinH:
In order to implement the solution to multiple screen sizes, i used ResizeObservers https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver
which isn't supported on IE (but its at end of life anyways so I don't think we should let this stop us) and Firefox mobile (which is the main issue i see with using this. Not sure if the required browser versions are too recent though...)

client/src/panels/panel_declarations/misc/key_concept_panels/KeyConcepts.js
```diff
@@ -0,0 +1,160 @@
...
+          localStorage.getItem(`should_pin_key_concepts`)
+        );
+      } catch {
+        should_pin = null;
+      }
+    }
+
+    this.keyConceptsContainerRef = React.createRef();
+
+    this.state = {
```
ktw1016, line 55: two initial state declaration, I think better to keep it at one place

ktw1016: Also I'm not sure why there are two different states. What different roles are they serving? What does `sticky` mean, and what does `should_pin` mean? Can you also come up with better names? Something like `is_pinned` that name show the state more intuitively?

client/src/panels/panel_declarations/misc/key_concept_panels/KeyConcepts.js
```diff
@@ -0,0 +1,160 @@
...
+                </Overlay> */}
+              </ButtonToolbar>
+            </div>
+          )}
+        </MediaQuery>
+      </div>
+    );
+  }
+
+  componentDidMount() {
```
ktw1016, line 135: This might be nitpicky but I think it's more logical that class has constructor -> componentDidMount -> Render.

ktw1016 (changes requested):

client/src/panels/panel_declarations/misc/key_concept_panels/KeyConcepts.js
GarvinH: whoops! I forgot to remove the state outside of the constructor after moving it into there. There's only one state in reality even if it's declared twice though! (I'm pretty sure at least). I've renamed the variables/keys to more appropriately match their uses.

client/src/panels/panel_declarations/misc/key_concept_panels/KeyConcepts.js
GarvinH: I was always wondering why it was structured like this when the mounting order is like this
image
found here: https://reactjs.org/docs/react-component.html

I do notice that programmers generally structure componentDidMount above render so I put it back.

Stephen-ONeil: If I had to come up with a reason, it'd be that
  a) it feels reasonable to group hooks together
  b) `componentDidMount` is most likely doing something that will prompt a re-render (loading data while the component is in a spinner state, making some calculations based on the location in the DOM, etc)

client/src/panels/panel_declarations/misc/key_concept_panels/KeyConcepts.js
```diff
@@ -0,0 +1,146 @@
...
+          this.setState({ key_concepts_out_of_view: true });
+      } else {
+        this.state.key_concepts_out_of_view &&
+          this.setState({ key_concepts_out_of_view: false });
+      }
+    });
+
+    this.setState({
+      key_concepts_width: this.keyConceptsContainerRef.current.offsetWidth,
+    });
```
Stephen-ONeil, line 70: Personally, I'd move this to the end of `componentDidMount`. It's going to happen async, so it becomes unclear wether the state update has occured\* or not yet during the rest of `componentDidMount` finishes running.

\* actually, I'm not sure if React will flush a state update queue for a component before all of its running lifecycle methods are resolved? Better than raising these in-the-weeds questions is just to only make state updates at the end of a lifecycle method

client/src/panels/panel_declarations/misc/key_concept_panels/KeyConcepts.js
```diff
@@ -0,0 +1,146 @@
...
+    } = this.state;
+
+    const sticky_data = {
+      can_pin: can_pin,
+      pin_pressed: () =>
+        this.setState((prev_state) => {
+          localStorage.setItem("can_pin_key_concepts", !prev_state.can_pin);
+          return { can_pin: !prev_state.can_pin };
+        }),
+    };
```
Stephen-ONeil, line 97: This has to assume `prev_state` at the time of this call back executing is the current state when it was declared, right? Accessing `prev_state` is unnecessary then, you have that state in scope when this call back is declared. could be 
```
pin_pressed: () => {
  localStorage.setItem("can_pin_key_concepts", !can_pin);
  this.setState({ can_pin: !can_pin });
}
```

(and maybe call it `toggle_is_pinned` or something; better to name a function what it does than what it is)

Stephen-ONeil: Side note: manually keeping `localStorage` in sync along with every `setState` is something a future dev updating this component could miss. I'd suggest keeping the `localStorage` updating to a call in a `componentDidUpdate` that resets the `localStorage` value to the state value on every update.

client/src/components/Accordions.js
```diff
@@ -89,10 +89,44 @@ const StatelessPullDownAccordion = ({
   isExpanded,
   children,
   onToggle,
+  showPin,
+  sticky_data, //sticky data only applies if show pin is true
```
Stephen-ONeil, line 93: This is an old component that used inconsistent prop type casing. `show_pin` would be more in-line with current code style.

I'd question grouping the `can_pin` and the `onClick` callback inside `sticky_data`. It obscures what that prop actually is (you have to look through render and see what properties are accessed from it). Plus, it's arguably not really data.

client/src/components/Accordions.js
```diff
@@ -89,10 +89,44 @@ const StatelessPullDownAccordion = ({
...
+          ) : (
+            <IconUnpin
+              height="25px"
+              width="25px"
+              vertical_align="top"
+              alternate_color="false"
+            />
+          )}
+        </button>
+      )}
```
Stephen-ONeil, line 129: None of this belongs in `Accordions`, being pinned isn't an "accordion" concept. 

To be fair, it's not really a `KeyConcepts` concept either. The best approach would be to try an make a higher order component, something like `PinnedContent`, that wraps arbitrary children and gives them sticky position/pinning behaviour. I'm not going that far right now, but the pinning should at least be isolated to `KeyConcepts`.

Maybe the pin icon should float off to the left of the pinned content? Or render it within `KeyConcepts` and do some hacky negative margin stuff, or something, to position it back over the accordion (as long as this isn't a generic wrapper and is only used within `KeyConcepts` to stick an `Accordion`).

Stephen-ONeil (changes requested):
See comments

Got around to looking in `Accordions.js` last, so my comments in there take precedence over any contradictory comments elsewhere. Leaving all comments in though, because I think there's useful general advice in all of them. Let me know if you want clarification on next steps 👍 

Optional other suggestions:
  - [ ] could be good to have it pin itself on a key-up event from the tab key on the pin icon. Could take that as an indication that the user is keyboard navigating, in which case it's taking up screen real-estate for no reason (they have to tab all the way back to it's original position to access it anyway)

Stephen-ONeil:
Oh right, should also:
  - [ ] not show the pin button and disable the pin behaviour in the text-only app. It's not useful to keyboard navigation in general
  - [ ] even if you're disabling in text-only, the button should have the proper attributes to describe it's behaviour to screen readers

client/src/panels/panel_declarations/misc/key_concept_panels/KeyConcepts.js
```diff
@@ -0,0 +1,146 @@
...
+      key_concepts_width: this.keyConceptsContainerRef.current.offsetWidth,
+    });
+
+    const key_concepts_observer = new ResizeObserver((entries) => {
+      this.setState({
+        key_concepts_width: this.keyConceptsContainerRef.current.offsetWidth,
+      });
+    });
+
+    key_concepts_observer.observe(this.keyConceptsContainerRef.current);
```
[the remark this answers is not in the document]
GarvinH: ContainerEscapeHatch ended up making the key concepts a little uglier than i'd like haha.

To clean up the code a little, I found this library https://www.npmjs.com/package/react-resize-detector

GarvinH:
To update BackToTop to use the intersection react component, I'll need hooks which means waiting for the #813 to merge in. Pausing this for now.

Stephen-ONeil:
Hooks have been in since 16.8, looks like we're at 16.7. Can probably bump to 16.8 trivially on this branch in the mean time, while waiting on the 17 update.

Stephen-ONeil:
Alternatively, can you separate the React 17 upgrade from the bootstrap upgrade? Finding and fixing everything that needs to be fixed after rolling bootstrap forward is going to take a while, would like to see less PRs blocking each other so we can keep things moving.

GarvinH:
ok yea i'll move this one to 16.8 for the time being then. the bootstrap branch needs hooks as well

GarvinH:
I just realized that I read the docs wrong... Won't be upgrading to 16.8. I'll be using the polyfill for intersection

client/src/panels/panel_declarations/misc/key_concept_panels/KeyConcepts.js
```diff
@@ -0,0 +1,146 @@
...
+      } catch {
+        can_pin = null;
+      }
+    }
+
+    this.keyConceptsContainerRef = React.createRef();
+
+    this.state = {
+      key_concepts_out_of_view: false,
+      can_pin: _.isBoolean(can_pin) ? can_pin : true,
```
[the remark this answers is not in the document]
GarvinH: I tried removing it but when reading from the localstorage, and empty/non-existant value turns into null (i think cause of the json parse to check if its true or false)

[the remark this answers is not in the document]
Stephen-ONeil: How about `JSON.parse(...) || False` then?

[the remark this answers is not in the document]
GarvinH: 👍

client/src/hoc/PinnedContent.scss
```diff
@@ -0,0 +1,15 @@
...
+
+    &:hover {
+        svg > g {
+            path, rect {
+                fill: $textLightColor;
+                stroke: $textLightColor;
+            }
+        }
+    }
+}
```
Stephen-ONeil, line 15: Look at `_IconWrapper` in `icons.js`. To get hover colours on any of the icon components, just need to pass in an `alternate_color` prop

Stephen-ONeil (changes requested):
- [x] expanding an item with lots of text steals space from the pin button (bar?)
image
- [ ] optional: I think it looks better like this
image
  - that's a quick stab I took at it with `button-unstyled` and `height: 50px; position: absolute;` on the button, and the pin icon coloured white for contrast. It's easy to get it on the left, would be trickier to get it floating on the right of the dropdown (but not impossible)
  - if you do use `position: absolute`, just a reminder to add enough padding to the accordion header text so that it never appears underneath the icon
  - Huh, `.btn-toolbar` gives this whole thing a small negative left margin that puts it slightly out of line with the rest of the content. Overrule that margin to 0, I say
image

Stephen-ONeil (approved):
Approved! 🎉 

Important though! One more feature that occured to me. Please look in to this and either open an issue or a follow up PR to deal with it.

So, if you're linked directly to panel, and you've not previously pinned the keep-in-mind, it will be covering the panel's title and possibly even some important panel content. If you've not been to an infographic before (directly linked to a panel from off-site, or following a home page quick link, etc), you'd not know that there's important missing content under the pinned banner. You may not even necessarily realize that banner's pinned, since you didn't see it pin itself while scrolling past it.

I think the `PinnedContent` component needs a prop to override the `localStorage` to set the initial pinning state. More troublesome will be making the key concept panel recognize when the route was a direct panel link and respond accordingly, I'll leave that up to you.

PR #794, opened 2020-09-29
service stats including general and types

ktw1016:
Paused due to the gov, dept level service stuff being wiped out


server/src/models/services/models.js
```diff
@@ -87,11 +87,21 @@ export default function (model_singleton) {
...
+    ...bilingual_str("id"),
+    ...bilingual_str("label"),
+    value: { type: Number },
+  });
 
   model_singleton.define_model("ServiceReport", ServiceReportSchema);
   model_singleton.define_model("StandardReport", StandardReportSchema);
   model_singleton.define_model("ServiceStandard", ServiceStandardSchema);
   model_singleton.define_model("Service", ServiceSchema);
+  model_singleton.define_model("ServiceGeneralStats", ServiceGeneralStats);
```
AlexCLeduc, line 103: Are we actually populating the `ServiceGeneralStats` and `ServiceTypeStats`  models? I can't see them get used anywhere

ktw1016: Nope they're not!

Issue #598, opened 2020-04-30
Review tables for graphs


Stephen-ONeil:
A couple oddballs in there, like this one:
![image](https://user-images.githubusercontent.com/11301919/80759354-2cbe7c80-8b05-11ea-8e8e-7bb28f819f72.png)

... also "Table of data" is a weak title. This feels too minor for its own issue, just noting it here while I'm looking at it. Maybe "Table for graph" as a generic title on these (or review all panels and pass a proper descriptive title to the modal, but that's possibly more effort than it's worth).

PR #800, opened 2020-10-05
A11y reviews

ktw1016:
Main changes:
1. Bunch of little a11y changes (aria-label, aria-hidden, alt text, heading order, etc)
2. compact_written instead of other formats
3. Landmarks
4. Displaytable changes:
- Moved utils into table, so that screen reader reads table -> utils for table. Instead of Utils for table -> table (Just makes a bit more sense)
- Skip to data util
5. DropdownMenu a11y-ify

### DIFF
--- a/client/src/charts/legends/SelectAllControl.js
+++ b/client/src/charts/legends/SelectAllControl.js
@@ -1,15 +1,8 @@
 import { TrivialTM } from "../../components/index.js";
 
 export const SelectAllControl = ({ SelectAllOnClick, SelectNoneOnClick }) => (
-  <div
-    role="group"
-    aria-label={"SelectAllControlGroup"}
-    style={{ display: "flex", flexDirection: "row" }}
-  >
-    <div
-      id={_.uniqueId("SelectAllControl_SelectText")}
-      style={{ lineHeight: 2 }}
-    >
+  <div role="group" style={{ display: "flex", flexDirection: "row" }}>
+    <div style={{ lineHeight: 2 }}>
       <TrivialTM k="select" />:
     </div>
     <button
--- a/client/src/charts/legends/SelectAllControl.js
+++ b/client/src/charts/legends/SelectAllControl.js
@@ -3,10 +3,13 @@ import { TrivialTM } from "../../components/index.js";
 export const SelectAllControl = ({ SelectAllOnClick, SelectNoneOnClick }) => (
   <div
     role="group"
-    aria-labelledby={"SelectAllControlGroup"}
+    aria-label={"SelectAllControlGroup"}
     style={{ display: "flex", flexDirection: "row" }}
   >
-    <div id={"SelectAllControl"} style={{ lineHeight: 2 }}>
+    <div
+      id={_.uniqueId("SelectAllControl_SelectText")}
+      style={{ lineHeight: 2 }}
+    >
       <TrivialTM k="select" />:
     </div>
     <button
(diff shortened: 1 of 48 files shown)


ktw1016:
Just left out 3 things from Axe testing documented here:
https://docs.google.com/spreadsheets/d/1IGTSfIxblY3RgHekJfplLqLcUWrRkMj1gVJwAo5_Bw0/edit#gid=0

Basically 3 things on:
- typeahead
- aria hidden in rpb
- circle proportion chart alt text

client/src/charts/wrapped_nivo/wrapped_nivo_common.yaml
```diff
@@ -3,5 +3,5 @@ show_table:
   fr: Afficher le tableau de données
 
 default_table_name:
-  en: Table for graph
-  fr: Tableau pour graphique
+  en: Table of data
```
GarvinH, line 6: #598 

This was originally Table of data. Changed it because it was a weak name...
Although table of graph was a pretty weak name in itself lol

Not sure how we should go about this

ktw1016: 😂 so my rationale for changing it to "Table of data" is that, in a11y, screen reader will read this table name first. But there isn't actually a graph for in a11y so it didn't really made sense

client/src/panels/panel_declarations/people/employee_age.js
```diff
@@ -166,31 +166,27 @@ export const declare_employee_age_panel = () =>
...
-                <TabbedContent
-                  tab_keys={["age_group", "avgage"]}
-                  tab_labels={{
-                    age_group: text_maker("age_group"),
-                    avgage: text_maker("avgage"),
-                  }}
-                  tab_pane_contents={{
-                    age_group: (
-                      <div id={"emp_age_tab_pane"}>
-                        <GraphOverlay>
```
GarvinH, line 179: Import can be removed

client/src/charts/legends/SelectAllControl.js
```diff
@@ -3,10 +3,13 @@ import { TrivialTM } from "../../components/index.js";
 export const SelectAllControl = ({ SelectAllOnClick, SelectNoneOnClick }) => (
   <div
     role="group"
-    aria-labelledby={"SelectAllControlGroup"}
+    aria-label={"SelectAllControlGroup"}
```
GarvinH, line 6: Screenreader now says "SelectAllControlGroup", not sure we need this? inside the group it already says "select", "all", and "none"

client/src/panels/panel_declarations/people/employee_age.js
ktw1016: Oh I think removing this was a mistake from rebase-ing

ktw1016: ..Or else I can't remember why I removed GraphOverlay haha

client/src/charts/legends/SelectAllControl.js
ktw1016: yeah I actually don't even think we need aria label here

ktw1016 (commented):
1. Bunch of little a11y changes (aria-label, aria-hidden, alt text, heading order, etc)
2. compact_written instead of other formats
3. Landmarks
4. Displaytable changes:
- Moved utils into table, so that screen reader reads table -> utils for table. Instead of Utils for table -> table (Just makes a bit more sense)
- Skip to data util
5. DropdownMenu a11y-ify

PR #803, opened 2020-10-20
Typescript migration

ktw1016:
Currently having difficulty:
- Trying to set global library like react, lodash, d3, etc. (Complier giving error, code still works somehow)
- Complier can't find yaml's (Complier giving error, code still works somehow)


client/src/components/LabeledTable.tsx
```diff
@@ -1,15 +1,19 @@
...
-export const LabeledTable = ({ title, TitleComponent, contents }) => (
-  <section className="labeled-table" aria-label={title}>
-    {title && (
+interface LabeledTableProps {
+  title: string;
+  contents: any;
+  TitleComponent?: any;
+}
+
+export const LabeledTable = (props: LabeledTableProps) => (
```
AlexCLeduc, line 10: It's possible to still use destructured function args via something like 
```typescript
interface OptionalMiddleName {
    firstName: string;
    middleName?: string;
    lastName: string;
}
function Component({firstName, middleName = "N/A", lastName}:OptionalMiddleName){
    // If middleName wasn't passed in, value will be "N/A"
}
```
No strong opinion one way or the other, but this could cut down the amount of `props.` we need to add

ktw1016: Ah yeah I think this is better

PR #805, opened 2020-10-20
add step to do a pull before deploying

ktw1016:


Stephen-ONeil:
Hmm, what if we specifically needed to push from a local branch that's diverged from the remote repo? I think I've done that when there's a big bug somewhere in the recent commits. Roll back to the last known-good commit and deploy that while things are getting fixed on master.

Instead of always pulling, we could add a check like this:
``
REMOTE_MASTER_SHA = (git rev-parse origin/master | cut -c1-7)
while [ $CURRENT_SHA != $REMOTE_MASTER_SHA ]; do
  read -p "You are deploying from a commit that does not match the head of origin/master. Do you want to continue? [YES/oops]" yn
  case $yn in
    [YES]* ) break;;
    [oops]* ) exit;;
    * ) echo "Please answer YES or oops.";;
  esac
done
```

Maybe add a third case to pull and continue deploying too.

Stephen-ONeil:
I didn't do that as a review because it's more of a discussion. Not saying there's only the one "right way" to do this.

ktw1016:
Yup agreed, I don't think there's need to add third option. I think that'd be a bit unintuitive

Stephen-ONeil (approved):
👍

Issue #739, opened 2020-08-14
Deprecated react lifecycle hooks in dependencies


GarvinH:
The new package-lock.json after adding lodash eslint reveals that react-bootstrap(-typeahead) at minimum need to be updated to use the new lifecycle hooks.

![image](https://user-images.githubusercontent.com/25855114/90283080-855c2000-de3d-11ea-9186-fa56324cca1e.png)

Should wait for #728 to be merged before beginning.

Edit: should also probably wait for #741 to merge as well

GarvinH:
Note: the current react-bootstrap-typeahead has custom implementation done which may not be necessarily addressed in the newer versions.

AlexCLeduc:
I believe the library has fixed this, there should be a [stable version with the unsafe rename](https://github.com/ericgio/react-bootstrap-typeahead/issues/518), and a more recent version that uses [the new lifecycle methods](https://github.com/ericgio/react-bootstrap-typeahead/commit/7d40eee385f04a6bc447e5ff4ca1ac913f9760a7). 

Any reason we can't just bump to version 5? We're on 3 right now

Stephen-ONeil:
It's been locked to 3 for a while because I had to make use of some internal undocumented APIs to get the behaviour we wanted out of the library. Didn't want it rolling forward automatically and breaking in the process. Someone should try rolling forward to 5, checking the release docs for any documented breaking changes. Cross our fingers that those internal APIs didn't change, otherwise it may be hard/impossible to upgrade while keeping the current behaviour.

ktw1016:
This will be a lot more work actually as version 5 drops:
- Bootstrap 3
- React version < 16.8

Not sure about about React upgrade but Bootstrap might be tricky. Related: #84

ktw1016:
This will be a bit more work actually as version 5 drops:
- Bootstrap 3
- React version < 16.8
Not sure about about React upgrade but Bootstrap might be tricky. Related: #84

Stephen-ONeil:
Bootstrap upgrade will be painful. The only thing blocking the React upgrade though, which should be just about trivial otherwise, are the deprecated lifecycles in the typeahead dependency (i.e. this PR, haha).

AlexCLeduc:
I'm confused, what does this issue have to do with lodash-eslint? Why would adding a plugin unrelated to react shed light on a dependency that isn't even covered by our eslint rules?

GarvinH:
[...]

PR #813, opened 2020-11-02
Replacing bootstrap 3 with 4

### LINKED ISSUE #84: Upgrade from bootstrap3 to bootstrap4
Migration docs: https://getbootstrap.com/docs/4.0/migration/

A few annoying things:
  - will need to keep around panel classes from bootstrap3
  - container breakpoints have changed... that's a lot of media queries and hacky JS (like the container escape hatch component) to update
  - a fair bit more probably...

Good news:
  - all Sass, and has Sass maps for grid and container breakpoints, as well as Sass functions for media queries. We should think about adopting these ourselves

GarvinH:
closes #84 

related to #739 

With typeahead no longer holding us back from updating bootstrap, I've went ahead and started the process.
Immediate oddities have been fixed but I'm still searching around to see if there's any issues with the styling.

Notes:

- TreeMap no longer works
- I had to update react in order for react bootstrap to continue working
  - The lifecycle deprecation issues are now back and we're going to have to fix it
- The HTML file has been edited, and we will have to re upload it.


Stephen-ONeil:
Not testcafe related, prod builds are what broke. I see that, in the package lock, there were a number of babel related package changes. Those could be the culprit, have to check if we need to change our weback setup because of it (or if they might require webpack 5 or something, haha). Looks like the webpack update's a potential blocker for this. 

Also, if you run `npm install` with this `package.json` you'll see a LOT of warnings for missing/incorrect peer dependencies. None of them look like they'd be something that would cause prod builds to break while dev builds still work, but I am pretty surprised that things aren't generally far more broken with that many issues.

Stephen-ONeil:
Found some discussion of an error like this related to `uglifyjs-webpack-plugin`, supposedly fixed in 2.2.0 (we're on 2.1.1).

ktw1016:
oh haha and uglifyjs is deprecated now😂 well I guess may as well try for it (btw I'm replacing it with Terser in my webpack branch)

Stephen-ONeil:
Can confirm that updating to `uglifyjs-webpack-plugin` 2.2.0 fixed my local prod builds at least. If you want to keep working on this branch right now then I'd suggest you do that, even if we'll be switching to terser sooner rather than later, so that the route load tests work in the mean time.

ktw1016:
yup working, awesome👍😏

GarvinH:
waiting for typeahead

GarvinH:
I can take this back for the time being

Stephen-ONeil:
Going through, didn't see anything obvious in the standard view (🎉 ), starting to go through with mobile emulation now. Will be posting screenshots here as I go.

Glossary:
  - master
image
  - here
image

Stephen-ONeil:
Haha, maybe it's just the one, nothing else jumping out. Will continue a bit longer and try and check over on IE too, in a bit.

Stephen-ONeil:
- [ ] Ah, glossary tooltips are a bit off. The arrows are missing and I think the font size is smaller.

client/src/InfoBase/index.hbs.html
```diff
@@ -43,12 +43,12 @@
...
           {{/unless}}
           <div class="row">
-            <div class="col-md-4 col-xs-8">
+            <div class="col-md-4 col-8">
               <a href="{{logo_link}}" class="ib-site-header-logo gl-FIP" aria-label="{{logo_a11y_label}}">
                 <object type="image/svg+xml" tabindex="-1" role="img" data="{{CDN_URL}}/svg/sig-blk-{{lang}}.svg" aria-label="{{logo_a11y_label_2}}" class="canada-logo"></object>
               </a>
             </div>
-            <div class="col-md-8 col-xs-4">
+            <div class="col-md-8 col-4">
```
Stephen-ONeil, line 51: Hmm, we could avoid having to re-publish these files if we re-add support for `col-xs-`, at least 4 and 8 (maybe with some big comments explaining that they're for temporary legacy index support, and shouldn't be used elsewhere).

Is removing the divs below necessary, or just cleanup? If it was necessary, could be get around it with some other specific CSS elsewhere?

client/src/components/EmailFrontend.js
```diff
@@ -50,12 +50,12 @@ const EnumField = ({
   selected_enums_for_field,
   disabled,
   state_update_callback,
-  classes,
 }) => (
-  <div className={`${form_type} ${classes}`}>
+  <div className={`${form_type}`}>
```
Stephen-ONeil, line 54: Unnecessary template literal

client/src/components/EmailFrontend.js
```diff
@@ -50,12 +50,12 @@ const EnumField = ({
...
     <label htmlFor={`${field_id}--${enum_key}`}>
       <input
         id={`${field_id}--${enum_key}`}
+        style={{ marginRight: "1rem" }}
```
Stephen-ONeil, line 58: Saw you using rem's, here and elsewhere. I'd not recommend it. rem and em are good if you're careful and consistent in their use, or if they're being used for specific layouts that really should scale with font size changes. I don't think we want some obscure margins shifting around if we ever change the font on the root element though.

client/src/components/EmailFrontend.js
Stephen-ONeil: Unless `form_type` can possibly be undefined? I don't think the email backend allows that, but could always use the `classnames` util to be safe.

client/src/components/EmailFrontend.js
```diff
@@ -220,31 +220,34 @@ class EmailFrontend extends React.Component {
...
-                    )
-                  }
-                />
-              ))}
+              <div
+                className={`${
+                  field_info.style &&
+                  field_info.style === "horizontal" &&
+                  "d-flex justify-content-between"
+                }`}
```
Stephen-ONeil, line 228: Replace template literal with `classnames` util, as is this is going to be setting a `className={"false"}` pretty often. OR add a `|| ""` to the end of the expression, but the `classnames` util is just a little cleaner/used more consistently than that.

client/src/glossary/glossary.js
```diff
@@ -70,7 +70,7 @@ const table_links_by_tag = _.chain(tables)
 //id tag is there for legacy styles
 const Glossary_ = ({ active_key, items_by_letter }) => (
   <div id="#glossary-key">
-    <div className="col-sm-12 col-md-8 col-md-offset-2 font-large">
+    <div className="col-sm-12 col-md-8 offset-md-2 font-large">
```
Stephen-ONeil, line 73: Switch to `fcol`s

client/src/search/EverythingSearch.js
```diff
@@ -258,27 +258,25 @@ const EverythingSearch = withRouter(
...
-              }
-              page_size={page_size}
-              on_query={on_query}
-              on_select={this.onSelect}
-              additional_a11y_description={text_maker(
-                "everything_search_additional_a11y_description"
-              )}
-            />
-          </div>
+        <div className="col-md-12">
```
Stephen-ONeil, line 261: Switch to `fcol`s. I'm going to stop commenting on these if I see others, something we should try and do a general sweep for.

Stephen-ONeil: Oh, hahaha, this was the last file in the diff XD

client/src/InfoBase/index.hbs.html
Stephen-ONeil: If we're keeping them around, the right place to stick them would be `wet-bootstrap3-holdovers.scss`.

Stephen-ONeil:
Just a reminder that we've got a bunch of bootstrap overwrites and holdovers from the previous CSS framework WET (a fork/extended version of bootstrap some GoC folks maintain) in `src/extended_bootstrap_css`. Should check if of the classes being extended in there were deprecated or changed significantly by bootstrap 4, and wether we're even using any of them still. It might largely be stuff used by the home page at this point (in which case it should stay more or less as-is for now).

Doesn't have to be super in-depth though. Cleaning up that, and everything in `src/common_css`, is a whole kettle of fish on its own.

Stephen-ONeil:
One more general issue:
  - [ ] the spacing between the `.well`s in this panel has changed. Very minor here, but I think the RPB and maybe the graph legends also use the `.well` class, so double check that this doesn't mess more with anything else (or figure out what changed and patch it, either one)

IE 11 issues:
  - [ ] Canada flag in header is a weird size (kinda hard to see, haha, the dev link banner covers it)
  - [ ] legend content under the treemap explorer is squished
  - [ ] these IE issues appear to be problems on master too, oops
    - [ ] typeahead options dropdown is extra wide in IE. Other dropdowns (`DisplayTable` menus etc) look fine
    - [ ] typeahead results are extra tall in IE
    - [ ] legends with lots of items overflow, e.g. transfer payments expenditures panel for agriculture

Stephen-ONeil:
This comment still applies https://github.com/TBS-EACPD/infobase/pull/813#issuecomment-776155388

client/src/components/Tombstones.js
```diff
@@ -17,11 +16,14 @@ const UnlabeledTombstone = ({ items }) => (
...
 
 const LabeledTombstone = ({ labels_and_items }) => (
-  <dl className="dl-horizontal tombstone-data-list">
+  <dl className="row tombstone-data-list">
     {_.map(labels_and_items, ([label, item], ix) => (
       <Fragment key={ix}>
-        <dt>{label}</dt>
-        <dd>{item}</dd>
+        <dt className="col-2">{label}</dt>
+        <dd className="col-10">{item}</dd>
```
Stephen-ONeil, line 23: Not handling small screens. Probably want these to be something like `md`, and to give both the label and item 12's for `sm`. Should probably be switched to `frow`/`fcol`s too?
image

There was one other `dl-horizontal` in this PR I think, apply the same fixes there.

client/src/components/WellList.js
```diff
@@ -1,15 +1,14 @@
 import _ from "lodash";
 import React from "react";
 
-
 const WellList = ({ elements }) => (
```
Stephen-ONeil, line 4: Maybe rename to `CardList`, haha, not `well`-based anymore.

Stephen-ONeil (changes requested):
Two more things, plus the glossary tooltips mentioned earlier. After that, I think we should just get one more set of eyes (@ktw1016 ?) to look over everything on the site and we might be good!

ktw1016:
Everything looks good. I just found a minor weird margin issue on Estimates compare. Otherwise LGTM!
Screen Shot 2021-02-12 at 10 59 29 AM

Stephen-ONeil (changes requested):
Found some glossary tooltips with strange placement, please investigate.
image

client/src/models/glossary.js
```diff
@@ -26,7 +26,7 @@ const compiled_definitions = _.memoize((glossary_id) =>
   sanitized_marked(GlossaryEntry.lookup(glossary_id)._def_text)
 );
 
-const glossary_display = (item) => `<div>
+const glossary_display = (item) => `<div aria-live="polite">
```
Stephen-ONeil, line 29: Good idea! I believe we still replace glossary tooltips with glossary links in the text-only site, but it's worth checking if this behaves well enough that we can switch to using tooltips on the text-only version too,

client/src/glossary/TooltipActivator.js
```diff
@@ -1,102 +1,12 @@
 import _ from "lodash";
 import React from "react";
-import Tooltip from "tooltip.js";
+import tippy from "tippy.js";
 
 import { get_glossary_item_tooltip_html } from "../models/glossary.js";
 
-// Patch over Tooltip's _scheduleShow and _scheduleHide to not use setTimeout with a 0 second delay
-// The 0 second delay could still result in the _show call being stuck pending for extended periods (was consistently > 1 second on mobile Chrome)
```
Stephen-ONeil, line 8: Glad to see this stuff go! `tippy.js` seems to work fine on my mobile chrome, nice.

I'd forgot all about this hackyness. If I'd remembered I would have redone it using `patch-package`, or thought to switch tooltip libraries sooner, ha.

client/package.json
```diff
@@ -22,11 +22,12 @@
     "@nivo/line": "^0.52.1",
     "@nivo/pie": "^0.52.1",
     "@nivo/stream": "^0.52.1",
+    "@tippyjs/react": "^4.2.0",
```
Stephen-ONeil, line 25: Unused now, right? Is it a sibling dependency of `tippy.js` or something?

client/src/glossary/TooltipActivator.js
```diff
@@ -183,15 +104,26 @@ const TooltipActivator = _.isUndefined(MutationObserver)
...
+                content: node.getAttribute("data-ibtt-glossary-key")
+                  ? get_glossary_item_tooltip_html(
+                      node.getAttribute("data-ibtt-glossary-key")
+                    )
+                  : node.getAttribute("data-ibtt-text"),
+                interactive: true,
+                duration: 0,
+                allowHTML: true,
+                hideOnClick: false,
+                appendTo: body, //silence the warning
```
Stephen-ONeil, line 126: Repeat of lines 70 to 81, right? Would probably be good to DRY these, just to avoid them going out of sync when future developers make edits.

Stephen-ONeil (changes requested):
Just some other comments on new tooltip related code. Minor stuff, main thing to do is still figure out the oddly placed tool tips from the comment above.

client/src/glossary/TooltipActivator.js
Stephen-ONeil: Ah, haha, wait. Unless I'm missing something I think they already went out-of-sync in 
b32d3a6. Or are they actually supposed to be different configurations:?

client/src/InfoBase/App.js
```diff
@@ -148,7 +148,7 @@ export class App extends React.Component {
             />
             {has_local_storage && <SurveyPopup />}
             <ReactUnmounter />
-            {!is_a11y_mode && <TooltipActivator />}
+            {<TooltipActivator />}
```
Stephen-ONeil, line 151: Super minor, but these curly brackets aren't necessary anymore

client/src/search/EverythingSearch.js
```diff
@@ -185,27 +185,25 @@ const EverythingSearch = withRouter(
...
```
Stephen-ONeil, line 188: IS there a reason it's not an `fcol`?

Stephen-ONeil (approved):
Couple more minor comments to address, but otherwise... approved! 🎉

GarvinH:
Yea, the `componentWillReceiveProps` function is deprecated. I'll look into the components which use it and see if I can switch it out.

The only component I know right away is that Nivo uses this function 😐

Stephen-ONeil:
Oof. We already know updating Nivo will be a whole thing (with all the undocumented API changes... and it will probably mean updating D3 at the same time, so we don't have to load two whole versions of it). If the deprecations are just warnings, that can be it's own issue/future PR.

Please check if FlipMove has a new version that's on React 17, and whether it's a quick enough upgrade to fit in to this PR.

GarvinH:
found a warning in RPB for package `react-aria-modal` which is about for `componentWillMount`.

Seems like there's an open PR for it so we're gonna have to wait a while before the warning is gone:
https://github.com/davidtheclark/react-aria-modal/pull/95

PR #815, opened 2020-11-04
Typeahead replacement

GarvinH:


ktw1016 (approved):
Great👍

Stephen-ONeil:
Give me a minute, taking a look my self.

Stephen-ONeil (changes requested):
Some polish suggestions:
  - [x] filter dropdown should close if something outside receives focus/a click event
     - would solve some weirdness like being able to click back in to the typeahead and get an overlapping dropdown (if anything's typed in), but not being allowed to type
     - less fiddly to close once opened
  - [x] filter dropdown should display over top search results
    - Either tab or mouse navigation is going to bring the text input to the user's attention first, so they'll probably have typed something in before they notice the filter (or maybe they type something in, get too many results, and then want to apply filters). Having it open behind the results makes it unusable once a search has started
    - semi-related, but it might also be worth tweaking the tab order to keyboard navigation hits the filter box first. Fine to leave it on the right for visual users though
  - [x] filter dropdown should include other search configs like glossary entries and RPB tables. It feels weird that you can filter out everything and still get search results for those
    - maybe have those filtered out by default? Certain keywords match a LOT of glossary entries, probably not what people are looking for in general. If there are default filters, that might change how we want any sort of attempts to indicate active filtering to work (maybe it would be indicating that a non-default set of filters is active, rather than indicating that some filtering is active?)
  - [x] maybe some way to indicate that there are active filters applied once the filter dropdown is closed? Something with changing the appearance of the icon, colour of button, or the text in the button maybe (maybe say "add filters" when no filters on, "edit filters" when a filter is active?)
  - [ ] arrow key navigation of typeahead drop down?
  - [x] need to test in IE and on slower devices; the reason I went through all that effort to get pagination working with that old library was because a long list of results got really choppy while scrolling in IE (with it's lack of modern optimizations) or on slow devices in general

I haven't tested with screen reads or looked at code yet, this is just first thoughts from the dev link. Will try and get to code review today though.

GarvinH:
For the arrow key nav, quick nav was used to test it. Not sure if that was enough

Stephen-ONeil:
> quick nav
Referring to tab navigation? Different thing. I'll clarify in a bit, just eating lunch and catching up on email

ktw1016:
Quick comment, tab navigation works but not arrow navigation (without screen reader). Quick navigation is a mode you can turn on when using voiceover. With quick nav on or off, it's fully navigable with screen reader on.

GarvinH:
There's an issue right now where typing a search then changing the filter options may close both menus... Reminder to investigate this

client/src/components/DropdownMenu.js
```diff
@@ -24,46 +49,46 @@ export class DropdownMenu extends React.Component {
...
     const { isOpen } = this.state;
 
+    const aria_label = _.isObject(dropdown_trigger_txt)
+      ? dropdown_a11y_txt
+      : dropdown_trigger_txt;
+
     return (
-      <div className="dropdown">
+      <div className="dropdown" ref={this.dropdown_ref}>
         <button
```
Stephen-ONeil, line 62: The button should have `aria-haspopup="true"` and `aria-expanded={is_open}`

Stephen-ONeil: With `aria-expanded` set you shouldn't need to manually report open/closed in the labels.

client/src/components/DropdownMenu.scss
```diff
@@ -3,6 +3,8 @@
 .dropdown {
```
Stephen-ONeil, line 3: Hmm, these class names didn't come from your commits, but I'm going to comment on them while I'm here.

`dropdown__content__is-open` should be `dropdown__content--is-open`. Better yet though, I think `--is-open` should modify `dropdown` itself.

And all of
```
<button
  ...
>
  {isOpen ? (
    <div className="close-dropdown">
      <span
        aria-label={trivial_text_maker("close")}
        className="close-dropdown__x"
      >
        X
      </span>
      <span className="close-dropdown__x" aria-label={aria_label}>
        {dropdown_trigger_txt}
      </span>
    </div>
  ) : (
    <span aria-label={`${trivial_text_maker("open")} ${aria_label}`}>
      {dropdown_trigger_txt}
    </span>
  )}
</button>
```

Could be 
```
<button
  ...
>
  {dropdown_trigger_txt}
</button>
```
with:
  - `aria-expanded` replacing the closed/open in the label
  - this scss doing everything else:
```
.dropdown--is-open {
  & > button {
   &::before {
      content: "X";
      color: $highlightColor !important;
    }
    &:hover, &:focus {
      color: $tertiaryColor;
    }
  }

  & > .dropdown__content {
    display: block;
  }
}
```

Stephen-ONeil: The `!important` on the X's colour's because of the hover over-ridding it, right? I guess I can't think of a reason for that X to not be red later, so it's fine, but still better generally to avoid `!important`. Almost never the only option.

client/src/extended_bootstrap_css/bootstrap-wet-fixes-extensions.scss
```diff
@@ -273,6 +273,15 @@ div.label {
   border: 2px solid $backgroundColor;
   color: $primaryColor !important;
   border-radius: $standardBorderRadius;
+
+  &--with-icon{/* used for the typeahead filter*/
```
Stephen-ONeil, line 277: Do you think these would be useful elsewhere, or are they too specific to this use case? If they'd be useful elsewhere then you don't need the comment, if they're too tailor-made to the typeahead then move them to local typeahead scss.

`&--modifier` cool! Hadn't seen that use of scss before, handy.

Stephen-ONeil (changes requested):
Not done going through everything, just submitting the comments I've left so far (still need to look at the `Typeahead` code itself).

client/src/search/Typeahead/TypeaheadMenu.js
```diff
@@ -0,0 +1,110 @@
...
+                  <ListGroupItem
+                    key={`header-${group_index}`}
+                    className="dropdown-header"
+                  >
+                    {config_groups[group_index].group_header}
+                  </ListGroupItem>,
+                  <div
+                    key={`group-${group_index}`}
+                    role="group"
+                    aria-label={config_groups[group_index].group_header}
```
Stephen-ONeil, line 69: Screen reader tries to count the groups, but with the current layout that count is wrong. Navigation skips the first and last "group" (the result count and close button respectively), so the first group is announced as "2 of N" and the last is "N-1 of N". The group names are also not read out until after the full name of the first option, which is likely both easy to miss and kind of confusing.

The first and last items should probably not be groups. Pretty sure if the groups themselves are focusable then it will announce the name of the group and give the option to enter in to it or continue to the next group. That'd be better both for announcing group names and navigation, I think.

client/src/search/Typeahead/TypeaheadMenu.js
```diff
@@ -0,0 +1,110 @@
...
+      return (
+        <ListGroup className="rbt-menu dropdown-menu show">
+          <ListGroupItem className="dropdown-header">
+            {text_maker("no_matches_found")}
+          </ListGroupItem>
+        </ListGroup>
+      );
+    } else {
+      return (
+        <ListGroup className="rbt-menu dropdown-menu show">
```
Stephen-ONeil, line 43: `ListGroup` and `ListGroupItem` are probably not the right choices.

Should probably be a `ul` with `role="listbox"`. Everything inside should be in `li`'s, and the result items should have `role="option"`.

Stephen-ONeil (changes requested):
Lots of a11y work left to do here. It's definitely not currently screen reader usable. It can be used from a keyboard (although not through the standard arrow key navigation a dropdown like this normally uses), but a screen-reader only user is not given nearly enough context to navigate it.

For instance, as you type in to the text input, there is no indication at what point a list of results has appeared. Need to alert them when the results menu opens and update them on the number of matched results as they type. See my code comments, I address a way to do that at some point

Further screen reader usage notes (in addition to code comments):
  - [x] ~because the first thing within the header to take focus is the add filter button, it reads the add filter content and then goes back and reads the page h1 and h2. Confusing, maybe add focus to both of those, should be more clear then.~
    - with Voiceover at least, only a problem when tab navigating actually. Assuming screen reader users tend to use their AT's own keyboard navigation, so this is fine (pending ever getting more screen readers to test with)
  - [x] no context given to the screen reader user on what these filters are for. Maybe I was wrong saying that they should be the first thing the screen reader hits. Maybe it should land in the text input, with part of the help text being a note that there are filter options that can be applied to the results
  - [x] never explains what is going to happen if an item from the search is selected. They're also always buttons, even when their behaviour is to act as links. Not a problem that can be solved in the `Typeahead` itself, since it will depend on use case, but at least the advanced search should explain that more clearly in some of it's aria text

Stephen-ONeil:
I implied at some point that arrow key navigation would come for free with the right `role` and aria attributes. It will for some screen readers in certain modes, but to get consistent arrow key navigation for everyone it's going to take implementing it ourselves, I think.

General note on all this a11y feedback, I'm only pulling some of this off the top of my head. I'm doing a lot of research here (part of why these reviews take so long, sorry). Lots of googling for accessibility best practices/reading MDN docs etc., and even more hunting around for actual typeahead components in the wild to test with a screen reader/pick apart in the dev console. The previous library's a good place to start, although we definitely shouldn't aim to rebuild it exactly (lots of things we can tailor to our own needs).

client/src/components/DropdownMenu.scss
GarvinH: seems like the `!important` isn't needed!

client/src/search/Typeahead/Typeahead.js
```diff
@@ -0,0 +1,229 @@
...
+            placeholder={placeholder}
+            aria-label={placeholder}
+            onChange={update_search_text}
+            ref={this.typeaheadRef}
+            onKeyDown={this.handleKeyDown}
+            value={search_text}
+            onFocus={() => this.setState({ can_show_menu: true })}
+            onKeyDown={this.handle_key_down}
+          />
+        </div>
```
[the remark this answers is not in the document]
GarvinH: the voiceover likes to overlap the content it reads....

"72 results are f-"(status text) "tre"(search text)

tried putting aria live to assertive but doesn't seem to do much. any tips?

client/src/search/Typeahead/TypeaheadMenu.js
GarvinH: Seems like all children of `ul` are to be counted as their own groups...

I'll have to look into workarounds for this.

client/src/search/Typeahead/Typeahead.js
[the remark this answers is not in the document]
Stephen-ONeil: Huh, polite is supposed to wait and atomic is supposed to prevent it from being read in pieces. Nothing off the top of my head, give it a google. If you don't find anything let me know and I'll look in to it myself.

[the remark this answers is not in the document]
GarvinH: I think it reads it, but the input aria is a lot more assertive? I've tried using aria-disable on it but to no avail. Rather than being read in pieces, it straight up stops...

I've found a post on SO similar to what I'm facing except its broken in chrome for me as well...
https://stackoverflow.com/questions/62882886/safari-voiceover-issue-when-reading-input-field

GarvinH:
I investigated the code for the react typeahead bootstrap package (the one currently in master).
They do not have any status reading.
What they do have is a readout of the first typeahead recommendation (although it doesn't work consistently) and that one also gets cut off immediately.

Also they sacrifice tab navigation with arrow key navigation. Is this something we'd rather have? When I tried doing it earlier, i couldn't figure out how to get the scrolling to match with the "active" element. But if this is preferable I'll definitely investigate more into it.

Stephen-ONeil:
This example seems to have everything working nicely: https://alphagov.github.io/accessible-autocomplete/examples/form.html

GarvinH:
searching for a large group such as "can" can take between 1 sec (on the macbook pro) to at least 5 seconds (ie on the surface). I've reimplemented the pagination.

Stephen-ONeil:
Should have mentioned it earlier, but there's also [react-virtualized](https://github.com/bvaughn/react-virtualized)

It works by only inserting the currently visible elements in to the DOM, and just enough of the overflow (non-visible) elements above/below that scrolling is still smooth. Adds/removes elements as you scroll.

I think pagination's good enough for this, just mentioning for completeness.

GarvinH:
Currently investigating react-virtualized/react-window. May or may not just open a new issue for that one.

Stephen-ONeil:
By the way, was deleting the client package lock intentional? Please generate and commit a fresh one.

AlexCLeduc:
I can't get the "show more items" button to work, it crashes the whole app. You can reproduce the issue by searching for "canada" on the home page and clicking the "show 30 more" button

ktw1016:
Just started looking into this. I see a lot of work has been put in to make arrow navigation work and that's great! Really like the `aria-live` giving feedback to type in 3 more or characters. (Hmm.. but where did "Enter" button event to trigger the focus to go into search result?😔) Also the Dropdown Menu functionality to close the menu when clicking anywhere else is awesome.

However I have quite a lot of problems playing around with these new updates on the a11y front(sorry, only tried it now after all these new commits):
- While keyboard navigation is working now, it seems like tab navigation is broken
- I was able to use it decently enough before with VoiceOver (Mac screen reader). Now, it is unusable for me. I'm not sure if I'm using it incorrectly, and I'm wondering how you guys are able to use it with a screen reader? I'm suspecting broken tab navigation broke the whole screen reader experience (at least for me)

What is broken on VoiceOver (for me):
- "list box" (search result box) doesn't have description
- Seems like voiceover cursor (or focus) doesn't move with the arrow keys. Although it is reading them out
- Hitting "Enter" on an item opened gov infographic, or opened up filter Dropdown menu.

ktw1016:
Minor fixes (reminder to myself to implement):
- [ ] Better filter button "X" position?
- [ ] Type something into search box, use arrow key to "Close Menu", keep typing to search. Search doesn't happen

Stephen-ONeil (changes requested):
Haven't fully gone through everything, might have more later, but here are some comments to start on.

In addition to the code comments:
  - [ ] merge `AdvancedSearch` in to `EverythingSearch`. Only place left directly using `EverythingSearch` is panel inventory, and there's no reason it shouldn't also have the filter options
    - this'll be great, because the two files were always a mess with those comments reminding us to manually keep the options in sync between the two
    - [ ] just in case we might still want an `EverythingSearch` without a filter later, maybe add a prop to disable the filter while merging the two
  - [ ] bug, can't recreate consistently but toggling "filter" options while a search is active occasionally, but not always, closes the filter menu? 
    - some of my other comments involve changing the filter controls somewhat, might be worth doing those first, bug might go away in the process if things are rewritten enough?
  - [ ] bit of a tangent, but why are the close buttons inside of `DropdownMenu`s unstyled?

client/src/home/home.js
```diff
@@ -112,6 +124,7 @@ const HomeLayout = (props) => (
                 MISC2: query,
               });
             }}
+            include_gov={false}
```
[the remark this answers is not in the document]
Stephen-ONeil: Maybe it felt redundant to the three big links that take you to gov infographics already? I think it'd be better to put it back in, since the search bar is above those and a user won't necessarily know those links go to the gov level unless they've already used them before.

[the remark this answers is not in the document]
ktw1016: yeah agreed

client/src/components/DropdownMenu.scss
```diff
@@ -13,14 +33,11 @@
...
-}
 .close-dropdown {
   &:hover, &:hover .close-dropdown__x {
     color: $tertiaryColor;
   }
 }
 .close-dropdown__x {
-  color: $highlightColor;
+  color: $highlightColor !important;
 }
```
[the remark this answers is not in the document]
ktw1016: good catch

client/src/extended_bootstrap_css/bootstrap-wet-fixes-extensions.scss
```diff
@@ -373,6 +400,21 @@ li.list-group-item {
   color: lighten($tertiaryColor, 30%);
 }
 
+.list-group-item.disabled {
+  background-color: $backgroundColor !important;
```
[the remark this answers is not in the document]
ktw1016: it seems like this section can just be wiped. Checked over where it's used and nothing seem affected. Let me know if removing this cause problems

client/src/search/Typeahead/Typeahead.js
```diff
@@ -0,0 +1,577 @@
...
+          {filter_content && (
+            <DropdownMenu
+              dropdown_trigger_txt={
+                <div
+                  style={{
+                    textAlign: "start",
+                    whiteSpace: "nowrap",
+                    display: "inline-block",
+                  }}
+                >
```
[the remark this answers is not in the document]
ktw1016: Okay, contemplated if I should make `utility_buttons` a prop inside `AdvancedSearch` but decided it wouldn't be useful since it needs state management at AdvancedSearch level

GarvinH:
Trivial problem: the sentences the voiceover reads when selecting typeahead runs on

> [...] make a selection.Selecting a result [...]

I tried to do a string concat with a space, but it only fixes it on FFox.

client/src/search/Typeahead/Typeahead.scss
```diff
@@ -1,124 +1,110 @@
...
-.rbt-menu.dropdown-menu > li:hover:not(.dropdown-header),
-.rbt-menu.dropdown-menu > li > a:focus,
-.rbt-menu.dropdown-menu > li > a:hover {
-  &:not(.disabled){
-    background-color: lighten($tertiaryColor, 30%);
-    color: $linkFocusColor;
-    text-decoration: underline;
-  }
+.typeahead__control {
+  text-align: center;
```
ktw1016, line 96: Is there a reason for this to be separated?

client/src/search/Typeahead/Typeahead.js
```diff
@@ -287,239 +122,366 @@ export class Typeahead extends React.Component {
...
+                        k="paginate_previous"
+                        args={{ page_size: page_size }}
+                      />
+                    </a>
+                  </li>
+                )}
+                {_.chain(results_on_page)
+                  .groupBy("config_group_index")
+                  .thru((grouped_results) => {
+                    let index_key_counter = needs_pagination_up_control ? 1 : 0;
```
ktw1016, line 193: Something minor but this can be `const` I think

ktw1016 (approved):
Impressive..!! Please merge it in if you're done! 🚀

Stephen-ONeil:
> Trivial problem: the sentences the voiceover reads when selecting typeahead runs on
> 
> > [...] make a selection.Selecting a result [...]
> 
> I tried to do a string concat with a space, but it only fixes it on FFox.

Hmm, good catch! I've stuck a `br` in there now, screen readers should always read the additional description as a new sentence now, or potentially a new paragraph. Either seems fine.

client/src/search/Typeahead/Typeahead.scss
Stephen-ONeil: I figured they might diverge more at some point. I also wanted to keep the terminology around the menu clear, because we've got both controls and results side-by-side in it (and awkwardly sharing things like the selection cursor).

PR #817, opened 2020-11-04
Better GCF logging

Stephen-ONeil:
TODO:
  - [x] better use express' error handling middleware, rather than sprinkling logs around
  - [x] check for DB connection on each request, and attempt to reconnect 
  - [x] update mongoose (chance this was all a bug in the version we're on), maybe mongo connection strings too
    - there's a closed issue about the same sort of intermittent timeout errors with mongoose when using `useUnifiedTopology: true`. Not closed because of a specific fix, but because mongoose has updated to a newer version of the mongo driver since
    - if updating mongoose fixes nothing, we could consider disabling `useUnifiedTopology: true` (and just live with the deprecation warning for now) 
  - [x] apply a similar approach to the email backend server
    - [x] does the client log timing/errors for email backend requests? Maybe I should add that too. **Edit:** just logging errors, don't feel that timing is critical here
  - ~change how the client responds to errors in data API requests. Should probably retry once and then display an actual error page, rather than hanging forever (at least by default)~
    - ~hmm, or not. Since we've got a lot of mixed routes that might mostly work without the API right now, it could result in the site appearing to work up to the max timeout length(s x retries) and then suddenly going to the error page. That'd be a pretty rough experience, especially if the user cared about some data not coming from the API~
    -  ~we could make all routes do a server health check during the initial spinner, and always error if it's not functioning. Actively inviting the opportunity for a self-created full site outage, but maybe that's better than an inexplicable degraded experience?~
    - eh, out-of-scope for now. The client's use of GraphQL is unstable until the full migration happens, not the right time for more complexity
  - [x] bunch of linting in `/server` and `/email_backend`


server/src/app.js
```diff
@@ -1,62 +1,76 @@
...
 
-app.use("/", function (req, res, next) {
-  try {
-    res.header("cache-control", "public, max-age=31536000");
-    res.header("Access-Control-Allow-Origin", "*");
-    res.header(
-      "Access-Control-Allow-Headers",
-      "Content-Type, Authorization, Content-Length, X-Requested-With, encoded-compressed-query"
-    );
-    if (req.method === "OPTIONS") {
```
ktw1016, line 33: Might be a stupid question but why was this here and why are you removing it?

email_backend/src/email_backend.js
```diff
@@ -1,41 +1,47 @@
...
+        error_message,
+        request_content,
+      }),
+      sha: process.env.CURRENT_SHA || "dev, no sha env var set",
+    })
+  );
+};
+const log_success_case = (request) => {
+  const request_content = get_request_content(request);
+  console.error(
```
ktw1016, line 40: This shouldn't really be console.error, it's logging success case. Could be. .log or .info?

ktw1016 (approved):
Left few comments, looks good to me otherwise !

email_backend/src/email_backend.js
Stephen-ONeil: Oops, typo, thanks!

server/src/app.js
Stephen-ONeil: Haha, I think it was there for a stupid reason. That or it was a totally out of date reason. This was all CORS header configuration, and handling CORS pre-flight requests (`OPTIONS`). Well, a couple lines up we had `email_backend.use(cors())`... and I happened to glance at the docs for that `cors` middleware during this PR and realized that:
  A) by default, it doesn't call `next()` after handling `OPTIONS` requests, so those were never reaching this `if` block
  B) it has config parameters for all the CORS headers we were manually setting here (plus, if `OPTIONS` never get past it, we were never actually properly sending these headers in the pre-flight response)

PR #818, opened 2020-11-05
remove react-share and have custom share button

ktw1016:


client/src/components/ShareButton.js
```diff
@@ -47,6 +51,12 @@ export class ShareButton extends React.Component {
       icon_size,
     } = this.props;
 
+    const common_icon_props = {
+      width: "50px",
+      height: "50px",
+    };
+    const hashtag_replaced_url = _.replace(url, "#", "%23");
```
Stephen-ONeil, line 58: I think you want
```
const fully_encoded_url = encodeURIComponent(url);
```

(note `encodeURI` URL encodes a string with the exception of semanticly meaningful URL characters like ? or #. `encodeURIComponent` encodes everything)

Stephen-ONeil (changes requested):
One small change, approved after that.

One note about doing this ourselves is that we'll have to update those strings if Reddit etc. change their share URLs. They probably won't (who knows how many links it'd break) but at least with a library it was someone else's job to pay attention to that. Like I said though, probably fine.

PR #819, opened 2020-11-05
Fix eslint across project

Stephen-ONeil:
A mix of bad configs/missing dependencies had eslint not working outside of the `/client` dir (the root config set the parser to `babel-eslint` but only the client imported it, other sub repos only important `eslint`).

`babel-eslint` (and pretty much all other eslint related dependencies) now imported in the root package.json, sub-projects now share that one copy along with much more of the config settings.

Module import and lodash rules will now be applied in the server and email backend. Not going to fix all those warnings right now though, they'll mostly auto-clean-up whenever those files are touched going forward.


ktw1016:
okay nice, I'll just have to rebase on my webpack5 branch and update it back to eslint 7. Having it all coming from root sounds good to me (and should be easier to update)

ktw1016 (approved):

PR #820, opened 2020-11-06
Webpack5

### LINKED ISSUE #808: Use different eslint rules for prod builds than in dev
Specifically the `no-debugger` rule. Currently, the project settings run eslint fix on save, so to have a debugger statement during dev you HAVE to set the line to be ignored by eslint... at which point why have the rule? 

We should be able to use debugger statements in dev without flagging them as ignored. Only prod builds should complain about them being left in.

### LINKED ISSUE #811: Upgrade to Webpack 5
https://webpack.js.org/blog/2020-10-10-webpack-5-release/

Likely going to be a bit of a fiddly upgrade, but lots of great bundle optimization benefits to be had. May even speed up build times with the new caching options!

Skimming the migration docs:
  - lots of webpack config options have changed, will have to go through those
  - some `src` code might need to change (I think the `.side-effects.js` prefix is no longer necessary, named chunks work differently, ~async modules work differently~ oh, we didn't have any, etc.)
  - I haven't looked in to whether the webpack plugins we use need/have webpack 5 compatible upgrades of their own

Who ever does this (fine with it being me) will have to look through the migration docs carefully.

ktw1016:
closes #811 
closes #808


client/package.json
```diff
@@ -8,32 +8,32 @@
...
-    "apollo-cache-inmemory": "^1.4.2",
-    "apollo-client": "^2.4.12",
-    "apollo-link-http": "^1.5.9",
     "axios": "^0.19.2",
-    "babel-loader": "^8.0.5",
+    "babel-loader": "^8.1.0",
     "bootstrap": "^3.4.1",
     "bundle-stats": "^1.10.1",
-    "circular-dependency-plugin": "^5.0.2",
+    "bundle-stats-webpack-plugin": "^2.3.0",
```
Stephen-ONeil, line 29: You added `bundle-stats-webpack-plugin`, but it looks like the webpack config still imports from `bundle-stats`?

Bundle stats output is junked either way:
http://dev.ea-ad.ca/webpack5/bundle-stats.html

(might be that the stats snapshot it wants to compare to from master isn't compatible with webpack 5 stats output eitherway, but I still think it might be that the new plugin isn't actually being used)

client/src/models/organizational_entities.js
```diff
@@ -201,6 +201,8 @@ const Dept = class Dept extends static_subject_store_with_API_data() {
       } else {
         return "separate_agencies";
       }
+    } else {
+      return undefined;
```
Stephen-ONeil, line 205: Huh, did the new lint rules suggest this?

When execution reaches the end of a function's body without an explicit return, it will return undefined by default.

dockerfiles/build/Dockerfile
```diff
@@ -25,9 +25,12 @@ ENV PATH $SERVICE_HOME/google-cloud-sdk/bin:$PATH
 USER root
 
 # image specific
-RUN apk add --no-cache \
+RUN apk add --no-cache --virtual .gyp \
+  make \
   openssh \
-  make
+  g++ \
+  && npm install \
```
Stephen-ONeil, line 32: What node modules are installed in the docker images?

.circleci/config.yml
```diff
@@ -159,7 +159,7 @@ jobs:
             - prod-build-{{ .Branch }}-{{ .Revision }}
       - run:
           command: | # if the build job starts failing with exit code 137 (running out of memory), add the build_all.sh option -c "half" to cut the concurrency
-            [ -f "client/$BUILD_DIR/InfoBase/app/app-a11y-en.min.js" ] || (cd client && ./deploy_scripts/build_all.sh -m 256 -c "half")
+            [ -f "client/$BUILD_DIR/InfoBase/app/app-a11y-en.min.js" ] || (cd client && ./deploy_scripts/build_all.sh -m 256 -c "none")
```
Stephen-ONeil, line 162: Was it still running out of memory even with the final, fixed, version of `build_all.sh` and `half`?

Stephen-ONeil (changes requested):
Questions mostly, maybe one thing to fix. Gonna check out the branch and look at things locally in a bit too.

client/src/models/organizational_entities.js
ktw1016: yeah exactly, it was `getter-return`, which is a rule now included in eslint recommended list. I just simply put return undefined to exactly mimic the function and resolve the linter

dockerfiles/build/Dockerfile
ktw1016: Shouldn't be there!

.circleci/config.yml
ktw1016: Yup half runs out of memory often

Stephen-ONeil:
Reminder to me: just be bundle stats left to get working, take a look at that

ktw1016:
Filed an issue here for bundle-stats: https://github.com/relative-ci/bundle-stats/issues/1067

PR #821, opened 2020-11-06
Drop budget

Stephen-ONeil:
No more budget data, no more budget presentation.


Stephen-ONeil:
Trivial enough, self approved.

PR #823, opened 2020-11-12
Tweak memory usage of CirclCI build job

Stephen-ONeil:
Ugh, this alpine linux image has a real weird implementation of `ps`, couldn't find the right man pages for it and it's help command is useless. A bit of trial and error though and I landed on the following for a decent logging of memory usage during the build job:

```
- run:
          command: |
            while true; do
              sleep 5
              echo "====="
              ps -eo comm,vsz,rss | sort -k 3 -nr
              echo "====="
            done 
          background: true
```

(sorting's not perfect because the output mixes units, ah well)
image

Ignore the `sh`. Those four nodes are the parallel prod builds. First memory column is "vsz", basically the max amount the system will let the process use. Second is actual usage. Whole container only has 4GB available though and will exit if it's all used (CircleCI doesn't allow swap/virtual memory, boo).

Not clear what changed to push the build jobs to suddenly start eating up more than a gig each. Individually, they still barely hit 1 gig in general though, so when it's failing it's only just barely. 

Options:
  1) stop running all four build in parallel
  2) dig in to the node/V8 config options, may be able to limit the individual memory limits (but will the builds be able to run to completion with limited memory? Gonna take some testing)

**Edit:** so what I've done is add some new options to the InfoBase `build_all.sh` script. One for choosing build concurrency ("full" for all four at once, "half" to run the two standard and two a11y in pairs, and "none" to run all four in series), and one's just an easy way to pass down the `max_old_space_size` value to the node process for each build.

Setting a `max_old_space_size` of 256 is maybe enough to get us back under memory budget (default is 512). `max_old_space_size` doesn't set a hard limit, it just tells V8 when to start spending time on garbage collection. If the build process actually need all the memory they've been using then it can't help us. Worst case we start using the "half" flag in CI and live with a slightly longer build step.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -148,7 +148,7 @@ jobs:
           keys:
             - prod-build-{{ .Branch }}-{{ .Revision }}
       - run:
-          command: |
+          command: | # if the build job starts failing with exit code 137 (running out of memory), add the build_all.sh option --concurrency="half"
             [ -f "client/$BUILD_DIR/InfoBase/app/app-a11y-en.min.js" ] || (cd client && ./deploy_scripts/build_all.sh --max_old_space_size=256)
       - save_cache:
           paths:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -159,7 +159,7 @@ jobs:
             - prod-build-{{ .Branch }}-{{ .Revision }}
       - run:
           command: |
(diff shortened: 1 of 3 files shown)


ktw1016:
Hmm and this actually passed the point where it fails, and failed at a later point

ktw1016 (approved):

PR #824, opened 2020-11-12
ESLint conditional rules

ktw1016:


ktw1016:
This will be completed in #820 as it will need to be done in `eslint-webpack-loader`, instead of `eslint-loader`. because it's deprecated and already updated in #820

PR #826, opened 2020-11-18
Handle broken infograph links

### LINKED ISSUE #443: More gracefully handle traffic to dead infographics

### LINKED ISSUE #822: Handle broken links better
If the panel part of the link is broken, just direct them to whatever infographic page it should be

If the whole thing is broken, just redirect them to #home

Maybe a banner to let them know they were redirected because link is broken

ktw1016:
closes #822 
closes #443 

Errors to try:
#orgs/ERROR/1/infograph/financial
#orgs/program/AGR-ERROR/infograph/financial
#orgs/crso/FO-ERROR/infograph/financial
#orgs/dept/ERROR/infograph/financial
#orgs/dept/1/infograph/financial/.-ERROR.-(panel_key.-.-'welcome_mat)
#orgs/dept/1/infograph/ERROR

### DIFF
--- a/client/src/InfoBase/App.js
+++ b/client/src/InfoBase/App.js
@@ -15,9 +15,8 @@ import { ErrorBoundary } from "../core/ErrorBoundary.js";
 import { has_local_storage } from "../core/feature_detection.js";
 import { InsertRuntimeFooterLinks } from "../core/InsertRuntimeFooterLinks.js";
 import { ReactUnmounter } from "../core/NavComponents.js";
-import { get_session_storage_w_expiry } from "../general_utils.js";
+import { RedirectHeader } from "../core/RedirectHeader.js";
 import { TooltipActivator } from "../glossary/TooltipActivator.js";
-import { create_text_maker } from "../models/text.js";
 import { SurveyPopup } from "../Survey/SurveyPopup.js";
 
 import { app_reducer } from "./AppState.js";
@@ -28,9 +27,6 @@ import {
 } from "./common_app_component_utils.js";
 
 import "./App.scss";
-import text from "./App.yaml";
-
-const text_maker = create_text_maker(text);
 
 const Home = retrying_react_lazy(() => import("../home/home.js"));
 const A11yHome = retrying_react_lazy(() => import("../home/a11y_home.js"));
@@ -90,7 +86,6 @@ export class App extends React.Component {
   }
 
   state = {
-    show_redirected_msg: get_session_storage_w_expiry("redirected_msg"),
     outage_message: null,
     showSurvey: false,
   };
@@ -119,7 +114,7 @@ export class App extends React.Component {
   }
 
   render() {
-    const { outage_msg, showSurvey, show_redirected_msg } = this.state;
+    const { outage_msg, showSurvey } = this.state;
     return (
       <div
         tabIndex={-1}
@@ -141,21 +136,10 @@ export class App extends React.Component {
                 }}
               />
             )}
-            {show_redirected_msg && (
-              <HeaderNotification
-                list_of_text={[
-                  text_maker("common_redirect_msg", {
-                    url: get_session_storage_w_expiry("pre_redirected_url"),
-                  }),
-                  get_session_storage_w_expiry("redirected_msg"),
-                ]}
-                hideNotification={() => {
-                  sessionStorage.removeItem("pre_redirected_url");
-                  sessionStorage.removeItem("redirected_msg");
-                  this.setState({ show_redirected_msg: false });
-                }}
-              />
-            )}
+            <RedirectHeader
+              redirect_msg_key="redirected_msg"
+              url_before_redirect_key="pre_redirected_url"
+            />
             {has_local_storage && <SurveyPopup />}
             <ReactUnmounter />
             {!window.is_a11y_mode && <TooltipActivator />}
--- a/client/src/InfoBase/App.js
+++ b/client/src/InfoBase/App.js
@@ -1,7 +1,7 @@
 import axios from "axios";
 
 import { Suspense } from "react";
-import { connect, Provider } from "react-redux";
+import { Provider } from "react-redux";
 import { Route, Switch, Redirect } from "react-router-dom";
 import { createStore } from "redux";
 
@@ -77,7 +77,7 @@ const Survey = retrying_react_lazy(() =>
 
 const store = createStore(app_reducer);
 
-class _App extends React.Component {
+export class App extends React.Component {
   constructor() {
     super();
     initialize_analytics();
@@ -116,11 +116,6 @@ class _App extends React.Component {
 
   render() {
     const { outage_msg, showSurvey, show_redirected_msg } = this.state;
-    const { redirect_target, redirect_msg } = this.props;
-    const store_redirect_target = store.getState().redirect_target;
-    if (store_redirect_target) {
-      return <Redirect to={`/${store_redirect_target}`} />;
-    }
     return (
       <div
         tabIndex={-1}
@@ -142,9 +137,9 @@ class _App extends React.Component {
                 }}
               />
             )}
-            {redirect_msg && (
+            {show_redirected_msg && (
               <HeaderNotification
-                text={redirect_msg}
+                text={sessionStorage.getItem("redirected_msg")}
                 hideNotification={() => {
                   sessionStorage.removeItem("redirected_msg");
                   this.setState({ show_redirected_msg: false });
@@ -257,10 +252,3 @@ class _App extends React.Component {
     );
   }
 }
-
-const mapStateToProps = (state) => ({
-  redirect_target: state.redirect_target,
-  redirect_msg: state.redirect_msg,
-});
-
-export const App = connect(mapStateToProps)(_App);
--- a/client/src/InfoBase/App.js
+++ b/client/src/InfoBase/App.js
@@ -1,7 +1,7 @@
 import axios from "axios";
 
 import { Suspense } from "react";
-import { Provider } from "react-redux";
+import { connect, Provider } from "react-redux";
 import { Route, Switch, Redirect } from "react-router-dom";
 import { createStore } from "redux";
 
@@ -77,7 +77,7 @@ const Survey = retrying_react_lazy(() =>
 
 const store = createStore(app_reducer);
 
-export class App extends React.Component {
+class _App extends React.Component {
   constructor() {
     super();
     initialize_analytics();
@@ -116,6 +116,11 @@ export class App extends React.Component {
 
   render() {
     const { outage_msg, showSurvey, show_redirected_msg } = this.state;
+    const { redirect_target, redirect_msg } = this.props;
+    const store_redirect_target = store.getState().redirect_target;
+    if (store_redirect_target) {
+      return <Redirect to={`/${store_redirect_target}`} />;
+    }
     return (
       <div
         tabIndex={-1}
@@ -137,9 +142,9 @@ export class App extends React.Component {
                 }}
               />
             )}
-            {show_redirected_msg && (
+            {redirect_msg && (
               <HeaderNotification
-                text={sessionStorage.getItem("redirected_msg")}
+                text={redirect_msg}
                 hideNotification={() => {
                   sessionStorage.removeItem("redirected_msg");
                   this.setState({ show_redirected_msg: false });
@@ -252,3 +257,10 @@ export class App extends React.Component {
     );
   }
 }
+
+const mapStateToProps = (state) => ({
+  redirect_target: state.redirect_target,
+  redirect_msg: state.redirect_msg,
+});
+
+export const App = connect(mapStateToProps)(_App);
--- a/client/src/InfoBase/App.js
+++ b/client/src/InfoBase/App.js
@@ -125,11 +125,7 @@ export class App extends React.Component {
         }`}
       >
         <Provider store={store}>
-          <ErrorBoundary
-            set_redirect_msg={(msg) =>
-              sessionStorage.setItem("redirected_msg", msg)
-            }
-          >
+          <ErrorBoundary>
             <DevFip />
             <InsertRuntimeFooterLinks />
             <EasyAccess />
--- a/client/src/InfoBase/App.js
+++ b/client/src/InfoBase/App.js
@@ -5,7 +5,7 @@ import { Provider } from "react-redux";
 import { Route, Switch, Redirect } from "react-router-dom";
 import { createStore } from "redux";
 
-import { HeaderNotification } from "../components/HeaderNotification";
+import { HeaderNotification } from "../components/HeaderNotification.js";
 import { PageDetails } from "../components/PageDetails.js";
 import { SpinnerWrapper } from "../components/SpinnerWrapper.js";
 import { initialize_analytics } from "../core/analytics.js";
@@ -86,13 +86,9 @@ export class App extends React.Component {
   }
 
   state = {
-    showNotification: false,
-    message: "",
+    show_redirected_msg: sessionStorage.getItem("redirected_msg"),
+    outage_message: null,
     showSurvey: false,
-  };
-
-  hideNotification = () => {
-    this.setState({ showNotification: false });
   };
 
   toggleSurvey = (override = null) => {
@@ -111,8 +107,7 @@ export class App extends React.Component {
           const data = res.data;
           if (data.outage) {
             this.setState({
-              showNotification: true,
-              message: data[window.lang],
+              outage_msg: data[window.lang],
             });
           }
         });
@@ -120,6 +115,7 @@ export class App extends React.Component {
   }
 
   render() {
+    const { outage_msg, showSurvey, show_redirected_msg } = this.state;
     return (
       <div
         tabIndex={-1}
@@ -129,14 +125,29 @@ export class App extends React.Component {
         }`}
       >
         <Provider store={store}>
-          <ErrorBoundary>
+          <ErrorBoundary
+            set_redirect_msg={(msg) =>
+              sessionStorage.setItem("redirected_msg", msg)
+            }
+          >
             <DevFip />
             <InsertRuntimeFooterLinks />
             <EasyAccess />
-            {this.state.showNotification && (
+            {outage_msg && (
               <HeaderNotification
-                text={this.state.message}
-                hideNotification={this.hideNotification}
+                text={outage_msg}
+                hideNotification={() => {
+                  this.setState({ outage_msg: null });
+                }}
+              />
+            )}
+            {show_redirected_msg && (
+              <HeaderNotification
+                text={sessionStorage.getItem("redirected_msg")}
+                hideNotification={() => {
+                  sessionStorage.removeItem("redirected_msg");
+                  this.setState({ show_redirected_msg: false });
+                }}
               />
             )}
             {has_local_storage && <SurveyPopup />}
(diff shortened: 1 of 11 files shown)


client/src/core/ErrorBoundary.js
```diff
@@ -1,126 +1,199 @@
...
-    this.setState({
-      testing_for_stale_client: false,
-    });
+      if (_.isNull(error)) {
+        return this.props.children;
+      }
+      if (testing_for_stale_client) {
+        this.catch_stale_client_error_case();
+        return null;
+      } else {
```
ktw1016, line 94: For ease of reading, the meaningful part that changed start here

client/src/core/ErrorBoundary.js
```diff
@@ -1,126 +1,199 @@
...
+            });
+          } else {
+            // subject doesn't exist, redirect to home
+            window.location.replace("#home");
+            window.location.reload();
+          }
+        } else if (infograph_regex_without_panel) {
+          // Without panel part, url will only fail with subject not matching
+          window.location.replace("#home");
+          window.location.reload();
```
ktw1016, line 139: to here

client/src/core/ErrorBoundary.js
```diff
@@ -1,126 +1,230 @@
...
+      } else {
+        const redirect_to_parent_dept_or_home = (level, subj_id) => {
+          const parent_dept_code = _.split(subj_id, "-")[0];
+          const parent_dept = Dept.lookup(parent_dept_code);
+          if ((level === "program" || level === "crso") && parent_dept) {
+            set_redirect_msg(
+              `The ${trivial_text_maker(
+                level
+              )} you tried to visit is either dead or invalid. We redirected you to its parent department.`
+            );
```
ktw1016, line 105: Obviously all of this redirect text can be improved, and also we could possibly redirect to `related` instead of `intro` for this case

client/src/InfoBase/App.js
```diff
@@ -118,15 +118,11 @@ export class App extends React.Component {
   }
 
   state = {
-    showNotification: false,
-    message: "",
+    show_redirected_msg: sessionStorage.getItem("redirected_msg"),
```
Stephen-ONeil, line 121: Ooh, instead of `sessionStorage`, this could finally be a good reason to make use of the app-wide provider. Would deal with that old `No reducer provided for key "app"` error message in the process, haha.

client/src/core/ErrorBoundary.js
```diff
@@ -1,126 +1,230 @@
...
-      testing_for_stale_client: false,
-    });
+      if (_.isNull(error)) {
+        return children;
+      }
+      if (testing_for_stale_client) {
+        this.catch_stale_client_error_case();
+        return null;
+      } else {
+        const redirect_to_parent_dept_or_home = (level, subj_id) => {
```
Stephen-ONeil, line 97: I think all the infographic redirect stuff should live in the infographic route rather than the `ErrorBoundary`. Would prefer to only try and handle generic/app-wide errors at this level.

Stephen-ONeil: If we want sub-app specific redirect rules in other places, maybe it could be a configurable part of the `StandardRouteContainer` or it's own higher order component for wrapping routes in.

client/src/core/ErrorBoundary.js
```diff
@@ -1,126 +1,230 @@
...
+            set_redirect_msg(
+              "The organization you tried to visit is invalid. We redirected you to homepage."
+            );
+            window.location.replace("#home");
+          }
+          window.location.reload();
+        };
+        const current_url = location.pathname;
+        const regex_template =
+          "orgs/(gov|dept|program|tag|crso)/(.*)/infograph/(.*)";
```
Stephen-ONeil, line 119: Using `withRouter`, isn't the parsed URL available in the `props`?

Stephen-ONeil (changes requested):
Couple comments, suggesting a bit of reorganizing

Seems to be working well though, nice!

client/src/InfoBase/App.js
ktw1016: Well, with how it is now, it'll need to keep its variables on reload, and I don't think redux storage will keep that! (Although if we do this at Infographic level, then maybe we can use react router dom and it might have some redirect stuff that doesn't reload? I'd have to try that out)

Stephen-ONeil: Ah, hmm, yeah. Just need to catch the error inside the route rather than let it hit the error boundary and junk everything up 👍

client/src/core/ErrorBoundary.js
ktw1016: Yeah, I'm using that just above `location.pathname`. `regex_template` is the regex to run against the url. Hmm, am I missing something here?

Stephen-ONeil: `match`, https://reactrouter.com/core/api/match

E.g. `const { match: {
    params: { level, subject_id, active_bubble_id, options },
  }} = this.props;`

Stephen-ONeil: Where `level`, `subject_id`, etc. are the parameter names from the URL path definition
```
<Route
  path="/orgs/:level/:subject_id/infograph/:active_bubble_id?/:options?/"
  component={Infographic}
/>
```

client/src/infographic/Infographic.js
```diff
@@ -351,17 +352,76 @@ class InfoGraph_ extends React.Component {
...
+      set_redirect_msg(
+        "The organization you tried to visit is invalid. We redirected you to homepage."
+      );
+      window.location.replace("#home");
+    }
+    window.location.reload();
+  } else if (!bubble_id) {
+    redirect_to_intro(subject);
+  } else if (options) {
+    (() => {
```
GarvinH, line 403: I think we can remove the anonymous function now and execute the try block directly?

ktw1016:
So I've been trying to use app wide redux store instead of session storage, and use `<Redirect/>`. The main problem of it was that it was trying to update redux state, and also return `<Redirect/>` at the same time. So what happened was that it redirects correctly, but gives error that react can't update state, and banner never shows up (from redux store)

I've tried few different ways to set redirect banner in componentdidmount, or do redirect in componentdidupdate, or use redux-thunk (to make a callback to dispatch). But nothing really worked out. If anyone knows how to solve it, please enlighten me!

client/src/core/ErrorBoundary.js
ktw1016: Oh yup you're right. As it moved to Infographic level, it's using it now

Stephen-ONeil:
Using `<Redirect/>` within `Infographic.js`, right? What if what you stored in global state was both the redirect banner content and and the redirect target? Then `App.js` (or some HoC inside it, like the error boundary) could react to the global state change by redirecting and showing the banner (maybe break that in to two steps if necessary). That'd straighten it out, right?

client/src/infographic/Infographic.js
```diff
@@ -350,56 +352,127 @@ class InfoGraph_ extends React.Component {
...
+        params: { level, subject_id, active_bubble_id, options },
+      },
+      set_redirect,
+    } = this.props;
+
+    const is_level_valid = _.chain(Subject).keys().includes(level).value();
+    if (!is_level_valid) {
+      set_redirect(
+        "home",
+        "The level of organization you tried to visit is invalid. We redirected you to homepage."
```
Stephen-ONeil, line 370: Needs translation

client/src/core/ErrorBoundary.js
```diff
@@ -61,10 +63,12 @@ export class ErrorBoundary extends React.Component {
     throw this.state.error;
   }
   render() {
+    const { children } = this.props;
     const { error, testing_for_stale_client } = this.state;
+    const redirected_msg = sessionStorage.getItem("redirected_msg");
```
Stephen-ONeil, line 68: Still using `sessionStorage` rather than app store? Here and two other places. Are those leftovers, or was some mix necessary in the end?

client/src/infographic/Infographic.js
Stephen-ONeil, line 370: Needs translation (here and a few other spots)

client/src/infographic/Infographic.js
```diff
@@ -350,56 +352,127 @@ class InfoGraph_ extends React.Component {
...
+    const SubjectModel = Subject[level];
+    const subject = SubjectModel.lookup(subject_id);
+    const bubble_id = bubble_defs[active_bubble_id] ? active_bubble_id : null;
+
+    if (!subject) {
+      const parent_dept_code = _.split(subject_id, "-")[0];
+      const parent_dept = Dept.lookup(parent_dept_code);
+      if ((level === "program" || level === "crso") && parent_dept) {
+        set_redirect(
+          "orgs/dept/${parent_dept.id}/infograph/intro",
```
Stephen-ONeil, line 382: Oops, "s instead of `s, haha

Stephen-ONeil (changes requested):
Quick review on some small things. Almost good to go I think, I'll do one last quick look over after these are addressed.

client/src/infographic/Infographic.yaml
```diff
@@ -61,6 +61,26 @@ infograph_explorer_link:
   en: Advanced search
   fr: Recherche avancée
 
+invalid_level_redirect_home:
+  transform: [handlebars]
+  en: The "{{level}}" is invalid. We redirected you to homepage.
+  fr: TODO
+invalid_subject_redirect_parent_dept:
+  transform: [handlebars]
+  en: The organization "{{subject_id}}" is invalid. We redirected you to its parent department.
```
Stephen-ONeil, line 70: Your browser was attempting to visit one of our infographics. Unfortunately, we could not recognize "{{subject_id}}" as a valid infographic parameter. While "{{subject_id}}" itself couldn't be found, we were able to match the "{{dept_code}}" portion to a valid organization. We have redirected you to the infographic for that organization.

You were likely following a link intended to take you to an infographic for a core responsibility or program belonging to this organization. You can view and navigate to all of this organization's valid core responsibilities and programs under "Where can I go from here?" below.

If you believe something is incorrect on our end, we invite you to alert us using the Report a Problem form provided at the bottom of the page. Thank you!

client/src/infographic/Infographic.yaml
```diff
@@ -61,6 +61,26 @@ infograph_explorer_link:
...
+  transform: [handlebars]
+  en: The organization "{{subject_id}}" is invalid. We redirected you to homepage.
+  fr: TODO
+invalid_panel:
+  transform: [handlebars]
+  en: The link to panel is invalid.
+  fr: TODO
+invalid_bubble_redirect_intro:
+  transform: [handlebars]
+  en: The "{{bubble_id}}" is invalid. We redirected you to intro page.
```
Stephen-ONeil, line 82: This error case can probably fail quietly. I think it would be fair to just leave them in the infographic with no selected bubble if the bubble portion is malformed, wouldn't even need to redirect or anything.

client/src/infographic/Infographic.yaml
```diff
@@ -61,6 +61,26 @@ infograph_explorer_link:
...
+  transform: [handlebars]
+  en: The organization "{{subject_id}}" is invalid. We redirected you to its parent department.
+  fr: TODO
+invalid_subject_redirect_home:
+  transform: [handlebars]
+  en: The organization "{{subject_id}}" is invalid. We redirected you to homepage.
+  fr: TODO
+invalid_panel:
+  transform: [handlebars]
+  en: The link to panel is invalid.
```
Stephen-ONeil, line 78: This doesn't need to be a redirect thing, right? The `JSURL` parsing in `componentDidUpdate` should handle errors itself (and probably fail quietly).

Stephen-ONeil (changes requested):
Misc comments, and tried my hand at writing text for the redirect banners. Might be a bit long in most cases though.

client/src/core/ErrorBoundary.js
```diff
@@ -78,24 +83,47 @@ export class ErrorBoundary extends React.Component {
             alignItems: "center",
           }}
         >
+          {redirected_msg ? (
```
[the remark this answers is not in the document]
ktw1016: This is kind of an artifact from when it was redirecting at error boundary level, but it's still applicable (sort of). Because `SafeJSURL.parse(options);` itself throws an error and hits the error boundary for a second before the reload happens.

Just left it in there to see the reaction since it was already written. I can remove it!

client/src/infographic/Infographic.yaml
ktw1016: okay sure

client/src/infographic/Infographic.yaml
ktw1016: Yes, if actual panel string is invalid. But if the whole structure is wrong, like `#orgs/gov/gov/infograph/financial/.-ERROR.-(panel_key.-.-'welcome_mat)`. It'll throw an error

client/src/infographic/Infographic.js
```diff
@@ -358,10 +363,63 @@ const Infographic = ({
     params: { level, subject_id, active_bubble_id, options },
   },
 }) => {
+  const set_redirect_msg = (msg_key, args) => {
```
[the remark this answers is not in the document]
ktw1016: Can you elaborate on what you mean by its own HoC? This doesn't render anything, so they're not really components. I did put replacing and reloading to here to centralize it now tho.

I could put all of this redirect code in a separate function, although checking for level have to happen before getting the subject, so then I'd need two different functions, which could be a bit messy in its own. Although that might look clearer in the bigger picture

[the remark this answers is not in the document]
Stephen-ONeil: Ahh, haha, know what? I was confused by the GitHub diff/not reading carefully.

image


Didn't notice the jump in line number, thought this was all inside the `AnalyticsSynchronizer` component for some reason, which would have been a weird place for it. You can ignore this one!

client/src/infographic/Infographic.yaml
Stephen-ONeil: I meant the actual errors thrown by `JSURL` on unparsable URLs, yeah. If those options can't be parsed, we don't actually know if they were a `panel_key` or something else (we use them for like linking to indicator modals too, and generally the idea was that panels could use them for arbitrary state if that ever became necessary). Figure those errors should just be suppressed (and maybe logged to analytics), but don't really need to be communicated up to the user.

client/src/core/ErrorBoundary.js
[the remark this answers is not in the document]
Stephen-ONeil: Do you mean an error in the `ErrorBoundary` itself, or in the redirect code? The `ErrorBoundary` itself is tested by the route load tests. We could add specific testcafe tests for the redirect code too, would be pretty easy.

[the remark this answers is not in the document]
ktw1016: Well, since I just removed SafeJSURL parse, I removed this as well. I guess when it gets stuck reloading, we'll put this back in haha

Stephen-ONeil (changes requested):
Still think the redirect code should be moved out of the error boundary. Other than that, just finishing the text (team feedback and translation)!

Stephen-ONeil:
Ah, haha, you cleaned up the `ErrorBoundary` while I was writing that. Ok, just text left then!

Stephen-ONeil (changes requested):

client/src/infographic/Infographic.yaml
```diff
@@ -61,6 +61,34 @@ infograph_explorer_link:
...
+invalid_redirect_home:
+  transform: [handlebars]
+  en: |
+    Your browser was attempting to visit one of our infographics. Unfortunately, we could not recognize "{{param}}" as a valid infographic parameter. You have been redirected to the GC InfoBase home page.
+
+    If you know the subject of the infographic you intended to view (e.g. the organization or program name), please try finding it using our search bar below.
+
+    If you believe something is incorrect on our end, we invite you to alert us using the Report a Problem form provided at the bottom of the page. Thank you!
+  fr: |
+    Votre navigateur a tenté de visiter l'une de nos infographies. Malheureusement, nous n'avons pas pu reconnaître "{{param}}" en tant que paramètre infographique valide. Vous avez été redirigé vers la page d'accueil l'InfoBase du GC.
```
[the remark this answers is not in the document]
ktw1016: This is from Andre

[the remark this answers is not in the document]
Stephen-ONeil: Ahh, ok, sorry!

client/src/InfoBase/App.js
```diff
@@ -169,10 +169,27 @@ export class App extends React.Component {
...
+                  text_maker("common_redirect_msg", {
+                    url: get_session_storage_w_expiry("pre_redirected_url"),
+                  }),
+                  get_session_storage_w_expiry("redirected_msg"),
+                ]}
+                hideNotification={() => {
+                  sessionStorage.removeItem("pre_redirected_url");
+                  sessionStorage.removeItem("redirected_msg");
+                  this.setState({ show_redirected_msg: false });
+                }}
```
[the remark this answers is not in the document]
ktw1016: Can you elaborate on what you mean by separate components, do you mean this should use a different alert component from outage?

This is not exactly "redirect component", it's just used to alert user on the error. The actual "redirect" is happening in Infographics level, so I think it should do the logging there

client/src/infographic/Infographic.js
```diff
@@ -148,7 +153,19 @@ class InfoGraph_ extends React.Component {
...
         reset_scroll();
       }
-      const options = SafeJSURL.parse(this.props.options);
+      const options = (() => {
+        try {
+          return SafeJSURL.parse(this.props.options);
+        } catch {
+          log_standard_event({
+            SUBAPP: window.location.hash.replace("#", ""),
+            MISC1: "ERROR_IN_JSURL_PARSE",
```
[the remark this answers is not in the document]
ktw1016: Agreed to put them as warnings for weekly reports on that, should be separate from this PR though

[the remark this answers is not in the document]
Stephen-ONeil: Agreed, that's why I opened the issue. Was just linking it to this because it's where I had the thought 👍

client/src/InfoBase/App.js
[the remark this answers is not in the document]
Stephen-ONeil: Oops, sorry for not getting back on this.

I meant more as a code organization/clean up thing. We're ending up with a lot of functionality directly embedded in the `App` component. I was suggesting these headers should be encapsulated as their own components and moved to their own modules in `src/core/` (for now, some day we'll reorganize the directory structure). Like `InsertRuntimeFooterLinks` etc.

The module for the `RedirectHeader` (or whatever) component could also export a function like `redirect_with_message` or something, with equivalent behaviour to the `set_redirect_msg` function currently in `Infographic.js`. Right now implementation details for this feature are split between `App.js` and `Infographic.js`, which a) makes it harder to maintain, and b) makes it harder\* to reuse elsewhere if needed.

\* _could_ use the redirect header from anywhere right now, I guess, but you'd need to understand (or copy) the implementation details from `Infographic.js`, which would just make maintenance and even bigger nightmare.

Stephen-ONeil (changes requested):
https://github.com/TBS-EACPD/infobase/pull/826#discussion_r551588660

Stephen-ONeil (approved):
👍

PR #827, opened 2020-11-18
Treemap page flex layout

AlexCLeduc:
Got the treemap working with bootstrap4. The treemap page has always had problems with flex, so we probably used the old classes (`col-*-*`,`row`, not  `fcol` and `frow`) to in order to use float-based columns instead. Once we upgraded to bootstrap 4, the float styles were replaced with flex, and the treemap was once again broken. It looks like we just needed to add a few `width:100%` to get the whole thing working again

Tested it with chrome and FF. Assuming we're going to test the rest of the BS4 upgrade against IE11, we should have a quick glance at this page too.

### DIFF
--- a/client/src/TreeMap/TreeMap.js
+++ b/client/src/TreeMap/TreeMap.js
@@ -521,9 +521,9 @@ export default class TreeMapper extends React.Component {
               </button>
               <TreeMapInstructions />
               {actual_to_planned_gap_year && <GapYearWarning />}
-              <div className="row">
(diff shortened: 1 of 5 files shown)


GarvinH (approved):
Nice!

Yea it seems to be working again. In fact I think its working even better than before! Overall seems to be less text being cut off compared to the one currently in master.

I tested this on IE on the surface and it seems to be working fine there too.

PR #831, opened 2020-11-23
Re-run of DRR19 data 

ktw1016:


ktw1016:
Just a note, we should disregard services csv's because it is being handled in #672

PR #833, opened 2020-11-24
Add simplified feedback in survey campaign

### LINKED ISSUE #816: Ask numeric questions directly in survey campaign popup
Next time we enable the survey campaign banner, Darren suggested embedding a mini survey directly in it. Probably just asking the two or three numeric questions in the popup, maybe with an alternate button for "take full survey".

ktw1016:
closes #816 
Also returns empty string for GA client_id, in case GA is not initialized (for ex, because of adblock)

### DIFF
--- a/client/src/core/SurveyPopup.js
+++ b/client/src/core/SurveyPopup.js
@@ -1,10 +1,10 @@
-
 import { Fragment } from "react";
 import { withRouter } from "react-router";
 
 import {
   FixedPopover,
   create_text_maker_component,
+  EmailFrontend,
 } from "../components/index.js";
 import { IconFeedback } from "../icons/icons.js";
 
@@ -15,7 +15,8 @@ import text from "./SurveyPopup.yaml";
 const { TM, text_maker } = create_text_maker_component(text);
 
 const page_visit_increment = 1;
-const survey_campaign_end_date = new Date(2020, 3, 31).getTime();
+//TODO Need to set the correct time
+const survey_campaign_end_date = new Date(2022, 3, 31).getTime();
 
 const get_path_root = (path) =>
   _.chain(path).replace(/^\//, "").split("/").first().value();
@@ -133,10 +134,12 @@ export const SurveyPopup = withRouter(
       const { active, page_visited, show_popup } = this.state;
       const { toggleSurvey } = this.props;
 
-      const should_show =
+      let should_show =
         !is_survey_campaign_over() &&
         (page_visited >= 3 || show_popup) &&
         active;
+      // TODO, should be removed before merging
+      should_show = true;
 
       if (should_show) {
         clearTimeout(this.timeout);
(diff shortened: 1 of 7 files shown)


client/src/core/SurveyPopup.js
```diff
@@ -161,18 +176,29 @@ export const SurveyPopup = withRouter(
...
+              }}
+            >
+              <TM k="survey_popup_body" />
+              {full_survey_button}
+              <TM
+                k="take_simplified_survey_text"
+                style={{ margin: "8px 0px" }}
+              />
+              <EmailFrontend
+                template_name="feedback_simplified"
```
GarvinH, line 194: Reminder to rebuild email_backend before merging! (Also reminder that test only works local for now until then)

email_backend/templates/feedback_simplified.json
```diff
@@ -0,0 +1,69 @@
...
+      }
+    },
+    "style": "horizontal"
+  },
+  "useful_info": {
+    "required": true,
+    "form_type": "radio",
+    "form_label": {
+      "en": "To what extent did you find the spending information on GC InfoBase useful? (1=not at all useful and 5=very useful)",
+      "fr": "Dans quelle mesure avez-vous trouvé les renseignements financiers utiles dans l’InfoBase du GC? (1=inutile et 5=très utile)"
```
Stephen-ONeil, line 42: I wonder if we should drop "spending" here and just ask about usefulness of information in general? Might be a bit complicated because the corresponding indicator might specifically refer to spending information though, so we should double check how strictly we have to match that before making changes.

client/src/core/SurveyPopup.yaml
```diff
@@ -6,17 +6,9 @@ survey_popup_body:
...
-
-    Choosing “Yes, after my visit” will open our survey in a new window. You can return to this window to complete the survey after you have finished browsing GC InfoBase; it will not time out.
-
-    We value your feedback. Thank you!
   fr: |
     À la fin de votre visite aujourd'hui, prenez le temps de nous parler de votre expérience avec InfoBase du GC.
 
-    En sélectionnant "Oui, après ma visite", vous accéderez à une nouvelle fenêtre. Quand vous aurez fini de parcourir InfoBase du GC, vous pourrez revenir à cette fenêtre pour répondre au questionnaire ; il vous restera du temps.
-
-    Vos commentaires nous sont précieux. Merci!
```
Stephen-ONeil, line 18: Please this text in for now, I'll slack you about why.

Stephen-ONeil (approved):
Couple notes (and those TODOs to remove), but generally approved 👍

email_backend/templates/feedback_simplified.json
ktw1016: General information usefulness would be better IMO. Can you elaborate what you mean on this indicator specifically referring to spending info? Not too familiar with email_backend code but doing a quick `useful_info` search didn't return much "useful info"😏

Stephen-ONeil: Methodology on this indicator refers to us asking a question specifically about spending information 
image

Need to double check with someone whether that restricts us from changing the wording of that question to be more general.

ktw1016: ohhhh our own result indicator, okay got it. Who should we ask for that?

client/src/core/analytics.js
```diff
@@ -21,13 +21,7 @@ const dimensions = {
 };
 
 let client_id;
-const get_client_id = () => {
-  if (_.isUndefined(client_id)) {
-    throw "Error: Google Analytics has not been initialized";
-  } else {
-    return client_id;
-  }
-};
```
Stephen-ONeil, line 30: Oh, maybe, leave this as a warning `console.warn(...)`, just to help with debugging if something in analytics breaks in the future.

PR #837, opened 2020-11-27
Unpin key concepts if quick link to panel

GarvinH:
I had to move the localstorage code from `componentdidupdate` back into the `pinpressed` to prevent the key concepts from unpinning when changing to another page or staying pinned when changing from a previous page.

The only time it would pin now is if the user is already on the page where the panel is.
For example going to a panel in results while in results will keep the key concepts pinned (assuming that the pin was already active)

### DIFF
--- a/client/src/hoc/PinnedContent.js
+++ b/client/src/hoc/PinnedContent.js
@@ -1,4 +1,5 @@
 import classNames from "classnames";
+import { withRouter } from "react-router";
 import { InView } from "react-intersection-observer";
 import "intersection-observer";
 import ReactResizeDetector from "react-resize-detector";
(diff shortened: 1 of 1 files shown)


client/src/hoc/PinnedContent.js
```diff
@@ -13,123 +14,130 @@ import text from "./PinnedContent.yaml";
...
-      try {
-        user_enabled_pinning = JSON.parse(
-          localStorage.getItem(local_storage_name)
-        );
-        user_enabled_pinning = _.isBoolean(user_enabled_pinning)
-          ? user_enabled_pinning
-          : true;
-      } catch {
-        user_enabled_pinning = true;
+      const local_storage_name = props.local_storage_name;
```
ktw1016, line 28: Can squeeze this in destructuring props above

client/src/hoc/PinnedContent.js
```diff
@@ -13,123 +14,130 @@ import text from "./PinnedContent.yaml";
...
-      user_enabled_pinning: !user_enabled_pinning,
-    });
-  };
+    pin_pressed = () => {
+      const { local_storage_name } = this.props;
+      const { user_enabled_pinning } = this.state;
+      this.setState({
+        user_enabled_pinning: !user_enabled_pinning,
+      });
+      localStorage.setItem(local_storage_name, !user_enabled_pinning);
```
ktw1016, line 59: Can we set the local storage before setState? I don't think it should actually be a problem but just in case it re-renders before local storage

ktw1016 (commented):
Just wondering, why do we want it unpinned if link is to a panel?

GarvinH:
>Important though! One more feature that occured to me. Please look in to this and either open an issue or a follow up PR to deal with it.

>So, if you're linked directly to panel, and you've not previously pinned the keep-in-mind, it will be covering the panel's title and possibly even some important panel content. If you've not been to an infographic before (directly linked to a panel from off-site, or following a home page quick link, etc), you'd not know that there's important missing content under the pinned banner. You may not even necessarily realize that banner's pinned, since you didn't see it pin itself while scrolling past it.

>I think the PinnedContent component needs a prop to override the localStorage to set the initial pinning state. More troublesome will be making the key concept panel recognize when the route was a direct panel link and respond accordingly, I'll leave that up to you.

ktw1016 (approved):

PR #844, opened 2020-12-03
Remove globals

### LINKED ISSUE #840: Eliminate globals
`inject_app_globals.side-effects.js` should be gutted. Not convinced anything should be on `window` at all. Global constants/env vars should be moved to their own module. React, D3, and Lodash could be imported as needed. D3, if not even Lodash (harder), could move to importing only what's required, instead of importing the whole library. `feature_detection` and the colour globals should never have been globals in the first place (just dev laziness there).

These are essentially all dependencies that we're not letting webpack help us with, so if we get out of it's way we'll benefit more from current/future optimizations too. Maybe somewhat minor in the end, but it'll help the entry point bundle and any optimization to the initial load time is worth it.

Only exception is `_DEV_HELPERS` which a) could use a rename and b) maybe shouldn't exist in prod (I mostly use it on the master dev link anyway). It's effectively lazy-loade, since things like subject only assign to it as they initialize, so it's not part of the optimization argument.

ktw1016:
closes #840


ktw1016:
Couldn't remove Handlebars from globals because circular dependency

I essentially tried exporting out Handlebars from handlebars/helpers.side-effects code. Then, importing that where we use Handlebars (format.js, text.js). Then, ran into so many circular dependencies problems that I couldn't fix

Stephen-ONeil:
Handlebars helpers are still a weird (implicit) circular dependency mess. Would be nice to come up with a better approach for that eventually, but out of scope here. I'm done making changes and have to get back to other priorities, so throwing this back to you @ktw1016. Please review (in broad strokes, too much minor changes/linting to go line-by-line) and, more importantly, test the dev link. Thanks!

client/src/InfoBase/bootstrapper.js
```diff
@@ -4,8 +4,7 @@ import "whatwg-fetch";
 
 import "../common_css/common_css_index.side-effects.js";
 
-import "./inject_app_globals.side-effects.js";
-import "../handlebars/initialize_handlebars.side-effects.js";
+import "../handlebars/register_helpers.side-effects.js";
```
ktw1016, line 7: Ah okay, I was essentially trying to remove side effects altogether, to put all registerHelpers inside `text` or `format` but this still removes globals !

ktw1016:
Looks good to me! I'll merge it in

client/src/InfoBase/bootstrapper.js
Stephen-ONeil: I'd like to still untangle it more. There's too much of a mess with there being, for example, helpers that need access to glossary while glossary itself needs access to text maker. We'd have to come up with our own way to manage the helper dependencies of yaml files or something though, so way too out of scope for here.

PR #847, opened 2020-12-07
Handle people panels text better

ktw1016:
Might be an overkill but this is for text of `employee_gender` and `employee_fol`.

Check if data has eng/fr or male/female, and get those data points only. If they don't have that, get top and bottom groups to calculate for text


Stephen-ONeil (approved):
👍

PR #848, opened 2020-12-07
Bubble menu cleanup

Stephen-ONeil:
Some ancient messy code in here, took a quick diversion to clean it up somewhat.

One more thing to do but I'll probably just open an issue for it after this merges instead of doing it now. The "bubbles" should use icon components instead of storing SVGs in yaml. Some of these are duplicates of existing icon components, so we're risking desyncs by duplicating them here.


client/src/infographic/bubble_definitions.js
```diff
@@ -1,85 +1,57 @@
...
-          return text.about_tag_title[lang];
-        case "crso":
-          return text.about_cr_title[lang];
-        default:
-          return "Introduction";
-      }
-    },
-    description: _.constant(""),
+    title: (subject) => text_maker(`about_${subject.level}_title`),
+    description: "Introduction",
```
ktw1016, line 17: Should be in French as well?

ktw1016 (approved):
👍

client/src/infographic/bubble_definitions.js
Stephen-ONeil: The word works in both languages 👍

Issue #838, opened 2020-11-27
Post-release covid estimates panel TODOs


Stephen-ONeil:
Rushed out to hit the deadline, lots of stuff between the API and client to polish off as things settle. Some changes may or may not be better to leave off pending changes to the form of the data.

TODOs:
  - [ ] actually use years on the client side, to future proof things (could wait till after initial release)
  - [x] server
    - [x] \*when querying initiatives by org, filter estimates included in response (see \* for related) #851
      - [x] think about this more, make sure it's not going to mess with/be messed with by Apollo's normalized caching
  - [ ] client subjects
    - [x] \*be more efficient when querying by dept (once server scopes estimates when querying by org, need to be able to merge in additional estimates data to initiative as additional queries made) #851
    - [x] ~populate `CovidEstimate` with rolled up data when fetching `CovidInitiative` for a given org~ #851
    - [ ] respect fiscal years
  - [ ] panels
    - [ ] I think the tables would be better as `(org or initiative) | ...est doc columns | cumulative authorities`... but dropping vote stat from them would mean something like an RPB table is necessary to still have that detail
    - [ ] gov tables are very long, could they be "top 10 depts", etc?
    - [ ] text is hard coded to current estimates docs, needs to be generalized. Maybe move list of tabled docs up from the overview so the above-tab text can just say "the documents listed below"
      - [ ] that might turn the "overview" tab in to a "vote stat" tab, which would leave more room for removing vote stat from the other tabs...
    - [ ] respect fiscal years
  - [ ] Going to either need to put the data in the report builder or give it a sub-app or something. Someone for sure will be asking to see all the data in full granularity all in one place, and linking to open data might not be satisfactory

Stephen-ONeil:
Expecting these to all be superseded by changing plans (if not, they'll return in futures issues, I'm tracking them elsewhere for now).

PR #851, opened 2020-12-10
Covid estimates by measure

Stephen-ONeil:
TODO:
  - [x] **fake data** for estimates by measure, drop initiatives from data
  - [x] drop initiatives from GraphQL, initial rework to be by-measure
  - [x] item from #838 on filtering when querying granular estimates by org still applies
  - [x] rework client subjects and loading
    - [x] TODO how to approach the measure subject/attaching data to it...
      - aha, `exstensibleStoreMixin` is finally good for something other than tags!
    - [x] ~set org `has_data` as other data loads~
      - scratch that, a little bit of efficiency isn't worth having multiple sources of truth for those `has_data` flags. The GraphQL `has_covid_data` query should be the only thing to set this value
    - [x] ~maybe start trying to address year items from #838 (in the server/subjects/loading only, don't bother in the panel yet)~
      - ... or not. The server and subject understand years, and we probably won't care about scoping the queries by year until this data really grows. In the end, only the panel actually needs to be year-aware in the medium term, and that's still out of scope for the short term!
    - [x] hmm, summary stats shouldn't be crammed in to any models. Gonna be a tangent, but these should be queries that panels make on their own (which may mean expanding the panel API to let panels declare queries to be loaded alongside the old style `ensure_loaded` call by the infographic itself)
      - stop gap version done. Panels can declare arbitrary initial queries, which are added to the ensure loaded promises on the infographic-wide spinner. The panel doesn't receive the response directly though, has to re-query for it from the rendered component (hits cache immediately though, so no secondary panel-level spinner for the user). Bothersome boilerplate in the panel's components though, and won't have the queried data available for aborting the panel from the calculate step. Lots of panel registry/infographic plumbing to rewrite to improve this any further
  - [x] ~polish the `CovidMeasure` subject a bit more?~ good enough for this PR, and will make more sense to polish it after/as additional layers are added by future PRs 
  - [x] refactor panel, switch from "estimates initiatives" to "COVID-19 measures" in text
  - [x] Glenn's text change suggestions:
    - [x] estimates panel, rename to something like "Spending Authorities by COVID-19 Measure as Presented in the Estimates Process"? The idea here is that "spending authorities" are more plain-language than "estimates", but that we then needed to be extra clear that these are still Estimates Process specific authorities. Pretty wordy though, try and think of a better middle ground here
       - Estimates wants "Spending Authorities for COVID-19 Measures in the Estimates"?
     - ~remove "actual" from "actual expenditures"?~ Going to not do this, at least for now, sentence reads better with actual in there I think
  - [x] ~might be possible for a dept to have measure estimates but not have standard estimates... make sure panel handles that~
    - actually, drop the comparison all together. It's apples to oranges until/unless the covid estimates break out bud/non-bud for us 
  - [x] update pipeline on new expected shape of the covid data files
  - [ ] get **real data**
    - [x] ~figure out if the final data includes any more non-budgetary values than the previous initiative data did. The initiative data included a single negligible partial non-bud amount so was still comparable to other estimates presented on the InfoBase, but might need to drop those comparisons if alignment changes that~ better safe than sorry, dropping that comparison
  - [ ] get estimates by initiative datasets on Open Data. Link to it from a footnote or something, to direct users to the old categorization after we've switched to by-measure
  - [ ] estimates team review, additional footnotes
  - [ ] translation


ktw1016 (changes requested):
I feel `covid_estimates.js` is a bit overcrowded with all the tabs and panel code in the same file. It'd make it more readable and organized if you decentralize the tabs code and panel code.

And looking over data loading mechanism here, I find it a bit hard to follow. So for example `gov_covid_estimates_summary_query`, it first loads in `initial_query` (temporarily to load behind infograph wide spinner). Then again in `CovidEstimatesPanel`, And then again in `SummaryTab` with `load_data`. Can you explain what's going on, and the need for loading it 3 times if I'm understanding correctly?

Stephen-ONeil:
I agree on `covid_estimates.js` being a bit bloated. Will hopefully split in to multiple panels down the road, but moving individual tabs to their own modules could help clean things up to 👍 

As for loading, it's partially an imperfect stop-gap and partially a new paradigm. 

Ideally, panels will be able to declare the queries they want data from and have that data available within their `calculate`s (just for initial render, data required inside non-default tabs can be put off). That's a lot of additional plumbing to set up though, so for now I just added the `initial_queries` option to panels and make those queries in `ensure_loaded` (but don't do anything with the data at that point). Apollo caches by query, so when the panel re-queries during render it just immediately hits that cache. Moves the panel's GraphQL data dependency up to load alongside everything else and get's rid of the annoying panel-level spinner (only headache's that the panel still needs to manage a loading state since the query, despite immediately hitting cache, still needs to be made async).

The "new paradigm" is just the standard Apollo use case. Apollo's cache is meant to, effectively, be your app's datastore (as far as data from GraphQL is concerned). Any time you need data you make a query, you make no assumptions about wether it will hit the cache or not\*. Only one step as compared to a separate `ensure_loaded` and `subject.lookup`. Safer and simpler (aside from all data access being async). We probably won't move to making raw queries everywhere, but I think it would be good to remove the `StaticStore` aspect of subject classes and just turn them in to utility wrappers around pre-baked queries\*\*.

Short term, big changes to subjects aside, this specific summary data doesn't fit in to the subject stores well. Great candidate for the Apollo-cache-as-data-store approach.

\* ... unless you're trying to pre-populate the cache as an optimization
\*\* not 100% set on this approach yet. We do some complicated optimizations right now, where loading certain data has the side effect of populating other, derivable, stores as well. Might be worth dropping that complexity, might be worth figuring out a way to continue doing it. 🤷

ktw1016:
I'm very for removing static stores, and having queries baked straight into individual panels. A lot more intuitive and better to decentralize like that, and having a utility wrapper around the query to use that cross the panels sound good to me!

I guess all of that's a big change and this should be good to go provided the other changes will happen eventually !

Stephen-ONeil:
Temporary change of plans, shelving.

PR #864, opened 2021-01-07
People panel text calculation cleanup

Stephen-ONeil:
Post-hotfix cleanup, caught and squashed some other minor bugs in the process (first active year index was always calculated as 0, last year total population panel didn't handle departments without abbreviations, etc.).


ktw1016 (approved):
Looks good to me!

PR #866, opened 2021-01-13
Filter infograph panels alternate

Stephen-ONeil:
Refactor of #843, trying to better separate the concerns of filtering and rendering panels in the Infographic.

... Infographic sure is a mess. Not just `Infographic.js`, but the whole panel ecosystem at this point. Cleaning that up will need to be ongoing (a fair bit will have to be rewritten either way in the GraphQL switch-over).


client/src/infographic/Infographic.js
```diff
@@ -274,49 +274,12 @@ class InfoGraph_ extends React.Component {
...
-          .filter()
-          .value();
-        const valid_subject_panels_by_bubble_id = panel_keys
-          ? {
-              ...panel_keys_by_bubble_id,
-              [active_bubble_id]: _.map(valid_panel_objs, "key"),
-            }
-          : panel_keys_by_bubble_id;
-
+  load({ subject, level, active_bubble_id }) {
```
Stephen-ONeil, line 277: I should have commented on this. Earlier I was saying it should be two different loading steps, but then I realized that half of the bubble menu loading step was not necessary. I think ultimately this giant `Infographic` component should be broken up more at which point there will be two distinct loading steps, but I collapsed them together for now because I was in a bit of a rush.

client/src/infographic/PanelFilterControl.js
```diff
@@ -0,0 +1,164 @@
...
+    });
+  };
+  onSelectNone = () => {
+    const { table_tags } = this.state;
+
+    this.setState({
+      table_tags: _.mapValues(table_tags, _.constant(false)),
+    });
+  };
+  panel_filter_factory = () => {
```
ktw1016, line 138: I find this setup a bit confusing. It feels more natural to me to bring this into `Infographic` level, and just pass `number_of_active_panels` in as  props. Upto you on this change tho

client/src/infographic/InfographicPanelRenderer.js
```diff
@@ -1,141 +0,0 @@
...
-          active_panel_filter_keys,
-          panel_obj.depends_on
-        );
-        // Show the panel regardless of filter status if it's neither financial or people. Also if there is no panel filter applied.
-        const is_panel_valid =
-          show_all_panels || is_panel_filter_empty || filtered_keys.length > 0;
-
-        // I think it makes sense to not include "static" panels (non table data panels) in showing number of active panels
-        // But, always display them
-        if (!panel_obj.static && is_panel_valid) {
```
ktw1016, line 53: Need to check these `static`

ktw1016 (changes requested):
Things are missing:

- panel counts are off due to counting for `static`. We need to check those when filtering
- I purposely made panel filter flag inverse, similar to how TablePicker for RPB works. (Initially setting all filters false, but showing all panels if all filters are false)

client/src/infographic/InfographicPanelRenderer.js
```diff
@@ -1,141 +0,0 @@
...
-      active_bubble_id,
-      total_number_of_panels,
-      panel_filter_by_table,
-      set_panel_filter_by_table,
-    } = this.props;
-
-    const show_all_panels = !(
-      active_bubble_id === "financial" || active_bubble_id === "people"
-    );
-    const is_panel_filter_empty = _.filter(panel_filter_by_table).length === 0;
```
ktw1016, line 33: Gotta check this one

client/src/infographic/PanelFilterControl.js
Stephen-ONeil: This approach was the most decoupled one I could think of. `Infographic.js` knows as little as it can about the implementation of `PanelFilterControl`.  Leaned hard in this direction since `Infographic.js` is already, IMO, doing too many things, and because it's a minority of infographics that actually get filters for now.

For `number_of_active_panels` specifically, if it was up to `Infographic.js` to calculate that then `Infographic.js` would need to know about the special case with "static" panels. There's no other reason for `Infographic.js` to be concerned if a panel is "static" or not, it's only a meaningful distinction when it comes to the implementation of the filter. Just good separation of concerns to move all that logic down in to `PanelFilterControl`.

PR #872, opened 2021-01-26
Filter infograph panels alternate

ktw1016:


ktw1016:
Actually, found an issue with this architecture. Due to how `panel_filter_factory` is passed. When Infographic changes, `panel_filter` doesn't update because it doesn't need to but it should. E.g.:
1. Apply filter on finance bubble
2. Switch to People bubble
3. No panels are shown for the reason above I think

Can't think of a easy quick solution without changing the flow direction of `panel_filter_factory`. Thoughts @Stephen-ONeil ?

Stephen-ONeil:
Hmm, don't know when I broke that. Easy fix. I'll push something but I'm going to re-rebase this branch first because it's got copies of some of the commits from the base branch in it right now.

Stephen-ONeil:
Sorry, to clarify, `PanelFilterControl` knows to reset the filter to `_.identity` when the set of tables changes (i.e. infographic change)... except the `PanelFilterControl` component doesn't live through that change, it unmounts when the `Infographic` goes in to a loading state, however brief. Haha, I remember adding the logic to `PanelFilterControl` to respond to the set of tables changing out of caution for `Infographic`'s render potentially changing and nolonger unmounting it on loads, but then I forgot that the checks weren't ever reached and started relying on them, oops!

I'll add comments so that doesn't happen again.

Anyway, `PanelFilterControl` could just use the callback to set a default filter (`_identity`) on mount. Alternatively, could give `Infographic` a "get_default_state_from_props" method and reuse that between the constructor and `getDerivedStateFromProps`. Right now they use almost the same logic, but the derived state misses setting the filter to `_.identity` like the constructor does. That would fix it too.

PR #875, opened 2021-01-27
Break up d3 bundle

### LINKED ISSUE #869: Break up d3-bundle.js
Really should be only importing from D3's sub-package where and as needed, rather then bundling them all in to a single object and always importing that. Should be a nice bundle size reduction in lots of places. Something to do after #844 merges.

Stephen-ONeil:
Closes #869


ktw1016 (approved):
I checked over first 20ish files and everything seems pretty trivial

PR #877, opened 2021-02-01
whole bunch of removing the and putting GC in front of infobase

ktw1016:
As well as for French


Stephen-ONeil (approved):
👍  for the English at least

PR #879, opened 2021-02-02
Filter by footnotes

### LINKED ISSUE #871: Add footnotes to infographic filtering
Add filter by footnotes to infographic filter

ktw1016:
closes #871

### DIFF
--- a/client/src/components/Panel.js
+++ b/client/src/components/Panel.js
@@ -2,7 +2,6 @@ import classNames from "classnames";
 import _ from "lodash";
 import React from "react";
 
-
 import { Details } from "./Details.js";
 import { FootnoteList } from "./FootnoteList.js";
 import { GlossaryItem } from "./glossary_components.js";
--- a/client/src/infographic/PanelFilterControl.js
+++ b/client/src/infographic/PanelFilterControl.js
@@ -112,7 +112,7 @@ export default class PanelFilterControl extends React.Component {
         persist_content={true}
         content={
           <LabeledBox
-            label={text_maker("filter_panels_description")}
+            label={text_maker("filter_by_datasets")}
             children={
               <div>
                 <TagCloud tags={tags} onSelectTag={this.onSelect} />
--- a/client/src/infographic/PanelFilterControl.js
+++ b/client/src/infographic/PanelFilterControl.js
@@ -15,9 +15,11 @@ import { SelectAllControl } from "src/charts/legends/index.js";
 
 import { PanelRegistry } from "src/panels/PanelRegistry.js";
 
+import footnote_text from "../models/footnotes/footnote_topics.yaml";
+
 import text from "./PanelFilterControl.yaml";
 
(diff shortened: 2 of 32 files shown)


Stephen-ONeil (commented):
Hmm, I'm wondering if there's an easier way to present this to users. The idea of "filtering by footnotes" will be pretty weird to them, I think. 

What if it was just on list of options vaguely called "Filter by topic"? Maybe make it unique by label text? A bit fiddly because there'd be one button that toggles both the table and footnote state behind the scenes, but I think that would be simplest for users. 

Thoughts?

ktw1016:
Hmm, yeah I see what you mean. Most of footnotes are actually just basically datasets. But we can't really make them unique by label text, there would be like: "Authorities and Expenditures" but also "Authorities" and "Expenditures".

How about actually removing filter by footnotes altogether? I noticed that in most cases, list of footnotes are subset of datasets with the exception of variances.

Stephen-ONeil:
Hmm, I think the real, more long term, fix is to expand the set of topic tags and their usage, probably invert their relationship with footnotes too. Rather than the footnotes being the source of these tags we'd have a larger list of tags managed elsewhere and footnotes would just make use of them. We'd have to manually manage the list and manually apply the appropriate tags to panels, but it wouldn't be too much more work than applying footnote tags and would cover footnotes in the process. Then we could have a more useful set of tags to filter things by in general (probably both panels and whatever datasets in the report builder look like post-GraphQL?).

That would better solve the root problem I was hoping footnote filtering would solve, which is that filtering panels by dataset is going to stop making sense as they gradually move to GraphQL.

Does that seem worth doing to you? Would need to involve the pipeline a fair bit since we'd want it to be the source of topics and also enforce that the footnotes only use valid topic tags. Fair bit of manual data management to make the shift happen/come up with all the topic tags. 

If you've got a different idea of how to make this feature useful without tables, let me know!

ktw1016:
Yeah that makes sense. For the footnote, I think the idea is kind of already there, in a sense that there are list of topics for each footnotes, then it drills down to each individual departments. I guess the difference would be that we would place footnotes for a panel by topics, then each individual departments.  I think the process of that should be:
1. Come up with list of topics, make a relationship with panels in a csv
2. Change how footnotes are loaded (render a panel -> lookup list of topics for this panel -> load footnotes with this topic and its department specific footnotes)
3. Replace existing filters with filter by these topics

Let me know if I understood that correctly, if so, I'll close this PR and open new issues

Stephen-ONeil:
I think topics should exist entirely on their own in the data, they shouldn't be intrinsically linked to panels etc. They're just a common language for other systems to speak. I.e. panels, footnotes, report builder pages (meaning Tables for now, but they'll be their own thing eventually), etc. can all individually list the topics that apply to them. The central topic list is just a source of truth for what keys are valid and what their plain language names are.

Not centralizing the relationships between panels and topics, etc., will make it easier for additional systems (search maybe?) to also make use of the topics "ecosystem".

PR #884, opened 2021-02-08
Svg cleanup

### LINKED ISSUE #874: Standardize on src/icons for all svgs
Home page linkcards still use images with static urls to files in `src/svg`. Bubble menu icons are in yaml. Etc. 

Anything in `src/svg`s, etc, that isn't available as a React icon component should be ported. Anything like that linkcards that expects a url to an actual svg file should be switched to icons. Might be some exceptions, such as setting an svg as the background of an element through CSS, but we'll account for those as we go.

GarvinH:
closes #874


GarvinH:
The remaining ones are the background/static html renders

client/src/about/about.js
```diff
@@ -1,13 +1,23 @@
...
+  IconHTML,
+  IconNodeJS,
+  IconReact,
+  IconGit,
+  IconGitHub,
+  IconPython,
+  IconSass,
+  IconGraphQL,
+  IconBaselineCloud,
+} from "src/icons/icons";
```
Stephen-ONeil, line 14: Minor note, will eventually add a linter for this probably, but please use the full file name including extension when importing. Less ambiguous and can also have an effect on the automatic import order rule in some cases, so good to be consistent about.

client/src/components/CardLeftImage.js
```diff
@@ -15,15 +15,7 @@ const CardLeftImage = ({
...
-      {img_src && (
-        <div className="left-img-card__left">
-          <img
-            aria-hidden="true"
-            src={img_src}
-            className="left-img-card__img"
-          />
-        </div>
-      )}
+      {svg && <div className="left-img-card__left">{svg}</div>}
```
Stephen-ONeil, line 18: Used to be aria-hidden, but isn't any more.

Hmm, maybe the icons should default to aria-hidden if they aren't supplied a `title` prop.

client/src/core/ErrorBoundary.js
```diff
@@ -92,14 +94,11 @@ export class ErrorBoundary extends React.Component {
...
-            aria-hidden={true}
-            id="error-boundary-icon"
-            src={get_static_url("svg/not-available.svg")}
-            style={{
-              maxWidth: "100%",
-              width: "400px",
-            }}
+          <IconNotAvailable
+            color="#8c949e"
+            width="400"
```
Stephen-ONeil, line 99: Missing units.

client/src/infographic/bubble_definitions.js
```diff
@@ -2,14 +2,15 @@ import _ from "lodash";
...
+
 import { infograph_href_template } from "./infographic_link.js";
 
 import text from "./bubble_definitions.yaml";
-import svgs from "./bubble_svgs.yaml"; // TODO: use icon componenets instead (some of these are already icon components, some not)
 
 const text_maker = create_text_maker(text);
 
-const get_svg_content = (bubble_id) => _.get(svgs, `${bubble_id}.text`);
+const get_svg_content = (bubble_id) => _.get(svgs, bubble_id);
```
Stephen-ONeil, line 13: Not necessary anymore, cleaner I think to drop this helper func  and `bubble_svgs.js`, just import the icons and put them directly in to `base_configs`.

client/src/components/CardLeftImage.js
Stephen-ONeil: This happens in a number of places in this branch, not going to point them all out, changing the defaults will cover them all.

client/src/infographic/BubbleMenu.js
```diff
@@ -52,11 +52,20 @@ const BubbleMenu = ({ items, active_item_id }) => {
...
-                />
+                <div className="bub-svg" title={description}>
+                  {(() => {
+                    const SVG = (svg_content && svg_content.svg) || null;
+                    return (
+                      SVG && (
+                        <SVG
+                          alternate_color={false}
+                          width="90%"
+                          direct_render
```
Stephen-ONeil, line 63: Is direct render because `.bub-svg` has some specific selector rule that doesn't expect the `g` added by the icon wrapper? I'm not a fan of the `direct_render` prop because it, not super obviously, disables other props like rotating. Also makes it harder to write other selectors targeting the inner content of icons if they don't have a consistent structure.

Could just change the `.bub-svg` selector here. Better would probably be giving the icons a prop like `inner_class_name` or something (probably to add to the `g`?), so that consumers can get a way to optionally target the svg content of an icon without having to hard code a selector against the implementation of the icon components.

client/src/icons/icons.js
```diff
@@ -29,10 +29,12 @@ class _IconWrapper extends React.Component {
       rotation,
       title,
       width,
+      max_width,
```
Stephen-ONeil, line 32: Before we get any more of these styling props, maybe the icons should just have a `svg_style` prop that takes an arbitrary style object and merges it in with the defaults in `svg_styles`? I think `vertical_align` could be something set through the `svg_style` prop. Might want to leave `height` and `width` out since they have the quirk of defaulting to a square if `height` is missing.

Stephen-ONeil (changes requested):
Awesome! Bunch of fiddly comments, although I'm trying not to turn this in to a full icon refactor, haha.

Stephen-ONeil:
> The remaining ones are the background/static html renders.
> 
> I'm thinking right now we can either:
> 
> * Skip these files?
> * embed them into the HTML index?

Meant to respond to this. I'd go with skipping them for now, don't want to require updates to the html index files every time one of those svgs needs to be edited/a new one needs to be added. Will want to batch a bunch of changes involving the html index files eventually though, can revisit this then.

client/src/infographic/bubble_definitions.js
```diff
@@ -28,34 +35,38 @@ const base_configs = [
...
     id: "related",
     title: text_maker("related_title"),
     description: text_maker("related_desc"),
+    svg_content: { svg: IconRelatedBubble },
   },
   {
     id: "all_data",
     title: text_maker("all_data_title"),
     description: text_maker("all_data_desc"),
+    svg_content: { svg: IconDataset },
```
Stephen-ONeil, line 63: Any reason to wrap these in an object? Feels like these lines could just be `svg: IconDataset,` etc.

Stephen-ONeil (changes requested):
Just the one comment! If there's a good reason not to change it, let me know and I'll approve as-is.

client/src/infographic/bubble_definitions.js
GarvinH: Earlier when I was running it without the extra object layer, it would be executed as a function immediately. I think there was some code in `infographic.js` which tried to run all functions immediately

Stephen-ONeil: Ahh, haha, ok. I remember that, these are all set up to optionally be functions (I think of `subject`) rather than values, but the check of `_.isFunction` would return true for these components so they'd be miss-used. 

Well, same effect but a bit cleaner would be `svg: () => IconDataset,` I believe. A little funky, but less weird than the bubble menu having to unwrap the component from an object. Or `svg: [IconDataset]` I think, although react would likely complain about missing `key` values.

Either way, if you leave it as is or change it, please add a comment in `bubble_definitions.js` explaining the weird gotcha that these approaches are handling.

Stephen-ONeil: Or perhaps better, where the bubble definitions get consumed, check `React.isValidElement()` or something alongside checking `_.isFunction`, to avoid calling the components in the wrong place. Then everything is neat and pretty self explanatory, hack-warning comments needed.

Stephen-ONeil (approved):
👍

PR #885, opened 2021-02-08
Replace TODO in typeahead options

GarvinH:
Need to get the translations double checked


Stephen-ONeil:
Oops, thanks!

AlexCLeduc (approved):
Looks good to moi

Issue #865, opened 2021-01-11
Survey popup erroneously logs "displayed" event multiple times 


Stephen-ONeil:
Double check the logging logic for that event, it's inside the render and is logged any time "should_show" is true. Had thought that would be at most once per session, but might be missing something.

Stephen-ONeil:
Might also not be logging all submissions of the short-survey. Double check that too.

GarvinH:
I've momentarily spent some time looking into this, I wasn't able to get should_show to be try more than once...
well there are the super unlikely cases where if you go back and forth between pages that can +1 to the # of pages visited, it would log another display event. But who's really going back and forth between pages and not closing the survey??? (Considering that the survey blocks all other clicks on the site). Or by refreshing the page while the survey is still open, then it logs another event, (but again who is actually doing that?)

Interestingly, the short survey appears to log events twice, instead of once.

GarvinH:
More plausible that a user that revisits a decent amount and clicks on "later" every time but never does it...

Although I'm not sure of the extent that "displayed" gets logged for that to be a reasonable explanation

PR #893, opened 2021-02-12
short survey submitted action is being over submitted, although googl…

ktw1016:
The actual code and google analytics report is contradicting each other:
- Code is over-logging "short_survey_submitted"
- Google Analytics only logged "short_survey_submitted" 9 times, even though it should've logged it at least hundred times

### DIFF
--- a/client/src/components/EmailFrontend.js
+++ b/client/src/components/EmailFrontend.js
@@ -88,6 +88,7 @@ class EmailFrontend extends React.Component {
 
     this.state = {
       template_name: props.template_name,
+      run_on_submitted: true,
       loading: true,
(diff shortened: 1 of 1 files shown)


ktw1016:
I don't think this solves the problem, but not sure how Google Analytic is under-logging "short_survey_submitted" and would be nice if someone can look it over

Stephen-ONeil:
Don't have any immediate thoughts. If you're fully stuck, please document what you've tried back in the original issue and close this PR for now.

ktw1016:
Don't have much to comment on it. We would need to find out the cause of google analytics under logging the short survey submitted first.

ktw1016:
related #865

PR #895, opened 2021-02-14
Infographic search behaviour

Stephen-ONeil:
With the old search, selecting a new infographic subject from an infographic route would (try to) keep you on the same bubble. This behaviour was lost in the port. 

Slightly more annoying than just re-adding a custom `href_template` prop to the infographic's `EverythingSearch`, realized that there was no way to fully disable/exclude config options from the search options menu, which would have meant any custom `href_template` prop would need to be able to handle any sort of searchable subject... boo.

Tweaked `EverythingSearch` so that totally excluding an option key from `initial_search_options` would both remove it from the default search configs AND remove it from the options menu.


ktw1016 (approved):
Great!

Issue #880, opened 2021-02-04
Extract search logic from Typeahead


Stephen-ONeil:
Follow up on #815, refactor the `Typeahead` component to extract the `/search` specific logic from it and make the component useful on its own.

GarvinH:
#880 may help simplify some of the work

Stephen-ONeil:
To clarify, I think completing #8080 might make revisiting #881 worthwhile, as doing this right should simplify the `Typeahead` component itself and make it easier to work on.

PR #899, opened 2021-02-15
Typeahead no options crash

### LINKED ISSUE #896: New search crashes when disabling all search options with an active search
Type in the search bar till you get results, open the search option menu, deselect all options. Crash.

Something I regressed on at some point, oof.

_Could_ handle this at the `EverythingSearch` level, it's fully responsible for the search options menu and could just prevent the final option from being deselected. Sounds reasonable, but there's certain cases where that could be annoying for a user. Consider it carefully before going that direction.

Either way, the bug must also be fixed at the `Typeahead` level, shouldn't hard crash when it gets something unexpected like that.

ktw1016:
closes #896

### DIFF
--- a/client/src/search/EverythingSearch.js
+++ b/client/src/search/EverythingSearch.js
@@ -183,7 +183,6 @@ const EverythingSearch = withRouter(
         search_options_hierarchy,
         this.option_node_to_component
       );
-      console.log(this.state);
 
       return (
         <div className="fcol-md-12 p-0">
--- a/client/src/search/EverythingSearch.js
+++ b/client/src/search/EverythingSearch.js
@@ -183,6 +183,7 @@ const EverythingSearch = withRouter(
         search_options_hierarchy,
         this.option_node_to_component
       );
+      console.log(this.state);
 
       return (
         <div className="fcol-md-12 p-0">
@@ -226,6 +227,18 @@ const EverythingSearch = withRouter(
             .map((_child_node, child_key) => this.state[child_key])
             .some()
             .value();
+          const falsified_child_nodes = {
+            ...this.state,
+            ..._.chain(option_node.child_options)
+              .map((_child_node, child_key) => [child_key, false])
+              .fromPairs()
+              .value(),
+          };
+          const is_only_active_node = _.chain(falsified_child_nodes)
+            .filter()
+            .size()
+            .isEqual(0)
+            .value();
 
           return (
             <div key={option_key} style={{ width: "100%" }}>
@@ -235,6 +248,7 @@ const EverythingSearch = withRouter(
                   active={has_checked_child_option}
                   container_style={{ padding: 3 }}
                   onClick={() =>
+                    !is_only_active_node &&
                     this.setState(
                       _.chain(option_node.child_options)
                         .map((_child_node, child_key) => [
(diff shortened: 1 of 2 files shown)


client/src/search/Typeahead/Typeahead.js
```diff
@@ -194,55 +194,56 @@ export class Typeahead extends React.Component {
...
-                      { is_first_in_group, group_index, result },
-                      result_index
-                    ) => {
-                      const adjusted_result_index = needs_pagination_up_control
-                        ? result_index + 1
-                        : result_index;
-
-                      return [
-                        is_first_in_group && (
+                {!_.isEmpty(this.config_groups) &&
```
ktw1016, line 197: This check isn't actually needed, but I left it there just in case something goes wrong

client/src/panels/panel_declarations/finances/internal_services/internal_services.js
```diff
@@ -40,6 +40,7 @@ export const declare_internal_services_panel = () =>
         const last_year_fte_col = "{{pa_last_year}}";
         const gov_fte_total = programFtes.q(Gov).sum(last_year_fte_col);
         const gov_isc_fte = programFtes.q(isc_tag).sum(last_year_fte_col);
+        console.log(programFtes.q(isc_crsos[0]));
```
Stephen-ONeil, line 43: Stray log

Stephen-ONeil (changes requested):
Opened the dev link, put in a search, clicked a couple options, and crashed with
image

Tried twice and it wasn't clicking the same options that crashed either time, so it depends on the content of the query too probably, I'll try again and a minute and actually note the query + button combo that I tried.

**Edit:** ok, try a search for "agri" and then toggling off the "Organizations" group

Stephen-ONeil:
I didn't look at the code yet, but I'll wait till that bug's fixed.

ktw1016:
So just to summarize, this PR is fixing 2 bugs with Typeahead:
1. If search option is empty (fixed by making checkboxes not clickable to be empty)
2. Typeahead indexes `this.config_groups` with `group_index` but config groups in `this.config_groups` get added or removed so index error happens

Stephen-ONeil:
Oof, hadn't realized the bug I found on the dev link was something that pre-dated this PR, my bad.

client/src/search/EverythingSearch.js
```diff
@@ -228,6 +228,18 @@ const EverythingSearch = withRouter(
...
+            ..._.chain(option_node.child_options)
+              .map((_child_node, child_key) => [child_key, false])
+              .fromPairs()
+              .value(),
+          };
+          const is_only_active_node = _.chain(falsified_child_nodes)
+            .filter()
+            .size()
+            .isEqual(0)
+            .value();
```
Stephen-ONeil, line 242: Had to look this over a few times to get it. It's kind of indirect, not clear what's being checked and why e.g. `falsified_child_nodes` is actually `state_with_current_node_chidlren_inactive` or something, but even with a clearer name it's not clear why we need the variable (which is so that all current node children excluded from the following filter?).

I believe something like this is equivalent
```
const is_only_parent_with_active_children = _.every(
  this.state,
  (is_active, option_key) =>  _.has(option_node.child_options, option_key) || !is_active
);
```

Stephen-ONeil (commented):
Drawback of this approach is that it adds even more coupling between the `EverythingSearch`, it's filter menu, and the `Typeahead` itself. Gets us further from #880 and makes it slightly harder to maintain the three (as coupling always does, more ways for a seemingly safe change in one to possibly break something about the other).

ktw1016:
This branch's original issue is not solved

PR #900, opened 2021-02-15
Storybook

ktw1016:
- [x] a11y
- [x] console


Stephen-ONeil (approved):
Nice! 👍 

Got a plan for how we'll track components that needs stories/organize the work of back filling them (for students this summer probably)?

If you've got any opinions on how stories will be most useful to us, or read any good blogs etc on them, might be good to add a wiki entry on that too. I think we will want some sort of project storybook usage guide eventually at least, maybe once we've used it more and formed opinions/patterns.

ktw1016:
Try visiting https://dev.ea-ad.ca/storybook/storybook/index.html !

Stephen-ONeil:
`DisplayTable` looking funky 
image

I think the storybook itself needs to import the extended bootstrap styles.

Stephen-ONeil (changes requested):
See comment above, need both `common_css` and `extended_bootstrap_css` loaded inside the storybook.

For the location of stories, I think they should either be outside of `src` or right beside their corresponding components. 

Little bit torn on including/excluding from `src`, mainly because I think it's common for devs to limit searches by `client/src`. Do we want storybooks adding a bunch of clutter when trying to search for actual site code? Is it too much of a risk that developers will forget to update things in them, when doing something like a find and replace scoped to `client/src`? Leaning towards leaving them in `src`, devs can just exclude them with `*.stories.js` in files to exclude if they're getting too much clutter.

Either way, I think if they stay in `src` they should live alongside their corresponding modules. Same as `.yaml`/`.scss`/etc. Makes it easier to see when a component does or does not have a story (and also makes it even more likely devs will remember to keep them in-sync with component changes).

ktw1016:
I think stories should co-locate with the components. Easier to find and update them, and less likely to forget to update story along with updating the component

Stephen-ONeil:
> I think stories should co-locate with the components. Easier to find and update them, and less likely to forget to update story along with updating the component
> 
> Question is, should we have a folder for each components? Most of them only have js, scss but if we add stories that's 3 files for a component. (4 if it has yaml). I feel like this can get cluttered pretty easily

Five when we add unit tests, maybe six if we have specific sets of components needing integration tests. Yeah, components will need better organization eventually for sure.

Stephen-ONeil:
Worked out extended bootstrap/common css and cleaned up a little over on #924, please review over there, thanks!

Stephen-ONeil (approved):
👍

Issue #901, opened 2021-02-16
Screen reader not reading properly results panel of dept


GarvinH:
Specifically this panel

![image](https://user-images.githubusercontent.com/25855114/108074870-acd3b580-7037-11eb-84a0-7759cd497542.png)

Tabbing into core responsibility from the top on chrome reads "collpase this panel.... GCInfobase..." and doesn't read the definition in the end. Tabbing backwards into it works fine however.

A more general problem in the code: tabbing into the status filter reads for each one `Activate to filter "[object Object]"` which happens in the non-a11y version only.
The "Text" that is passed into the status filter for the non-a11y version is actually a group of elements (presumably a span?).

One option we can take is to split the text prop into two props (render and sr-only props). There might be a better way of handling it so I'll leave that up in the air.

ktw1016:
Just tried it and reads correctly `Activate to filter "Met"...`, I'm guessing this was fixed in the PR

PR #903, opened 2021-02-16
Change filtertable to use newly added aria text prop

GarvinH:
related #901

### DIFF
--- a/client/src/panels/panel_declarations/results/result_components.js
+++ b/client/src/panels/panel_declarations/results/result_components.js
@@ -319,7 +319,7 @@ const StatusIconTable = ({
             active_list.length === 0 ||
             _.indexOf(active_list, status_key) !== -1,
           count: icon_counts[status_key] || 0,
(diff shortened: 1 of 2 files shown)


client/src/panels/panel_declarations/results/result_components.js
```diff
@@ -329,9 +329,8 @@ const StatusIconTable = ({
             >
               {result_simple_statuses[status_key].text}
             </span>
-          ) : (
-            result_simple_statuses[status_key].text
           ),
+          ariaText: result_simple_statuses[status_key].text,
```
Stephen-ONeil, line 333: Minor point, we usually use snake case for props. Not 100% consistent, we tended to go back and forth very early on. Personally, I prefer to reserve camel case for the HTML attribute props (like `className`, etc).

Stephen-ONeil (approved):
Approved, but see the one minor comment before merging.

PR #906, opened 2021-02-17
Handle Typeahead search_config prop changes properly

### LINKED ISSUE #896: New search crashes when disabling all search options with an active search

Stephen-ONeil:
Alternate to #899, closes #896 as well as the additional bug caught during #899 review.

### DIFF
--- a/client/src/search/Typeahead/Typeahead.js
+++ b/client/src/search/Typeahead/Typeahead.js
@@ -17,10 +17,6 @@ import "./Typeahead.scss";
 
 const { text_maker, TM } = create_text_maker_component(text);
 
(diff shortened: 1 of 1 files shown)


Stephen-ONeil:
Haha, branch name's too long for Mongo, will re-push

PR #908, opened 2021-02-18
remove inactive status on estimates departments

ktw1016:


client/src/EstimatesComparison/scheme.js
```diff
@@ -142,9 +142,7 @@ const get_keys_in_sups = (include_stat) =>
     .value();
 
 const calculate_percent_value = (current_value, comparison_value) => {
```
Stephen-ONeil, line 144: Fine if `current_value` is never undefined, but will return a `NaN` if it is. Maybe add a check for that or just an `|| 0`  on the default return case?

client/src/EstimatesComparison/EstimatesExplorerComponent.js
```diff
@@ -64,12 +64,12 @@ const DetailedAmountsByDoc = ({ amounts_by_doc }) => {
...
                   <td>{estimates_docs[doc_code][lang]}</td>
                   <td>
-                    {amount_last_year && (
+                    {(amount_last_year || amount_last_year === 0) && (
                       <Format type="compact2" content={amount_last_year} />
                     )}
                   </td>
                   <td>
-                    {amount_this_year && (
+                    {(amount_this_year || amount_this_year === 0) && (
```
ktw1016, line 72: This is kinda sketchy but theoretically this should be set to `false`.....
Screen Shot 2021-02-18 at 12 04 18 PM

Stephen-ONeil: Aren't these the cases where there's some non-zero data, but the net value is 0 because they cancel out? I think that's the point of that check, we only want 0s if they're because the net value is zero. If ALL values are zero then we want to display it as blank/what used to be inactive (except inactive was always being accidentally applied to the net-zero case as well).

ktw1016: I don't think these checks were for net value of zero.. because I think this is strictly for table of breakdown. The status (active or inactive) is determined by `current_value`

client/src/EstimatesComparison/scheme.js
```diff
@@ -172,7 +173,14 @@ function get_data_by_org(include_stat) {
 
       let current_value = 0;
       if (current_doc_is_mains) {
-        current_value = _.sumBy(rows, "current_value") || 0;
+        // Need to check if all current_values are 0. Inactive if all are 0, else sum of it. Need to do an extra check in case sum=0
+        const org_is_active = !_.every(
+          rows,
+          ({ current_value }) => current_value === 0
+        );
```
Stephen-ONeil, line 180: Not really worth changing, but `_.every` has to check every entry. `_.some` can exit on the first false. So, for a basically meaningless optimization, this is better:
```
_.some(rows, ({ current_value }) => current_value !== 0 )
```

PR #914, opened 2021-02-22
Use in-house aria modal

### LINKED ISSUE #909: Replace react-aria-modal in RPB
`react-aria-modal` hasn't updated in two years, uses deprecated React lifecycle hooks which will be broken by React 18.

We already use `react-bootstrap`'s `Modal` in some places. Might take a small bit of work to match the current RPB modal's behaviour, but should be an easy swap-out. Worth double checking `react-aria-modal`'s promised a11y, and make sure we have at least an equivalent level of a11y out of the bootstrap modal.

GarvinH:
closes #909


client/src/components/modals_and_popovers/AriaModal.js
```diff
@@ -0,0 +1,57 @@
...
+
+import "./bootstrap_modal_exstension.scss";
+
+export class AriaModal extends React.Component {
+  constructor(props) {
+    super(props);
+
+    if (!props.titleId) {
+      throw new Error("Must have a prop titleId");
+    }
```
Stephen-ONeil, line 14: Is using a separate, id'ed, component plus `aria-labelledby` just a holdover from the old `AriaModal`? Doesn't seem like it's the most flexible setup, probably more sense to give the modal an optional header (`Modal.Header` component with `react-bootstrap`) or just pass in a title string.

In fact, I don't see the RPB's supposed `#tbp-title` element in our source code so it may be unused. In that sense, this check's not actually very strong because it doesn't guarantee that a valid DOM node with id=`titleID` exists (or will exist, at render time). If this component wants to enforce a proper ARIA label then it would have to request it as a string and handle the labeling itself.

Stephen-ONeil (changes requested):
Just some little fiddly points, everything looks good on the site.

Other TODOs:
  - [x] drop the old component from package.json.

client/src/components/modals_and_popovers/AriaModal.js
```diff
@@ -0,0 +1,57 @@
+import classNames from "classnames";
+import React from "react";
+import { Modal } from "react-bootstrap";
+import FocusLock from "react-focus-lock";
+
+import "./bootstrap_modal_exstension.scss";
+
+export class AriaModal extends React.Component {
```
Stephen-ONeil, line 8: Eh, woops, hadn't hit submit on one of my comments when I put that review in.

Gist of it was something like this. I don't find `AriaModal` to be a useful name. For one thing, it looks like the react-bootstrap `Modal` uses all the appropriate ARIA attributes already, so all out modals/popups using that are also "AriaModal"s. Aside from that, ARIA !== a11y and a11y what we're more interested.

Anyway, I think the descriptive name for this would be something like `FocusLockedModal` because being focused locked it the distinctive property of this component.

client/src/components/modals_and_popovers/AriaModal.js
```diff
@@ -0,0 +1,57 @@
...
+    applicationNode.setAttribute("aria-hidden", value);
+  };
+
+  componentDidMount() {
+    this.hide_application_node_aria("true");
+  }
+
+  componentWillUnmount() {
+    this.hide_application_node_aria("false");
+  }
```
Stephen-ONeil, line 31: Looks like the `Modal` component has the `aria-modal="true"` attribute which should make this unnecessary (although it's relatively new for screen readers, so it's potentially worth keeping this step for backwards compatibility).

I don't see `getApplicationNode` actually being used though, so if this is some leftover code from elsewhere then feel free to drop it and rely on the `react-bootstrap` `Modal` handling things correctly.

Stephen-ONeil: Oh, double checked, the `Modal` does already take care of also setting `aria-hidden` everywhere outside of the open modal, so that's good!

client/src/components/modals_and_popovers/AriaModal.js
GarvinH: yea the titleId and aria-labelledby was taken from the old AriaModal.

As for `#tbp-title`, I managed to find it being used in `Tablepicker.js`

GarvinH: I think i'll go with the handling label by itself

GarvinH: so it seems like either it will

a) focus on the modal itself, which doesn't have proper focus trapping but will read the aria label
b) focus immediately on the first tabbable child, which means aria-label is skipped

I tried using the `Modal.Header` but doesn't seem to help...

Stephen-ONeil: ~How about try and aim for b and make sure the `Modal.Header` is tabbable (if not directly then wrap it in a tabbable element)? I think the header's text should usually be the same as the `aria-label`, but if there's a case for needing additional descriptiveness for screen readers then `aria-label` could go on that first focusable (as long as it's not just a repeat of the header content).~

Oh, that's likely just a matter of making a new, markdown free, text key to use for the label. The alternate hacky approach is to strip the markup from the string. If you want to go with hacky, look up `DOMPurify`. We're already using it to sanitize input received from Titan that can contain HTML, but it can also just fully strip all HTML from a string, I believe.

client/src/rpb/TablePicker.js
```diff
@@ -311,7 +311,7 @@ class TaggedItemCloud extends React.Component {
           <div>
             <div className="row item-cloud-row">
               {_.map(items_split, (item_group, group_index) => (
-                <div key={group_index} className="col-md-4 item-cloud-col">
+                <div key={group_index} className="col-lg-4 item-cloud-col">
```
Stephen-ONeil, line 314: More non-flex grid stuff here. Check rest  of file too, thanks.

client/src/rpb/index.js
```diff
@@ -345,7 +345,8 @@ class RPB extends React.Component {
                     }, 200);
                   }
                 }}
-                titleId="tbp-title"
```
Stephen-ONeil, line 348: The corresponding element with `id="tbp-title"` doesn't need that id anymore, might as well clean that up too 👍

client/src/components/modals_and_popovers/FocusLockedModal.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+  }
+
+  render() {
+    const {
+      mounted,
+      children,
+      additional_dialog_class,
+      onExit,
+      ariaLabel,
```
Stephen-ONeil, line 24: Type cases

client/src/components/modals_and_popovers/FocusLockedModal.js
```diff
@@ -0,0 +1,44 @@
...
+
+import "./bootstrap_modal_exstension.scss";
+
+export class FocusLockedModal extends React.Component {
+  constructor(props) {
+    super(props);
+
+    if (!props.ariaLabel) {
+      throw new Error("Must have a prop ariaLabel");
+    }
```
Stephen-ONeil, line 15: Only asserted at construction, what if the prop was updated to be empty later?

If you want to properly mark a prop required (and may as well mark it as a string type while we're at it), look in to [React propTypes](https://reactjs.org/docs/typechecking-with-proptypes.html).

Stephen-ONeil (changes requested):
Few more little things

Stephen-ONeil (approved):
👍  🎉

PR #918, opened 2021-02-23
Upgrade Nivo

### LINKED ISSUE #910: Outdated Nivo version, large number of React warnings
We're still locked to Nivo `0.52.1`. Now that we're on React 17, it's giving off an overwhelming number of React dev tool warnings for deprecated lifecycle methods. Nothing will break until React 18, but it's going to make it harder to see any new console warnings in the meantime.

Options:
  - [ ] upgrade Nivo
    - ... no migration docs, tons of undocumented and breaking prop changes
    - Nivo is STILL pre-1.0, so this won't be the only time we have to do a blind-upgrade 
    - allows/requires us to also upgrade D3 to the latest version at the same time
  - [ ] drop Nivo for a new chart library
    - unknown effort ceiling, but at minimum this is a very big task
  - [ ] ... use `patch-package` to add `UNSAFE__` to the deprecated lifecycle methods
    - extremely easy, relative to other options
    - delays the inevitable task of dealing with Nivo
    - React 18 proof, for these breaking changes at least...
    
Well, almost definitely going to be option 3, but I'd prefer we at least take a stab at option 1.

ktw1016:
closes #910

- [x] tooltip adjustments
- [x] goco broken
- [x] upgrade existing d3 (and transpile(?) for IE11)

Latest version still throwing React warnings:
https://github.com/plouc/nivo/issues/667
https://github.com/plouc/nivo/issues/884


client/src/charts/wrapped_nivo/WrappedNivoPie.js
```diff
@@ -182,6 +180,11 @@ WrappedNivoPie.defaultProps = {
     bottom: 60,
     left: 50,
   },
+  theme: {
```
ktw1016, line 183: Inconsistency from Nivo: This `boxShadow` exists for Nivo bar and circle-pack but not for line and pie. So can't put this all together into nivo_common tooltip factory. I'll try making an issue on Nivo but they'll probably make me start a mini project to replicate it🤮

Stephen-ONeil:
> Latest version still throwing React warnings:
> [plouc/nivo#667](https://github.com/plouc/nivo/issues/667)
> [plouc/nivo#884](https://github.com/plouc/nivo/issues/884)

😭

This can still close #910, but we should probably open a second issue linking those two nivo issues and reminding us to watch for a version that closes them. Or I guess now that we're caught up again it would be best to stay up to date, and just work through each nivo update as it comes in.

Stephen-ONeil (changes requested):
Just looking at the site to start:
  - [ ] tooltips don't update as expected when moving the cursor around on a graph
    - Example: hover over a year on a line graph. Dotted line and tooltip for set of points appears. Move cursor to another year. Dotted line and tootlip don't move, but contents of tooltip do update
image
  - [ ] graphs don't rescale well
    - Ex: this line graph get particularly messy, but in general the axis scales/graph area and actual graph content get out of sync/can go out of bounds
image
    - Oh, and some stay broken even after going back to the original size, haha
image
  - lots of problems in the people graphs with the toggle buttons, although I'm guessing some of these are from the toggling itself and some are just graphs with general problems/bad configs
    - [ ] toggle from normal bar to % bar, the y-axis value formatting is wrong
image
    - [ ] toggle to horizontal bars, y-axis values and horizontal grid lines near the top are messed up?
image
    - [ ] toggle twice to get to stacked lines, but the lines from the non-stacked versions stick around
image

ktw1016:
okay so dev build works fine, and doesn't cause any of those bugs. It actually seems like it's just prod build. Been playing around, specifically webpack's mode in `production` is breaking Nivo somehow.

client/build_code/webpack_common.js
```diff
@@ -58,9 +58,14 @@ const get_rules = ({ target_ie11, language, is_prod_build }) => {
...
     },
+    {
+      //NIVO_TODO: Temporary workaround for bug from Nivo. Will need to delete this as it gets fixed in newer versions
+      test: /react-spring/,
+      sideEffects: true,
+    },
     {
       // node modules that specifically require transpilation...
-      include: /node_modules\/(graphiql|graphql-language-service-.*|codemirror-graphql|codemirror)/,
+      include: /node_modules\/(graphiql|graphql-language-service-.*|codemirror-graphql|codemirror|d3-scale|@nivo\/bar|@nivo\/circle-packing|@nivo\/core|@nivo\/line|@nivo\/pie|)/,
```
Stephen-ONeil, line 68: Could be `@nivo\/*`? OR are there some nivo packages that don't need transpilation?

client/build_code/webpack_common.js
```diff
@@ -58,9 +58,14 @@ const get_rules = ({ target_ie11, language, is_prod_build }) => {
       use: js_module_loader_rules,
       sideEffects: true,
     },
+    {
+      //NIVO_TODO: Temporary workaround for bug from Nivo. Will need to delete this as it gets fixed in newer versions
+      test: /react-spring/,
+      sideEffects: true,
```
Stephen-ONeil, line 64: Seems good enough, please open an issue as a reminder about this once merged.

Stephen-ONeil: Likely a minor difference, but the fix nivo maintainers applied to their own site was more granular, identifying specific `react-spring` modules with side-effects rather than flagging the whole thing as having side-effects. Their approach is likely a little more optimized for bundle size (but if this is just temporary, it's probably a minor enough difference to ignore).

See https://github.com/plouc/nivo/blob/master/scripts/patch-react-spring.js

client/package.json
```diff
@@ -147,9 +144,7 @@
     "ci_route_load_tests": "node browser-tests/route-load-tests.js --browser chromium --sandbox false --concurrency 8",
     "ci_serve": "http-server -s -a 127.0.0.1",
     "serve-loopback": "http-server -a 127.0.0.1",
-    "serve": "http-server",
-    "postinstall": "npx patch-package",
-    "postci": "npx patch-package"
```
Stephen-ONeil, line 152: Could drop the patch-package dependency (unless we use it for the more granular fix mentioned above).

Stephen-ONeil (changes requested):
What's going on with the treemap? Guess I should have looked at Garvin's branch, but do you know why it's not actually using the nivo treemap? And why `TreeMapHtmlNode.js` seems to be a copy of a file from nivo's source code?

ktw1016:
Garvin's treemap is using nivo treemap, it's just HTML alternative instead of SVG. Found here: https://nivo.rocks/treemap/html/


`TreeMapHtmlNode.js` is almost a copy but not quite. He did a workaround on a style for one of the `span`, 
```
top: -5,
left: -5,
width: 10,
height: 10,
transform: animatedProps.labelHtmlTransform,
```
is replaced with `width: 100%, height 100%`

It's broken with the original style setting but you're right, this should be watched closer for sure..

Stephen-ONeil:
Ah, should use patch package for that fix (haha, guess it stays). That way we can keep importing the module from nivo and receiving updates to it. If nivo ever touches that specific style up-stream then patch package will alert us automatically. Much better than copying the nivo package in to our repo 👍

ktw1016:
yeah good point, thanks for catching that!👍

client/build_code/webpack_common.js
ktw1016: I should actually check these individual packages

client/build_code/webpack_common.js
ktw1016: I changed it, but then just checked that the modules listed there are all of the modules in `@react-spring`

Stephen-ONeil (approved):
Approved! 🎉  

One minor comment you can choose to address before merging, `FlatTreeMapViz` is a weird legacy name, I think for consistency it should be `WrappedNivoTreeMap` or something?

PR #920, opened 2021-02-25
Use nivo for flat tree maps

### LINKED ISSUE #897: Deal with legacy FlatTreeMapViz
`FlatTreeMapViz` is pretty legacy. A headache to use, one of the required props `node_render` is a function that returns an HTML string for it to internally feed to some D3, oof.

Pick one of:
  - see if we can't just use Nivo's treemaps, don't remember why we didn't
  - refactor it to be sensible (very likely that it does not actually need D3 imperative rendering and a bunch of HTML strings, probably could just use the D3 hierarchy packages to calculate scaling while still fully rendering with React)
  - drop it and redesign the single panel that uses it
  
I'd prefer the first if possible, because a good treemap has its uses. Three is the next up, because it doesn't really contribute much to the single panel using it right now. Two is only worth the effort if there's a real dealbreaker in Nivo's treemap implementation.

GarvinH:
closes #897

must merge into #918 as it has nivo-core


Stephen-ONeil:
@ktw1016 I think you should review this and merge it down in to the nivo update branch when it's ready, since the treemap will need the same tooltip fixes (whatever those end up being) as the rest of the updated nivo charts.

ktw1016 (changes requested):
Pretty much looks all good to me, great work! I have few comments:

Is it only me, trying to resize the window feels really slow for this branch? I wonder if it could be from react-spring. That's just for giving spring animation on resize right? Can you play around to see the performance difference on resize with the library and without the library?

It may need some break points, on chrome width 1190px:
Screen Shot 2021-02-26 at 11 04 09 AM

ktw1016 (approved):
👍

PR #921, opened 2021-03-01
Paginated displaytable

### LINKED ISSUE #890: Use React Virtualized in RPB
Or perhaps just in general on any `DisplayTable` over a certain size. Rendering and scrolling some of these giant tables can be sluggish even on an ok laptop with a modern browser. Very painful on an old unoptimized browser or a slower device.

GarvinH:
closes #890

### DIFF
--- a/client/.storybook/main.js
+++ b/client/.storybook/main.js
@@ -14,15 +14,6 @@ module.exports = {
     "@storybook/addon-links",
     "@storybook/addon-essentials",
     "@storybook/addon-a11y",
-    "@storybook/addon-postcss",
-    {
-      name: "@storybook/addon-postcss",
-      options: {
-        postcssLoaderOptions: {
-          implementation: require("postcss"),
-        },
-      },
-    },
   ],
   webpackFinal: async (config) => ({
     ...config,
--- a/client/.storybook/main.js
+++ b/client/.storybook/main.js
@@ -14,6 +14,15 @@ module.exports = {
     "@storybook/addon-links",
     "@storybook/addon-essentials",
     "@storybook/addon-a11y",
+    "@storybook/addon-postcss",
+    {
+      name: "@storybook/addon-postcss",
+      options: {
+        postcssLoaderOptions: {
+          implementation: require("postcss"),
+        },
+      },
+    },
   ],
   webpackFinal: async (config) => ({
     ...config,
--- a/client/src/components/DisplayTable/DisplayTable.js
+++ b/client/src/components/DisplayTable/DisplayTable.js
@@ -181,6 +181,7 @@ export class DisplayTable extends React.Component {
       util_components,
       show_pagination_controls,
       disable_pagination,
+      page_size_num_options_max,
     } = this.props;
     const {
       sort_by,
@@ -406,6 +407,7 @@ export class DisplayTable extends React.Component {
                           on_select={this.change_page_size}
                           page_size_increment={page_size_increment}
                           num_items={_.size(sorted_filtered_data)}
+                          num_options_max={page_size_num_options_max}
                         />
                       )}
                     </div>
@@ -613,6 +615,7 @@ export class DisplayTable extends React.Component {
 DisplayTable.defaultProps = {
   page_size_increment: 100,
   show_pagination_controls: true,
+  page_size_num_options_max: 5,
 };
 
 // Wrapper component that picks column configs based on the size of data. Currently cannot pick table utils
--- a/client/src/components/DisplayTable/DisplayTable.js
+++ b/client/src/components/DisplayTable/DisplayTable.js
@@ -365,7 +365,7 @@ export class DisplayTable extends React.Component {
 
     const number_of_pages = paginated_data.length;
 
-    const page_selector = enable_pagination && !loading && (
+    const page_selector = enable_pagination && (
       <SelectPage
         num_pages={number_of_pages}
         current_page={current_page}
--- a/client/src/components/DisplayTable/DisplayTable.js
+++ b/client/src/components/DisplayTable/DisplayTable.js
@@ -44,6 +44,39 @@ const get_col_configs_with_defaults = (column_configs) =>
     ...column_config_defaults,
     ...supplied_column_config,
   }));
+const get_default_state_from_props = (props) => {
+  const { unsorted_initial, column_configs } = props;
+
+  const col_configs_with_defaults = get_col_configs_with_defaults(
+    column_configs
+  );
+
+  const visible_col_keys = _.chain(col_configs_with_defaults)
+    .pickBy((col) => col.initial_visible)
+    .keys()
+    .value();
+
+  const sort_by = unsorted_initial
+    ? null
+    : _.chain(col_configs_with_defaults)
+        .pickBy((col) => col.is_sortable)
+        .keys()
+        .first()
+        .value();
+
+  const searches = _.chain(col_configs_with_defaults)
+    .pickBy((col) => col.is_searchable)
+    .mapValues(() => "")
+    .value();
+
+  return {
+    visible_col_keys,
+    sort_by,
+    descending: false,
+    searches,
+    initial_props: props,
+  };
+};
 
 /* Assumption: DisplayTable assumes 1st column to be string that describes its row
   - If total row exists, 1st column will have the word "Total"
@@ -88,6 +121,32 @@ export class DisplayTable extends React.Component {
       page_size: props.page_size_increment,
       current_page: 0,
     };
+  }
+  static getDerivedStateFromProps(nextProps, prevState) {
+    if (nextProps !== prevState.initial_props) {
+      const new_default_state = get_default_state_from_props(nextProps);
+
+      const prev_sort_still_valid = _.includes(
+        new_default_state.visible_col_keys,
+        prevState.sort_by
+      );
+
+      const reconciled_searches = {
+        ...new_default_state.searches,
+        ..._.pick(prevState.searches, new_default_state.visible_col_keys),
+      };
+
+      return {
+        ...new_default_state,
+        ...(prev_sort_still_valid && {
+          sort_by: prevState.sort_by,
+          descending: prevState.descending,
+        }),
+        searches: reconciled_searches,
+      };
+    } else {
+      return null;
+    }
   }
 
   change_page_size = (new_page_size) =>
--- a/client/src/components/DisplayTable/DisplayTable.js
+++ b/client/src/components/DisplayTable/DisplayTable.js
@@ -93,7 +93,7 @@ export class DisplayTable extends React.Component {
       sort_by,
       descending: false,
       searches,
-      state_paginate_by: 10,
+      state_paginate_by: 100,
       current_page: 0,
     };
   }
@@ -577,7 +577,7 @@ export class SmartDisplayTable extends React.Component {
       <DisplayTable
         {...this.props}
         column_configs={smart_column_configs}
-        size_to_paginate={size_to_paginate || 25}
+        size_to_paginate={size_to_paginate || 101}
       />
     );
   }
--- a/client/src/components/DisplayTable/DisplayTable.js
+++ b/client/src/components/DisplayTable/DisplayTable.js
@@ -18,6 +18,7 @@ import {
   DisplayTableCopyCsv,
   DisplayTableDownloadCsv,
   DisplayTableColumnToggle,
+  PageSelector,
 } from "./DisplayTableUtils.js";
 
 import text from "./DisplayTable.yaml";
@@ -303,13 +304,25 @@ export class DisplayTable extends React.Component {
       .value();
 
     const num_pages = _.toInteger(
-      paginate_by ? _.size(sorted_filtered_data) / paginate_by : 0
+      paginate_by ? _.size(sorted_filtered_data) / paginate_by : -1
     ); //start counting from 0 (-1 from the literal amount of pages)
 
     const paginated_data = _.chunk(
       sorted_filtered_data,
       paginate_by || _.size(sorted_filtered_data)
     )[current_page];
+
+    const change_page = (page) => {
+      this.setState({ current_page: page });
+    };
+
+    const page_selector = num_pages >= 0 && (
+      <PageSelector
+        num_pages={num_pages}
+        current_page={current_page}
+        change_page={change_page}
+      />
+    );
 
     return (
       <div
@@ -318,6 +331,7 @@ export class DisplayTable extends React.Component {
           util_components_with_defaults && "display-table-container--with-utils"
         )}
       >
+        {page_selector}
         <table
           className={classNames(
             "table",
@@ -504,6 +518,7 @@ export class DisplayTable extends React.Component {
             </tbody>
           )}
         </table>
+        {page_selector}
 
         {sorted_filtered_data.length === 0 && (
           <TM
--- a/client/src/components/DisplayTable/DisplayTable.js
+++ b/client/src/components/DisplayTable/DisplayTable.js
@@ -19,7 +19,7 @@ import {
   DisplayTableDownloadCsv,
   DisplayTableColumnToggle,
   PageSelector,
-  PageinateBySelector,
+  SelectPageSize,
 } from "./DisplayTableUtils.js";
 
 import text from "./DisplayTable.yaml";
@@ -117,7 +117,7 @@ export class DisplayTable extends React.Component {
       sort_by,
       descending: false,
       searches,
-      state_paginate_by: 100,
+      page_size: props.page_size_increment,
       current_page: 0,
     };
   }
@@ -146,15 +146,15 @@ export class DisplayTable extends React.Component {
         },
       }
       */,
+      page_size_increment,
       util_components,
-      size_to_paginate,
     } = this.props;
     const {
       sort_by,
       descending,
       searches,
       visible_col_keys,
-      state_paginate_by,
+      page_size,
       current_page,
     } = this.state;
 
@@ -314,61 +314,29 @@ export class DisplayTable extends React.Component {
       .reverse()
       .value();
 
-    const {
-      paginate_by,
-      paginated_data,
-      change_paginate_by,
-      show_pagination_controls,
-      page_selector,
-    } = (() => {
-      const paginate_by =
-        size_to_paginate && _.size(sorted_filtered_data) > size_to_paginate
-          ? state_paginate_by
-          : 0;
+    const paginated_data = _.chunk(sorted_filtered_data, page_size);
 
-      const num_pages_exact = paginate_by
-        ? _.size(sorted_filtered_data) / paginate_by
-        : -1;
+    const number_of_pages = paginated_data.length;
+    const show_pagination_controls = number_of_pages > 1;
 
-      const num_pages =
-        num_pages_exact >= 0 && _.isInteger(num_pages_exact) //if we have the exact amount of items to fill an entire page, it will start a new empty page
-          ? num_pages_exact - 1 //thus we want to get rid of the extra page (as it has no content)
-          : _.toInteger(num_pages_exact); //start counting from 0 (-1 from the literal amount of pages)
+    const change_page_size = (new_page_size) => {
+      this.setState({ page_size: new_page_size, current_page: 0 });
+    };
 
-      const paginated_data = _.chunk(
-        sorted_filtered_data,
-        paginate_by || _.size(sorted_filtered_data)
-      )[current_page];
+    const change_page = (page) => {
+      this.setState({ current_page: page });
+      this.first_data_ref.current.focus();
+    };
 
-      const change_page = (page) => {
-        this.setState({ current_page: page });
-        this.first_data_ref.current.focus();
-      };
-
-      const change_paginate_by = (new_paginate_by) => {
-        this.setState({ state_paginate_by: new_paginate_by, current_page: 0 });
-      };
-
-      const show_pagination_controls = num_pages >= 0;
-
-      const page_selector = show_pagination_controls && (
-        <PageSelector
-          num_pages={num_pages}
-          current_page={current_page}
-          change_page={change_page}
-          show_select
-          num_col={_.size(visible_ordered_col_keys)}
-        />
-      );
-
-      return {
-        paginate_by,
-        paginated_data,
-        change_paginate_by,
-        show_pagination_controls,
-        page_selector,
-      };
-    })();
+    const page_selector = show_pagination_controls && (
+      <PageSelector
+        num_pages={number_of_pages}
+        current_page={current_page}
+        change_page={change_page}
+        show_select
+        num_col={_.size(visible_ordered_col_keys)}
+      />
+    );
 
     return (
       <div
@@ -404,9 +372,10 @@ export class DisplayTable extends React.Component {
                     style={{ display: "flex", justifyContent: "space-between" }}
                   >
                     {show_pagination_controls && (
-                      <PageinateBySelector
-                        selected={paginate_by}
-                        on_select={change_paginate_by}
+                      <SelectPageSize
+                        selected={page_size}
+                        on_select={change_page_size}
+                        page_size_increment={page_size_increment}
                         num_items={_.size(sorted_filtered_data)}
                       />
                     )}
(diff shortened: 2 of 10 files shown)


GarvinH:
Just waiting on translations now

ktw1016:
Pretty cool! I haven't looked at the code but just looking at the design I have some opinions
- Can we have the page numbers like `1 2 3 4 5 ... 8` (if you know what I mean)
- Dropping "Select page",  "Next", "Last" buttons in favour of the above (potentially a11y concerns? How about just keeping "Select page" for a11y version?)
- How about showing 100 items by default (seems safe to be performant enough?), have a checkbox like "Show all items" so that toggling that either shows 100 or all items. And drop "Items per page" button

Just my quick initial thoughts. Any thoughts on these @GarvinH @Stephen-ONeil ?

Stephen-ONeil:
Filtering crashed, sorting only applies within the current page. I think what we'd actually want is for filtering and sorting to apply the the global dataset, and to reset the pagination state on change.

With that in mind, I'd suggest something more like this:
image
(although I like Taewan's suggestion to drop fist/previous/next/last for something like `1 ... 7 8 9 10 11 ... 57`)

Controls that can effect other controls need to come first. E.g. if sorting resets pagination, then pagination controls have to come after sorts. Changing the number of items per page doesn't change filtering/sorting, but does change pagination, so it should come in between filtering and pagination selection. "Before" can either be vertically or just first on the same line.

Stephen-ONeil:
Oh, and the `page-selector` class on the pagination control part of the header adds an annoying white line on the left. I don't see why that `tr` should just be a `table-header` like the rest?

Stephen-ONeil:
One more thing, these `react-select` `Select`s look pretty good (although their css-in-js generated classnames are a bother)! 

We already have out own `Select` component though. Please either just use that, or replace our existing `Select`s with this library's.

GarvinH:
the page-selector class allows for the controls at the bottom of the table to be regularly styled otherwise it would look like this

image


I was thinking that it would be nice to have it in case of larger # of items/ table so that users wouldn't have to scroll back to the top.


In retrospect the pagination doesn't even scroll users back to the top anyways... gonna remove it haha
(unless we wanted to keep it?)

GarvinH:
Decided to drop the select since users are (probably) more likely to be using the filter to isolate the pages rather than find by a specific page

ktw1016:
I like it much better ! Just some minor thoughts on the design:
- I think clicking "..." should go to next/previous pages for e.g., if it's like: "1 2 3 ... 9", clicking ... would click on 4
- Placement of the pages could be at the very top of the table I think. Either above the other util buttons, or on the same level as other util buttons in the middle?
- I liked having the pages at the bottom of the table

Let me know how you or others think !

Stephen-ONeil:
> * I think clicking "..." should go to next/previous pages for e.g., if it's like: "1 2 3 ... 9", clicking ... would click on 4
Haha, what if the `...` was the (hidden) drop down for selecting a page? Mostly joking about that.

> * Placement of the pages could be at the very top of the table I think. Either above the other util buttons, or on the same level as other util buttons in the middle?
I'm sticking to my advice of putting it below the sorting and filtering. Changing the page doesn't change sorting and filtering but changing sorting and filtering resets pagination. For screen readers (but really just anyone, IMO), anything that has a non-obvious effect/side-effect on another component should be placed above the effected component. What if someone picked a page, scrolled down till the paginate was out of view, and then changed a sort? They'd not see that pagination had reset.

GarvinH:
Spent a while thinking scrolling was broken, but it turned out that focusing on the button was also scrolling it to the bottom as well.

Decided to drop the page selector focus in favour for scrolling to the top of the table

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -266,6 +280,46 @@ export class DisplayTable extends React.Component {
...
+    const change_page = (page) => {
+      this.setState({ current_page: page });
+      this.first_data_ref.current.focus();
+    };
+
+    const change_paginate_by = (new_paginate_by) => {
+      this.setState({ state_paginate_by: new_paginate_by, current_page: 0 });
+    };
+
+    const show_pagination_controls = num_pages >= 0;
```
ktw1016, line 311: Minor thing, I think it'd look slightly cleaner if all of pagination variables were grouped inside an immediately invoked function
`const { ... } = () => {...}()`

ktw1016 (commented):
I think it mostly looks good, just pretty minor comments !

Stephen-ONeil:
Oops, haha, I'm also in the middle of a review. I'll try not to duplicate Taewan's comments, will hopefully finish soon (got a meeting though)

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -266,6 +280,62 @@ export class DisplayTable extends React.Component {
...
+
+      const page_selector = show_pagination_controls && (
+        <PageSelector
+          num_pages={num_pages}
+          current_page={current_page}
+          change_page={change_page}
+          show_select
+          num_col={_.size(visible_ordered_col_keys)}
+        />
+      );
```
Stephen-ONeil, line 328: I'd suggest putting `PageinateBySelector` in the same row as `PageSelector`. They share the same requirement (`show_pagination_controls`) and are tightly coupled conceptually.

Stephen-ONeil (changes requested):
Hmm, there's some thorny code here that's a bit hard to break down piece-by-piece. I'm going to put up a pull request off this branch implementing the suggestions I'd have and we can look that over together and discuss. It won't be all of the code changes I'd suggest, but the rest can happen once the more complicated set it dealt with.
**Edit:** see #943

Minor design thoughts (just things to consider for now, wait till the previously mentioned PR is merged in before continuing here):
  - [ ] would be nice if the number of pagination buttons was consistent (including the ...'s). Right now the number of buttons changes depending on the distance between the selected button and either end (for example, if there's 9 pages then the number of buttons changes every time you pick a new page, haha)
  - On the items per page menu:
    - [ ] maybe more in-between options for items per page? Increase by hundreds or something? Also maybe just say "all" instead of the exact number of items for the full view
    - [ ] clicking the number beside the radio button should also switch items
    - [ ] switching to a large number of items per page can be pretty slow, with no indication that your click was successful/something's happening. Might be fiddly, but can the menu switch to an intermediate state with a spinner before setting off the slow full render?
  - [ ] applying a filter that returns less than 100 items disables the pagination controls, is that intentional? Little fiddly, but just to reduce the amount of UI changes a user has to deal with when applying a filter, could the pagination controls stay? Base where they should exist on the total dataset rather than the filtered one? Obviously they're useless when you've filtered down to below the current page size, this is just a suggestion about consistency/reducing less obvious UI side effects

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -293,18 +348,31 @@ export class DisplayTable extends React.Component {
...
-                      onClick={() => this.refs.first_data.focus()}
-                    >
-                      {text_maker("skip_to_data")}
-                    </button>
-                    {_.map(util_components_with_defaults)}
+                  <div
+                    style={{ display: "flex", justifyContent: "space-between" }}
+                  >
+                    <div>
+                      {show_pagination_controls && (
```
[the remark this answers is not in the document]
GarvinH: Whoops. 

Yea I need the div since it keeps the rest of the displaytable controls on the right like it was before

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -18,6 +18,8 @@ import {
   DisplayTableCopyCsv,
   DisplayTableDownloadCsv,
   DisplayTableColumnToggle,
+  PageSelector,
+  SelectPageSize,
```
Stephen-ONeil, line 22: Guess, for consistency, either
  - `PageSelector` should be `SelectPage` or
  - `SelectPageSize` should be `PageSizeSelector`
  
I lean towards `Select...`, but I don't think we have a hard rule on how to name React components so either would be fine. Your choice.

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -266,6 +282,31 @@ export class DisplayTable extends React.Component {
...
+    const number_of_pages = paginated_data.length;
+
+    const change_page_size = (new_page_size) => {
+      this.setState({ page_size: new_page_size, current_page: 0 });
+    };
+
+    const change_page = (page) => {
+      this.setState({ current_page: page });
+      this.first_data_ref.current.focus();
+    };
```
Stephen-ONeil, line 299: General React best practice, functions used as DOM event callbacks shouldn't be declared inside the render, should be class methods.

In JS, functions can't be compared by content, only reference, e.g.
```
const a = () => {};
const b = () => {};
const c = a;

a === b; // false
a === c; // true
```

So, if DOM event callbacks are re-declared every render, React has to assume they've changed and will need to touch the DOM to update them on every render. Not huge, but something worth keeping in mind!

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -490,7 +556,11 @@ export class SmartDisplayTable extends React.Component {
       })
     );
     return (
-      <DisplayTable {...this.props} column_configs={smart_column_configs} />
+      <DisplayTable
+        {...this.props}
+        column_configs={smart_column_configs}
+        show_pagination_controls={_.size(data) > 100}
```
Stephen-ONeil, line 562: Small DRY point, could be `> DisplayTable.defaultProps.page_size_increment`

Stephen-ONeil (commented):
Some more comments in passing, I'll wait till those design comments are done/argued with (haha) before doing another full-pass

client/src/components/DisplayTable/DisplayTable.js
GarvinH: TIL 😮🙏

GarvinH:
When the page size is large (ex: 500) and switching pages, there's a chance that the focus on the first item will not work...

Gonna try to figure this out

GarvinH:
Okay, I'm really not sure why it's so inconsistent but, I'm gonna step aside for the time being.

How to replicate what I'm doing: 

1. use table `Program Expenditure by Authority Type" in RPB
2. set page size to 500
3. navigate to a different page using keyboard (if for whatever reason it manages to focus, refresh and try again, it should happen really rarely)

What should happen:
changing pages will focus on the first element of the table post render.

What happens right now:
If a page selector button is clicked with mouse, it focuses 99%(?) of the time (not a hard fix, just add a 10ms debounce on the current debounce should do the trick(?))
However, using keyboard to click, it will only focus about 5%(?) of the time

What I've tried doing:
- Extending the debounce to 500ms. => it's not very consistent
- Add more callback layers following set state
- Use a wrapper on the table content to check if page changed (and should focus)

Alternatively, if the focus on first item feature isn't needed, we could probably scrap it and do a final review

GarvinH:
Issue also in safari and FFox, don't know what's going on with FFox table accessibility though... That's a whole other thing on its own

GarvinH:
Apparently, only I'm having the issue with the focus - ignore whatever i wrote haha

GarvinH:
Just for reference in case someone else runs into this issue as well later on

https://user-images.githubusercontent.com/25855114/112180071-99d17980-8bd1-11eb-8a38-47d86db32585.mov

client/src/components/DisplayTable/DisplayTableUtils.js
```diff
@@ -62,3 +63,204 @@ export const DisplayTableDownloadCsv = ({ csv_string, table_name }) => (
...
+        ...option_range,
+        ..._.range(_.last(option_range) + 1, page_count + 1),
+      ];
+    } else {
+      return option_range;
+    }
+  };
+  const get_page_options = (page_count, current_page, option_window = 2) => {
+    const raw_page_options = _.chain(
+      //without ellipsis
```
ktw1016, line 133: I think due to the nature of this, it's really hard to follow the code for all the different cases. Would you mind adding very short comments explaining what that line is doing, for some of the code? Like you did here `without ellipsis`

client/src/components/DisplayTable/DisplayTableUtils.js
```diff
@@ -62,3 +63,204 @@ export const DisplayTableDownloadCsv = ({ csv_string, table_name }) => (
...
+    .value();
+
+  const dropdown_content = (
+    <form className="paginate_by_dropdown">
+      <div style={{ marginBottom: 10 }}>
+        {_.map(options, ({ label, value }) => (
+          <div key={value} onClick={() => on_select(value)}>
+            <input
+              type={"radio"}
+              name={"paginate_by_dropdown"}
```
ktw1016, line 246: So radio buttons are grouped with `name` I believe, so if there were two paginated tables on the same page, I believe this radio button won't work. I think just giving each instances a unique name should fix it

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -84,35 +54,73 @@ export class DisplayTable extends React.Component {
   constructor(props) {
     super(props);
 
-    this.state = get_default_state_from_props(props);
-  }
-  static getDerivedStateFromProps(nextProps, prevState) {
```
ktw1016, line 89: Looks like this code is from Stephen on covid branch. @Stephen-ONeil can these go away like this?

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -131,16 +139,28 @@ export class DisplayTable extends React.Component {
...
+          sort_func: (a, b) => ... <- (function) Custom sort func. Default to _.sortBy
           sum_initial_value: 0 <- (number) Default to 0
           visibility_toggleable: true <- (boolean) Defaults to false for index 0, true for all other indexes.
         },
       }
       */,
+      page_size_increment,
       util_components,
+      show_pagination_controls,
+      disable_pagination,
```
ktw1016, line 151: Hmm I think `show_pagination_controls` and `disable_pagination` is redundant. I think we can just remove `disable_pagination`, and rename `show_pagination_controls` to something like `enable_pagination`?

Because it doesn't really make any difference to show pagination ctrls with disabled pagination, and also vica versa.

ktw1016 (commented):
Just had some pretty minor comments !

client/src/components/DisplayTable/DisplayTableUtils.js
GarvinH: So I just gave it a quick test, it seems like everything continues to work fine

image


Should I add a name tag anyways?

client/src/components/DisplayTable/DisplayTable.js
Stephen-ONeil: Yeah, this change was important. If a component derives a bunch of state from props in the constructor only, then it's not going to respond to prop changes. There's some case added in the covid panels where `DisplayTable` props can be changed post-mount, so this regression breaks those at least.

GarvinH: I was mixing this up with the `derived_menu_state` from the `Typeahead` 😂🤦‍♂️

Thanks for pointing it out!

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -113,6 +122,37 @@ export class DisplayTable extends React.Component {
...
+      {
+        loading: true,
+      },
+      _.debounce(() =>
+        this.setState(
+          {
+            current_page: page,
+            loading: false,
+          },
+          _.debounce(() => this.first_data_ref.current.focus()) //for some reason if i don't debounce it by 0ms, it will sometimes execute without a proper focus
```
Stephen-ONeil, line 151: Yeah, React will call your `setState` callback before rendering but I believe React calls these callbacks asynchronously (so that it doesn't have to block rendering). Microtiming dependent, I think the callback could be executed either before OR after the resulting render is dome. When it happens before the render finishes, your focus probably went on to an element that was immediately removed from the DOM by the render.

Putting the callback itself in a debounce (indirect setTimeout), even with 0ms, means it's shunted back in to the task queue a second time. If the timing was close enough that it it was sometimes called before and sometimes after the render finished to start with, then it'd make sense that bumping it back in to the queue a second time would be good enough to let the render finish first (I don't know, maybe if you tried enough times you'd still not have the desired behaviour, it's out of your hands when messing with async like this). 

Read up on the JS event loop for a better understanding of how your code's being executed. If you don't find any explanations that click with you I can try and explain it my self some time. 

Anyway, as I said above, this isn't a good use case for `setState` callbacks (my bias perhaps, but I think there's almost no good use cases for them). I've once or twice had to use 0ms timeouts to try and trick the JS engine in to executing how I wanted, but it's kind of an absolute last-ditch hack. In a case like this where it's possible to make the order of events linear with a little bit of extra state, always go for that.

Stephen-ONeil (changes requested):
Some review to start, will have more later.

client/.storybook/main.js
```diff
@@ -23,6 +32,17 @@ module.exports = {
...
+    //Remove all existing css loaders provided by storybook
+    config.module.rules = config.module.rules.filter(
+      (f) => f.test.toString() !== "/\\.css$/"
+    );
+
+    config.module.rules.push({
+      test: /\.css$/,
+      loaders: ["style-loader", "css-loader", "postcss-loader"],
+      include: std_lib_path.resolve(__dirname, "./"),
+    });
```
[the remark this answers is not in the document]
GarvinH: I had to add the css rules because the leaf spinner (using spin.js) would otherwise throw an error if PostCSS didn't exist

client/src/components/DisplayTable/DisplayTable.js
GarvinH: Yea I initially also didn't want to use the `setState` callbacks either but for some reason it was the only way for me to be able to almost consistently have code run after the dom updates instead of the virtual dom.

I mean, OTOH it might just be my mac that has this issue since it didn't seem like anyone else had it. I can give `componentDidUpdate` a try again and see if it works for others

Stephen-ONeil: One of the official use cases for `componentDidUpdate` is inspecting and operating on the DOM, it's not supposed to be called until after the actual DOM has been updated 🤔

client/.storybook/main.js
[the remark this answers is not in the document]
Stephen-ONeil: What was the error?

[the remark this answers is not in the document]
GarvinH: Can't remember, but I remember it had trouble parsing one of the files in the spin.js package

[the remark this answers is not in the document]
Stephen-ONeil: Do you have some idea why that would occur in the storybook but not in the main build?

At a minimum, this needs an explanatory comment. It seems hacky that the storybook needs this loader while the client doesn't appear to use it. Even if you can't find an alternate solution, please try and understand (and document) more of the underlying problem here if possible.

I had a third (and maybe fourth) point I was going to make, but while thinking about it I may have noticed what was actually wrong my self. I'm going to play around with this a little locally and get back to you.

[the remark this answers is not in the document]
GarvinH: 👍 Will leave a comment

I remember reading something about storybook having its own css loader and won't use anything else unless specified or something along those lines. In which case the client is probably adding postcss loader on its own when plugging in spin.js.

ktw1016 (commented):
It looks good to me, tried with VoiceOver and all the focuses were working correctly for me. 

Just one minor thing with Voiceover for `ellipses` button.
Right now it treats it as just "next page" button. It should read out that the button is for next page and also that it'll expand the page window. Also, I'm thinking maybe focus should switch to the page number, instead of the first cell of the table content but probably not necessary🤷‍♂️

client/.storybook/main.js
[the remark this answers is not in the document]
Stephen-ONeil: See #985, fixed there.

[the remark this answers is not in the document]
Stephen-ONeil: Webpack's pretty arcane, and this was extra obtuse because I'd been slamming together our own rules with their defaults, and hadn't really thought too much about that because it was all working initially. The error was fairly misleading too, especially because we had a vestigial import for PostCSS in our package.json, from however long ago we used that. So, this definitely wasn't an easy one.

Still, and this is relevant to the use of debounce elsewhere in this PR, if you don't know _why_ a fix works, or what was actually broken in the first place, then it's really just a hack (no matter how clean or simple the "fix" itself might seem). Sometimes an error message is from a symptom, not from the actual disease, and treating the symptoms won't actually get us anywhere longterm. Sometimes you just need to go do a lot of googling/dig around in the actual source code of the library you're having problems with/read about how the JS event loop works 🤷 .

Or, again, at the very least slap a big comment on the "fix" to make clear that it's just a monkey patch, and that the real problem needs more investigation (comments are basically hack alerts, ha). Luckily my PR there should sort this one out.

Case in point, although it was a lot more subtle, we wouldn't have run in to this now if I'd bothered to roll up my sleeves and dig in to storybook's GitHub to find their default storybook config in the first place, ha. I must have had to much confidence in my own understanding of what my fix was accomplishing back when I fully merged their rules with ours.

Stephen-ONeil (changes requested):
Here are some misc smaller items I've been sitting on, need to pick through `SelectPage` still but don't want to keep holding these up in the mean time.

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -113,6 +122,42 @@ export class DisplayTable extends React.Component {
...
+    this.state.current_page !== page &&
+    this.setState({
+      loading: true,
+      current_page: page,
+    });
+
+  debounced_stop_loading = _.debounce(
+    () => this.setState({ loading: false }),
+    100
+  );
```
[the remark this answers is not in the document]
GarvinH: Without the debounce, the state update is too fast (for the browser?) haha. Without it, it straight up just doesn't show the spinner

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -113,6 +122,42 @@ export class DisplayTable extends React.Component {
     }
   }
 
+  change_page_size = (new_page_size) =>
+    this.state.page_size !== new_page_size &&
+    this.setState({
+      loading: true,
```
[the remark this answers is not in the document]
GarvinH: Is the current name okay?

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -113,6 +122,37 @@ export class DisplayTable extends React.Component {
...
+
+  debounced_stop_loading = _.debounce(
+    () => this.setState({ show_pagination_load_spinner: false }),
+    1
+  );
+
+  componentDidUpdate(prev_props, prev_state) {
+    const { show_pagination_load_spinner } = this.state;
+    if (show_pagination_load_spinner) {
+      this.debounced_stop_loading();
```
Stephen-ONeil, line 148: Debounce is still the wrong tool here, it's the time out that's relevant, right?

Add a comment in `componentDidUpdate` on why we're using this hackyish pattern (both in general why we need a loading spinner on pagination change and in particular why it needs to be bumped back in the event queue with a 1ms timeout).

client/src/components/DisplayTable/DisplayTableUtils.js
```diff
@@ -62,3 +63,229 @@ export const DisplayTableDownloadCsv = ({ csv_string, table_name }) => (
...
   </Fragment>
 );
+
+const _is_ellipsis = (page_options, page_num, index) =>
+  index !== 0 &&
+  index !== page_options.length - 1 &&
+  (page_options[index + 1] !== page_num + 1 ||
+    page_options[index - 1] !== page_num - 1);
+
+export const SelectPage = ({
```
Stephen-ONeil, line 73: To be honest, I'd want to try and clean this up a lot more, just so future devs will have an easier time maintaining it... but I've let this PR get held up too long already (sorry about that) and at least this is relatively isolated code. Not likely to need maintenance unless someone wants to change the behaviour, in which case they can just rewrite it rather than try and modify it, haha.

Maybe just to indicate to future devs that they don't need to try too hard to understand it unless they really need to, add a comment like `TODO cleanup or refactor`. I'd be fine with merging it in after that.

client/src/components/DisplayTable/DisplayTableUtils.js
Stephen-ONeil: Probably currently safe because the pagination options menu closes on click-out. Can't open two different table's menus at the same time. If that menu closing behaviour changed though, this would likely become a bug. Worth being safe/not leaving land mine bugs around, I'd say make them unique (see lodash's `_.uniqueId` util).

Stephen-ONeil (changes requested):
Couple small comments, I think we can move along and merge after this 👍

Stephen-ONeil (approved):
👍  🎉

PR #922, opened 2021-03-01
increase node memory limit to 4gb

ktw1016:


Stephen-ONeil (approved):
👍 

Duh, about the npm argument pass through syntax mix up

PR #928, opened 2021-03-03
React virtualized Typeahead

### LINKED ISSUE #881: Typeahead use react-window/virtualized
related #815

GarvinH:
closes #881

### DIFF
--- a/client/src/components/AutoHeightVirtualList.js
+++ b/client/src/components/AutoHeightVirtualList.js
@@ -1,0 +1,49 @@
+import PropTypes from "prop-types";
+import React, { Component } from "react";
+import { List } from "react-virtualized";
+
+export class AutoHeightVirtualList extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.max_height = props.max_height || 400;
+
+    this.state = {
+      list_height: this.max_height,
+    };
+  }
+
+  componentDidUpdate(prev_props, prev_state) {
+    const { list_ref } = this.props;
+    if (list_ref && list_ref.current) {
+      list_ref.current.Grid.measureAllCells();
+      const list_height = list_ref.current.Grid.getTotalRowsHeight();
+      if (list_height !== this.state.list_height) {
+        if (list_height < this.max_height) {
+          this.setState({ list_height: list_height });
+        } else if (this.state.list_height < this.max_height) {
+          this.setState({ list_height: this.max_height });
+        }
+      }
+    }
+  }
+
+  render() {
+    return (
+      <List
+        {...{
+          ...this.props,
+          ref: this.props.list_ref,
+          height: this.state.list_height,
+        }}
+      />
+    );
+  }
+}
+
+AutoHeightVirtualList.propTypes = {
+  list_ref: PropTypes.oneOfType([
+    PropTypes.func,
+    PropTypes.shape({ current: PropTypes.instanceOf(Component) }),
+  ]).isRequired,
+};
--- a/client/src/components/AutoHeightVirtualList.js
+++ b/client/src/components/AutoHeightVirtualList.js
@@ -1,4 +1,4 @@
-import React, { Component } from "react";
+import React from "react";
 import { List } from "react-virtualized";
 
 export class AutoHeightVirtualList extends React.Component {
--- a/client/src/components/AutoHeightVirtualList.js
+++ b/client/src/components/AutoHeightVirtualList.js
@@ -5,7 +5,7 @@ export class AutoHeightVirtualList extends React.Component {
   constructor(props) {
     super(props);
 
-    this.max_height = props.max_height || 400;
+    this.max_height = props.max_height;
     this.list_ref = this.props.list_ref || React.createRef();
 
     this.state = {
@@ -14,6 +14,12 @@ export class AutoHeightVirtualList extends React.Component {
   }
 
   componentDidUpdate(prev_props, prev_state) {
+    const { max_height } = this.props;
+
+    if (max_height !== this.max_height) {
+      this.max_height = max_height;
+    }
+
     if (this.list_ref && this.list_ref.current) {
       this.list_ref.current.Grid.measureAllCells();
       const list_height = this.list_ref.current.Grid.getTotalRowsHeight();
@@ -39,3 +45,7 @@ export class AutoHeightVirtualList extends React.Component {
     );
   }
 }
+
+AutoHeightVirtualList.defaultProps = {
+  max_height: 400,
+};
--- a/client/src/components/AutoHeightVirtualList.js
+++ b/client/src/components/AutoHeightVirtualList.js
@@ -5,29 +5,24 @@ export class AutoHeightVirtualList extends React.Component {
   constructor(props) {
     super(props);
 
-    this.max_height = props.max_height;
     this.list_ref = this.props.list_ref || React.createRef();
 
     this.state = {
-      list_height: this.max_height,
+      list_height: props.max_height,
     };
   }
 
   componentDidUpdate(prev_props, prev_state) {
     const { max_height } = this.props;
 
-    if (max_height !== this.max_height) {
-      this.max_height = max_height;
-    }
-
     if (this.list_ref && this.list_ref.current) {
       this.list_ref.current.Grid.measureAllCells();
       const list_height = this.list_ref.current.Grid.getTotalRowsHeight();
       if (list_height !== this.state.list_height) {
-        if (list_height < this.max_height) {
+        if (list_height < max_height) {
           this.setState({ list_height: list_height });
-        } else if (this.state.list_height < this.max_height) {
-          this.setState({ list_height: this.max_height });
+        } else if (this.state.list_height < max_height) {
+          this.setState({ list_height: max_height });
         }
       }
     }
--- a/client/src/components/AutoHeightVirtualList.js
+++ b/client/src/components/AutoHeightVirtualList.js
@@ -1,4 +1,3 @@
-import PropTypes from "prop-types";
 import React, { Component } from "react";
 import { List } from "react-virtualized";
 
@@ -7,6 +6,7 @@ export class AutoHeightVirtualList extends React.Component {
     super(props);
 
     this.max_height = props.max_height || 400;
+    this.list_ref = this.props.list_ref || React.createRef();
 
     this.state = {
       list_height: this.max_height,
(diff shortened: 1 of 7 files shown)


ktw1016:
Screen Shot 2021-03-04 at 3 04 25 PM

Are you able to cut this text off, instead of making it a 2 line?

GarvinH:
wait im so confused, im looking at the dev link and its showing me this 
image


I'll see if I can make it a little more consistent

Stephen-ONeil:
Results can be made to overlap by resizing the window down while a search is active, easiest to see on the glossary. 
image

client/src/search/Typeahead/Typeahead.js
```diff
@@ -40,60 +53,69 @@ export class Typeahead extends React.Component {
...
+        current_search_configs !== prev_search_configs
+      ) {
+        cache.clearAll();
+      }
+
+      if (
+        selection_cursor !== prev_selection_cursor &&
+        prev_selection_cursor > selection_cursor
+      ) {
+        this.list_ref.current.recomputeRowHeights(selection_cursor); //scrolling up is choppy if we don't do this
```
ktw1016, line 118: Can you quickly explain what's happening from line 94 to here?

client/src/search/Typeahead/Typeahead.js
```diff
@@ -109,23 +131,66 @@ export class Typeahead extends React.Component {
...
+        .groupBy("config_group_index")
+        .flatMap((results, group_index) =>
+          _.map(results, (result, index) => ({
+            is_first_in_group: index === 0,
+            group_index,
+            result,
+          }))
+        )
+        .flatMap(({ is_first_in_group, group_index, result }, result_index) => {
+          const adjusted_result_index = result_index;
```
ktw1016, line 166: Is there a reason for this variable?

client/src/search/Typeahead/Typeahead.js
```diff
@@ -40,60 +53,69 @@ export class Typeahead extends React.Component {
...
+    if (this.list_ref.current) {
+      this.list_ref.current.Grid.measureAllCells();
+      const list_height = this.list_ref.current.Grid.getTotalRowsHeight();
+      if (list_height !== this.state.list_height) {
+        if (list_height < 400) {
+          this.setState({ list_height: list_height });
+        } else if (this.state.list_height < 400) {
+          this.setState({ list_height: 400 });
+        }
+      }
```
Stephen-ONeil, line 105: Ouch, played around a little trying to keep the `height` prop static by moving the visible styles from `typeahead__dropdown` to the virtualized list items... but that creates awkward UX problems where parts of the UI that seem visible are actually covered by the virtualization window still.

Seems this is the best fix! I'd recommend offloading this state/update logic to a wrapped version of the virtualized `List` though. Abstract the weirder details of virtualization away from the `Typeahead`. Something like `VirtualizedListWithMaxHeight` (feel free to try and think of a better name!), would straight pass through all its props to the wrapped virtualized `List`, but it would take a `max_height` instead of a `height` and would do this state management internally.

... well, refs would be a small gotcha with something like that, but I think that would work out if the constructor did something like `this.ref = this.props.ref || React.createRef();`

ktw1016 (commented):
I checked over indexing stuff and it seems pretty good, Stephen would prob wanna take a look over tho, but looks pretty good to me !

Stephen-ONeil (changes requested):
- [x] pretty broad idea, just something to consider and try if you think it make sense, but a lot of the more troublesome book keeping in `derived_menu_state`/`selection_index` related logic could potentially be done away with if we just stopped giving them their own `li`s inside the virtualized `List`, right?
  1) since the status count item's always there, maybe move the `typeahead__dropdown` class up a level from the `List` to a div wrapping both the status count header and all the virtualized `List` related stuff
  2) put the group headings in the same `li` as the first item in the group (fiddly style wise, styling might need to move from the `li`'s to their children, but totally doable I think)
  3) unless I'm missing something, selection index is now one-to-one with selectable items in the search results (with the additional state corresponding to the input having focus)
 
- [x] weird little box at the top of the dropdown. Double check this, but I think it's just caused by some leftover padding and border styles on `.typeahead__dropdown`
image

client/src/search/Typeahead/Typeahead.js
GarvinH: Yea, so if we treat this as 3 separate sections, we have a height manager, and 2 cell-height refreshers.

### Height manager
For this section (lines 97 - 105), we are checking to see what the height of the list should be. A hard value of height has to be passed into `List` otherwise it will render with height 0. Thus we check if the height of all cells combined will be greater than 400px, and if not, we have to change the height we pass into `List`. The value 400 is taken from the typeahead max height in master.

###cache.clearAll()
Height for each cell of `List` is cached to prevent unnecessary recalculations (assuming height doesn't change). This was implemented since changing search configs were messing with the heights.

### Recompute height (refer to issue 610 of react-virutalized)
There's a known issue that causes jumping when scrolling from the bottom to the top. This was added so that on pressing the up arrow, it would fix the stuttering. Oddly enough, regular scrolling doesn't trigger the jumping issue for us though?

We'll want to remove that last bit if it ever gets fixed

Hope that clarifies it!

client/src/search/Typeahead/Typeahead.js
```diff
@@ -17,21 +25,26 @@ import "./Typeahead.scss";
 
 const { text_maker, TM } = create_text_maker_component(text);
 
+const cache = new CellMeasurerCache({
+  fixedWidth: true,
```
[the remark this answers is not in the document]
GarvinH: Hmm... I wonder if that would've fixed it too now that it's being brought up again.

But I don't think it would've been worth touching anyways because warnings are thrown if there isn't a fixed width or height set (and its advised against in the documentation): 

image

client/src/search/Typeahead/Typeahead.js
```diff
@@ -169,120 +223,56 @@ export class Typeahead extends React.Component {
...
+                      <CellMeasurer
+                        cache={virtualized_cell_measure_cache}
+                        columnIndex={0}
+                        key={key}
+                        parent={parent}
+                        rowIndex={index}
+                      >
+                        <div style={style}>
+                          {_.isEmpty(matching_results) && (
+                            <li className="typeahead__header">
```
Stephen-ONeil, line 263: For consistency, since everything else is just divs now, should switch to div here too

client/src/search/Typeahead/Typeahead.js
```diff
@@ -332,74 +322,26 @@ export class Typeahead extends React.Component {
   }
 
   get derived_menu_state() {
```
Stephen-ONeil, line 324: `derived_menu_state` doesn't really feel necessary anymore (yay! things are much cleaner now), maybe we don't need it?

client/src/search/Typeahead/Typeahead.scss
```diff
@@ -40,12 +40,12 @@
   overflow-y: auto;
   margin-bottom: 2px;
   margin-top: 12px;
-  padding: 8px 0;
   background-color: $backgroundColor;
   border: 1px solid $separatorColor;
   border: 1px solid rgba(0, 0, 0, 0.2);
   border-radius: $standardBorderRadius;
   box-shadow: 0 5px 10px rgba(0,0,0,.2);
+  list-style-type: none;
```
Stephen-ONeil, line 48: No lists anymore, this still necessary?

client/src/search/Typeahead/Typeahead.js
```diff
@@ -332,74 +322,26 @@ export class Typeahead extends React.Component {
...
-    const total_menu_items = (() => {
-      if (needs_pagination_up_control && needs_pagination_down_control) {
-        return results_on_page.length + 2;
-      } else if (needs_pagination_up_control || needs_pagination_down_control) {
-        return results_on_page.length + 1;
-      } else {
-        return results_on_page.length;
-      }
-    })();
+    const total_menu_items = matching_results.length;
```
Stephen-ONeil, line 329: There aren't any non-result menu items, would think `total_menu_items === total_matching_results`. Actually, the flatten was back from when `matching_results` was `matching_results_by_page`, so I'm pretty sure these ARE equal.

client/src/search/Typeahead/Typeahead.js
Stephen-ONeil: Especially `total_matching_results === total_menu_items`, anywhere using either of these can just use `matching_results.length` now 🎉

Stephen-ONeil (changes requested):
Small review notes, couple of bugs. Pretty close now though, and an a huge improvement! Both to the code and the component, I think.

Bugs:
  - [x] crashes if you change anything in the options dropdown while a search is active
  - [x] crashes if you make a search with no matching results

client/src/components/AutoHeightVirtualList.js
```diff
@@ -0,0 +1,41 @@
+import React from "react";
+import { List } from "react-virtualized";
+
+export class AutoHeightVirtualList extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.max_height = props.max_height || 400;
```
Stephen-ONeil, line 8: Ah, missed this. Would be better as 
```
AutoHeightVirtualList.defaultProps = {
  max_height: 400,
}
```
and checking the current props `max_height` value in the `componentDidUpdate`.

As is, this component won't respond to changes in the `max_height` prop (not currently done, but could imagine it being done in a future use case).

Stephen-ONeil (changes requested):
Hmm, curious why you're using `_.size`. I believe in all these cases we're just dealing with arrays, right? To me, `_.size` implies we're dealing with objects, if I'm right in thinking these are always arrays then just using `.length` is a little clearer.

(Well, could also use `_.size` if you input could be of arbitrary type, and it's useful when writing chains, but I still stand by my main point)

client/src/components/AutoHeightVirtualList.js
```diff
@@ -0,0 +1,51 @@
...
+    this.state = {
+      list_height: this.max_height,
+    };
+  }
+
+  componentDidUpdate(prev_props, prev_state) {
+    const { max_height } = this.props;
+
+    if (max_height !== this.max_height) {
+      this.max_height = max_height;
```
Stephen-ONeil, line 20: I don't see where `this.max_height` is necessary, am I pissing something? Can't you just always use the prop value?

Stephen-ONeil (approved):
Approved, pending that one last comment I left above! If there's a specific reason please explain it in a comment, otherwise please drop it. Good to merge after that 🎉

client/src/search/Typeahead/Typeahead.yaml
```diff
@@ -44,6 +23,6 @@ selected_result:
 menu_with_results_status:
   transform: [handlebars]
   en: |
-    {{total_matching_results}} results found. Showing results {{page_range_start}} to {{page_range_end}} of {{total_matching_results}}.
+    {{total_matching_results}} results found.
   fr: |
-    {{total_matching_results}} résultats trouvés. Affichage des résultats {{page_range_start}} à {{page_range_end}} sur {{total_matching_results}}.
+    {{total_matching_results}} résultats trouvés.
```
Stephen-ONeil, line 28: One other super minor thing, haha, something about the period at the end of this bugged me. That just me?

PR #935, opened 2021-03-05
changed local_mongod.sh to work better with WSL2

ChetanMandur:
While experimenting on getting the project up and running using WSL2, I noticed that the `local_mongod.sh` script was causing syntax errors. It seemed to work just fine on my MacBook, but let me know if there are any issues with this.

### DIFF
--- a/scripts/dev_scripts/local_mongod.sh
+++ b/scripts/dev_scripts/local_mongod.sh
@@ -11,18 +11,18 @@ while getopts ":p:" opt; do
       ;;
   esac
 done
-if [[ -z "$port" ]]; then
+if [ -z "$port" ]; then
    port="27017"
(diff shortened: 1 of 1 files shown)


ktw1016 (approved):
👍

Stephen-ONeil:
@ChetanMandur put simply, `[...]` is the standard form (just a shorthand for the unix `test` utility, check `man test`) and `[[...]]` is the extended form (more test flags, slightly different and expanded parsing). Support for `[[...]]` is shell dependant, so this likely means the WSL either doesn't have bash or has a weird version of bash. This specific jam up is going to be a problem in a lot of scripts (although, probably mostly deployment scripts), and there might be other quirks if the shell differs so much from standard bash. Keep an eye out for weirdness!

PR #937, opened 2021-03-05
Change bonus footer to inject during runtime

### LINKED ISSUE #60: About time we clean up the index.hbs.html files and associated code
We tend to put this off constantly, because it's painful to deploy these specific files through Comms, but there's at least some stuff in there that's stale.

For example:
  - [ ] the footer_list content is a mix of hardcoded footer links from the template and run-time footer links from InsertRuntimeFooters. Should get rid of the template links and do them all at runtime, because that's a lot more flexible for us
  - [ ] ... the id of that footer list area is `#footer_survey_link_area` but we've now moved the survey link out of it (should have given it a more generic name in the first place, it stoped being just for the footer link very quickly). Annoying to leave it with the misleading id, but not worth pushing the html files through comms just for this

GarvinH:
closes #60

### DIFF
--- a/client/src/InfoBase/index.hbs.html
+++ b/client/src/InfoBase/index.hbs.html
@@ -107,7 +107,6 @@
             <nav class="col-md-10 ftr-urlt-lnk" aria-labelledby="footer-nav">
               <h2 id="footer-nav" class="wb-inv">{{footer_nav}}</h2>
               <ul id="footer_main_link_area">
-                {{{footer_list}}}
               </ul>
             </nav>
             <div class="col-xs-6 visible-sm visible-xs tofpg"></div>
--- a/client/src/InfoBase/index.hbs.html
+++ b/client/src/InfoBase/index.hbs.html
@@ -97,8 +97,7 @@
       <div class="bonus-footer-landscape">
         <nav class="container wb-navcurr" aria-labelledby="canada-ca-footer-nav">
           <h2 id="canada-ca-footer-nav" class="wb-inv">{{canada_ca_footer_nav}}</h2>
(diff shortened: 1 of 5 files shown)


GarvinH:
I'm thinking... a default CRA (create-react-app) app only starts as a `<body>` and `<div>`... Maybe we should do the same? Doing so would give us full control over the structure of the website

what i mean is instead of our now 
```
<body>
  <header>
  <main>
    <div id=app>
  <footer>
```

we can do it as

```
<body>
 <div id=app>
    <header (jsx)>
    <main (jsx)>
    <footer (jsx)>
```

Alternatively maybe we can set up multiple entry points???

GarvinH:
Also with injecting footers at runtime, what should we do about the noscript page since we lose the minks?

Stephen-ONeil:
Hold up now, there are benefits to serving more than a blank page in the index html files! In fact, if they weren't an operational headache to update, we'd probably want to move a little more in to them.

We can use them to optimize for getting content on screen. Anything that requires React to render has to wait for a whole lot of loading, parsing, and executing (that's why there's a non-react version of the loading spinner, so that we can get one up ASAP and provide cover for React to load and the app to initialize). If we pulled as much as possible out of the index html files, then we'd have a mostly blank screen aside from the spinner, and then a whole lot of pop-in all at once when the app's ready to go. The content in the index files might not be useful (or even usable? the initial spinner might block interaction, don't remember) but it at least gives the user something to look at.

Now, if you want to totally replace the index html text templates with React, and then update `IB_base` to render that react to static html files? Yeah, that sounds good to me!

Issue #931, opened 2021-03-04
Drop explicit @wry/equality dependency (in a month or so)


Stephen-ONeil:
https://github.com/benjamn/wryware/issues/99

@apollo/client requires @wry/equality ^0.3.0. To lock that to an IE11 safe version, I've temporarily added @wry/equality 0.3.2 to our package.json. I've subscribed to the issue about IE11 support on the wry repo so I should see when it's safe to drop the version from our own package.json, this issue's just a reminder to check back up on that in case I miss the alert.

Stephen-ONeil:
Update: fixed in 0.3.4, just to play it safe, have an explicit `"@wry/equality": "^0.3.4"` in `client/package.json`. Come back in a month or so, drop that line, and close this.

PR #938, opened 2021-03-08
Switch @wry/equality from 0.3.2 pin to ^0.3.4 in client package.json

Stephen-ONeil:
@wry/equality transpilation fixed in 0.3.4 via https://github.com/benjamn/wryware/pull/102

Trivial, but waiting for dev link to double check.

- [x] ... hmm, small chance any other open branch with its own set of `client/package.json` changes and corresponding new `package-lock.json` could clobber this and regress to 0.3.3. Would be safest to explicitly pin to `^0.3.4`.

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -7999,9 +7999,9 @@
       }
     },
     "@wry/equality": {
-      "version": "0.3.2",
-      "resolved": "https://registry.npmjs.org/@wry/equality/-/equality-0.3.2.tgz",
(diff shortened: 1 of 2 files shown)


Stephen-ONeil:
Went with the safe, incremental fix. #931 will remain open for now, but once the branches have all rebased and cycled a bit I won't have caching concerns and we can drop the explicit `@wry/equality` requirement altogether and close that issue.

Issue #868, opened 2021-01-13
Review long relative imports, leverage absolute imports


Stephen-ONeil:
Great that we have absolute path imports working now. Lots of imports that look like `../../../...` that we should hunt down and clean up. Even better, modules like `src/panels/panel_declarations/shared.js`, that mostly just import modules to reexport them locally, can _finally_ die!

Stephen-ONeil:
Way I see it, we can think of anything under the directory `X` in `src/X` as being "local". "local" modules could (should?) probably import each other via relative imports. If module `x.js` in `src/X/x.js` wants to import from `src/Z/z.js`, it should probably use an absolute import.

Of course, unless we find a linting plugin for this, it's just an informal rule and will probably never be totally consistent. I also haven't thought too long about it, so any other opinions are, as always, welcome!

AlexCLeduc:
After a manual cleanup, we can add the https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-relative-parent-imports.md rule to ensure this practice stops

ChetanMandur:
Absolute imports have been implemented, making new issue regarding re-export files.

PR #942, opened 2021-03-08
Long (relative) to absolute imports

ChetanMandur:
related #868 

The rule I followed is that if the import isn't 'local', it becomes an absolute import.


ktw1016:
I just had a quick look over and this is great! Do you wanna try adding the eslint rule that Alex linked in the issue? and this PR can close the issue

ktw1016 (approved):
🎉👍

Stephen-ONeil:
One real minor thought, if you want to move this along it could be a follow up issue/PR, but I'm seeing imports like, e.g., `/components`, `/components/`, and `/components/index.js`. My preference is to always explicitly import the `index.js` in any case (for one thing, I believe the auto-ordering rule will treat it differently if it knows it's a JS file).

Stephen-ONeil:
One other note, this could close the related issue if you want, just need to open a follow up issue about breaking up the legacy re-exporting files like `client/src/panels/panel_declarations/shared.js`.

ChetanMandur:
I think breaking up legacy re-export files should be another PR. I'll make another issue once this PR gets merged.

Also, as a side note, does anyone know why the rebase and merge is causing conflicts?

Stephen-ONeil:
> I think breaking up legacy re-export files should be another PR. I'll make another issue once this PR gets merged.

Sounds good to me!

> Also, as a side note, does anyone know why the rebase and merge is causing conflicts?

Haha, it's a little quirk of the GitHub UI. GitHub isn't showing you as having conflicts with master because you have merge commits resolving those conflicts already. A rebase operation doesn't apply merge commits though! So the tricky bit is that GitHub both tells you there are no conflicts (because of the resolutions in the merge commits) but also says you can't rebase automatically because of conflicts (true, but in that case, when the rebase option is selected, they should ignore the merge commits and just tell you where the conflicts are, IMO!).

Anyway, you can either change the dropdown from "Rebase and merge" to "Create a merge commit" on GitHub, or you can rebase off master on your local branch and re-fix the merge conflicts as they arise in the process, the force-push that version of the branch back up to here. Only bother with the local rebase if the merge conflicts weren't too bad.

Either way, in the future, whenever you want to reconcile with upstream changes, please use `git rebase master`, etc, instead of creating merge commits. Just because we've decided to stick with rebasing over merging on this project. It'll keep things more straightforward.

Stephen-ONeil (approved):
👍  🎉 

Awesome! See my early comment on the rebase vs merge question.

PR #943, opened 2021-03-10
Paginated DisplayTable review

Stephen-ONeil:


client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -19,7 +19,7 @@ import {
   DisplayTableDownloadCsv,
   DisplayTableColumnToggle,
   PageSelector,
-  PageinateBySelector,
+  SelectPageSize,
```
Stephen-ONeil, line 22: Minor point, "page size" seemed generally a more accurate name than "paginate by". Conveys the concept more directly, and avoids the connotation of `___By` with methods (like `countBy`, `groupBy`, etc.)

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -113,15 +113,15 @@ export class DisplayTable extends React.Component {
         },
       }
       */,
+      page_size_increment,
       util_components,
-      size_to_paginate,
```
Stephen-ONeil, line 117: Bit of an unclear purpose, could only see it as being, in practice, used to disable pagination (in which case it could have just been a `disable_pagination` prop).

Stephen-ONeil: ... I didn't add one, but an explicit `disable_pagination` would still be useful, probably.

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -113,15 +113,15 @@ export class DisplayTable extends React.Component {
         },
       }
       */,
+      page_size_increment,
```
Stephen-ONeil, line 116: Small tangent from the specific cleanup feedback part of this, more of an API design point. Felt like a prop that let the consumer specify the page size was appropriate, but since the component itself allows users to change `page_size` this is instead more about setting the default page size and controlling the page size options available within the component.

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -280,61 +280,29 @@ export class DisplayTable extends React.Component {
...
-      page_selector,
-    } = (() => {
-      const paginate_by =
-        size_to_paginate && _.size(sorted_filtered_data) > size_to_paginate
-          ? state_paginate_by
-          : 0;
-
-      const num_pages_exact = paginate_by
-        ? _.size(sorted_filtered_data) / paginate_by
-        : -1;
```
Stephen-ONeil, line 297: Both of these encode future logic operation in their stored values by assigning special meaning to values which would otherwise hold no meaning (page size of 0, or a negative number of pages). I know this kind of thing goes on in the `Typeahead` to some extent, but I very strongly suggest against it! Bad for code legibility and a future-bug waiting to happen, IMO.

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -280,61 +280,29 @@ export class DisplayTable extends React.Component {
...
-
-      const num_pages =
-        num_pages_exact >= 0 && _.isInteger(num_pages_exact) //if we have the exact amount of items to fill an entire page, it will start a new empty page
-          ? num_pages_exact - 1 //thus we want to get rid of the extra page (as it has no content)
-          : _.toInteger(num_pages_exact); //start counting from 0 (-1 from the literal amount of pages)
```
Stephen-ONeil, line 302: Chunk doesn't make empty arrays, this is all about specially massaging the `num_pages` value because `PageSelector` doesn't behave well, right? Fix `PageSelector`'s behaviour, or at least push this logic down in to that component. Otherwise, this is just another time bomb.

client/src/components/DisplayTable/DisplayTableUtils.js
```diff
@@ -157,8 +157,17 @@ export const PageSelector = ({
...
+  selected,
+  on_select,
+  page_size_increment,
+  num_items,
+}) => {
+  const options = _.chain(num_items / page_size_increment)
+    .ceil()
+    .range()
+    .map((_, ix) => (ix + 1) * page_size_increment)
+    .value();
```
Stephen-ONeil, line 170: Could be better, but exactly how was an unanswered design question in the base PR. Do we only want a fixed number of options? An explicit "show all" button?

client/src/components/DisplayTable/DisplayTable.js
```diff
@@ -280,61 +280,29 @@ export class DisplayTable extends React.Component {
...
-          current_page={current_page}
-          change_page={change_page}
-          show_select
-          num_col={_.size(visible_ordered_col_keys)}
-        />
-      );
+    const paginated_data = _.chunk(sorted_filtered_data, page_size);
+
+    const number_of_pages = paginated_data.length;
+    const show_pagination_controls = number_of_pages > 1;
```
Stephen-ONeil, line 286: Small bug I introduced here, can lose the pagination controls if you pick a page size that results in `number_of_pages === 1`. A better idea would probably be to make `show_pagination_controls` independent of state (think I commented this on the base PR as a design point, but I don't think the controls should go away even if a filter is applied that makes them useless). Should probably just be a boolean prop. At most, the "smart" version of the display table to check the initial items against the `page_size` prop and switch `show_pagination_controls` to false if it is unnecessary.

client/src/components/DisplayTable/DisplayTableUtils.js
GarvinH: IMO, 100 - 500 then a show al button would be good. Right now the program expenditures by authority type table (RPB) has 36 page increment sizes which is making the dropdown gigantic haha.

Thoughts?

Stephen-ONeil: Haha, yeah, that's too many options. Sounds good to me, we don't really want multiple large options since switching between those will be slow anyway, maybe only give a couple options less than something reasonable (whatever number of cells most tables start getting chunky at) and then the "all" button.

Something I forgot to leave a comment on but was an intentional part of my changes here. In the base branch, 100 was the hardcoded page size in the `DisplayTable` itself, where it only showed up in one place and, thus, would look like it was trivial to change later. The wrinkle to that was that, in a separate module, the page size selection component was also hardcoded to work in units of 100's. Generally some sort of configuration option like page size should only have a single source of truth (or better yet, an overridable default) at the top layer of the abstraction (the `DisplayTable`), and any internal components (like the utils) should know nothing/make no assumptions about it on their own. 

Another way to put it is that you should try to keep code (and concepts represented inside that code) "perpendicular", meaning only intersecting at a single point, whenever possible. Code-wise, it's easy if you're working with something like a React component, where everything has a clear, single interface and is, by design, composable (... although CSS can create a second, messy, layer over all this). A change in either should, **at most**, require a change at the interface of the other. The harder level is keeping track of concepts/assumptions/knowledge. In this example, they were coupled both by their interface and by their (indirect) shared knowledge of the magic number 100.

Assumptions, for example, tangle things up if they aren't explicitly checked at the interface. This is heavily formalized in strongly typed languages, which generally also take care of checking the assumptions and reminding you of them prior to run-time. Of course, you still need to architect your code and put in the work to define these assumptions in the first place, but at least future devs have more guide rails in place when you've done that.

... a THIRD way you might hear basically the same idea communicated is when someone talks about a "leaky abstraction". Organizing code in modules/methods/components/objects etc is as much about abstraction as it is about reusability. As you probably know, abstraction means representing a more complex concept with a different, but workably equivalent, simple concept. When talking about code organization, that mostly means hiding the implementation details behind a name and interface. An abstraction becomes leaky if it makes (undeclared or unsafe) assumptions about where and how it will be used (generally, this is when we get a bug time bomb), or if too many implementation details surface up through the interface (generally, this is when the abstraction becomes useless as developers need to learn the implementation details anyway; may also couple the abstraction too tightly to the consumers and become brittle/hard to refactor in the future).

PR #944, opened 2021-03-11
Eslint no relative import rule

ktw1016:
The rule linked in the issue wasn't working, so I found a different core eslint rule


ChetanMandur (approved):
Looks good to me! Thank you for finding this!

It looks like by going with this rule, local modules have to import each other using absolute paths instead of relative paths ([which I was trying to avoid](https://github.com/TBS-EACPD/infobase/issues/868#issuecomment-759720413)). But I don't think it's much of an issue doing it this way.

PR #947, opened 2021-03-15
Service inventory with React Query

ktw1016:
Docs: https://react-query.tanstack.com/overview

I'm just trying to think of a more sustainable way to do this long term, to cover when we eventually drop all tables and go fully GQL. I take Service data as an example below but I'm really thinking how it should be we drop everything like subject, footnotes, etc

Right now, I'm using React Query for `services_types` panels only and need to scale it to all panels. Just some thoughts to try out
### 1. Simply copy and paste `useGQLReactQuery` to other panels, trust React Query caching. 
This would be simplest but it would mean separate loading for each panels. Also, there is always gonna be same post processing that will happen which I'm not sure if React Query caches that also. Doing this, we can even drop the models (like class `Service`) but maybe that's too far?

### 3. Hybrid of the above. Stick `useGQLReactQuery` in each panels, but inside `useGQLReactQuery`, use models store Map.
We can definitely cache post processed data we want this way, still gotta take care of loading stages for each panels

- [ ] object mapping for fetching all different types of data for scaling
- [ ] clean up all obsolete code
- [ ] think of better naming, modules organization


ktw1016:
Trying to include DevTools in dev link: https://github.com/tannerlinsley/react-query/discussions/1981

PR #948, opened 2021-03-16
Cleanup typeahead

### LINKED ISSUE #880: Extract search logic from Typeahead

GarvinH:
closes #880

Did we want to have more extracted, or is this okay?

Edit: Will move typeahead to `/components` before merging


Stephen-ONeil:
Oof, sorry this sat un-reviewed for so long. I wanted to go further than this. For example, "config groups" is a concept belonging to our old subject search config setup. If we're going to have a totally generic, reusable, `Typeahead` component inside the `/component` dir, then it shouldn't know anything about the config groups.

Think about the minimum responsibilities for something called a `Typeahead`. It's probably just having a text input (that, debounced, triggers a callback) and being able to render a list of results (with accessible navigation and screen reader status updates, which in our implementation means holding the state for currently selected item). What happens when the search input changes, where the list of results comes from, how items are rendered inside the list, what happens on selection, etc., is all something that should probably be handled by the consumer of the `Typeahead`.

After extracting all that out in to a general `Typeahead` component, we could then have wrapped version of it inside `search` that understands the subject search specific concepts, like search configs. `EverythingSearch`, `GlossarySearch`, etc. would use this subject-search specific wrapped `Typeahead`.

Benefit of all this is the decoupling of the subject-search concepts and the `Typeahead` component. We'd be able to refactor either without touching the other (assuming the API is stable), we could possibly use `Typeahead`s in place of other searches on the site (e.g. to get autocomplete options for filters?), and we'd have a much either time experimenting with alternatives to the subject-search (e.g. we could easily reuse the `Tyepeahead` for something like a server-side search).

Stephen-ONeil:
Let me know if you've got other questions, or if you want to stop here as an incremental step! Could review with that in mind, would just mean not closing the original issue yet.

Stephen-ONeil:
Bunch of different thoughts, to speed up the review process I'll be making a branch off this implementing my suggestions 👍

Stephen-ONeil (changes requested):
See #1014 for "review"

Stephen-ONeil (approved):
👍

PR #955, opened 2021-03-18
Covid multi-year rework

Stephen-ONeil:
TODO:
   - [x] better API structure for multiple years of data
     - hmm, still a bother in places, but better
   - [x] get data loading in panels
   - [x] ah, my `groupBy` pattern for rolling up data is a big source of converting ints to strings, clean that up
   - [x] maybe flatten more server responses in queries themselves? Lots of weird unwrapping and traversal necessary in cases, the consumers shouldn't have to know that much about the GraphQL schema structure and quirks
   - [x] use `calendar_month` column from `covid_expenditures.csv` in place of `COVID_CURRENT_YEAR`
     - ... might need to keep something like `COVID_CURRENT_YEAR` for the text referencing that date in the FAQs? Oof
     - need a formatter for year, month number -> plain text (+ end of month date? Can I get away from specifying that, or just adding "the end of ..." to the static text or something?)
   - [x] maybe the intro DOES need to have year tabs, it's a little weird to just be talking about the relatively small main estimates auth for next year along side the large expenditures from this year (even if this is only the case for a couple months a year)
   - [x] ~should the default tab be the in-year tab, rather than the latest available tab? Would generally keep the estimates and auth panels more in line throughout the year (although it might look like a weird choice with expenditures still disabled)~ well, it's 2021-22 now, so we can wait till next year's mains before asking this question again
   - [x] text...
     - this will be hell, need to try and get cover to rewrite a lot of things, I think
     - [x] get French changes in covid estimates panel confirmed
     
 Ugh, lots of additional text/special cases to clean up/re-think once expenditures turn on for good. Will likely leave most of that out of scope for this PR.


client/src/panels/panel_declarations/covid/covid_expenditures.js
```diff
@@ -399,49 +343,49 @@ class CovidExpendituresPanel extends React.Component {
...
   }
   componentDidMount() {
-    client
-      .query({
-        query: gov_covid_summary_query,
-        variables: {
-          lang: lang,
-          _query_name: "gov_covid_summary_query",
-        },
+    query_gov_covid_summaries().then((covid_summaries) =>
```
ktw1016, line 351: I'm only checking how queries are being done in this PR right now, but I think using hooks is the superior way if querying data panel by panel basis for a few reasons. Mainly, hooks take care of `loading`, as well as state management (since there is no need to do any mostly) and also less boilerplate code. Just my 2 cents after trying out hooks, and taking a super quick look on how queries are being done here

PR #956, opened 2021-03-18
Breaking legacy reexport

### LINKED ISSUE #945: Breaking up legacy re-exporting files
Related to #868 

Now that absolute imports are possible, these files can be broken up.

ChetanMandur:
closes #945


ktw1016:
It seems like you don't have eslint or prettier set up, so Slack me about that!

Have we decided on doing `import * as util_components from "src/components/index.js";` and then destructuring from `util_components`? Why don't we just import it directly? Oh, or was this just for the convenience thing because that's how it was doing it originally?

Also just made me question, should `InfographicPanel` be part of common panel components?

ChetanMandur:
@ktw1016 Yeah, I just chose to import `util_components` the way I did just because that is how it was done originally. I also (think I) saw some other imports from `util_components` being done the same way, so I just followed suit.

As for the `InfographicPanel`, I'm not too sure haha, might need some more insight from others. I personally think it can be kept separated, sort of as its own thing as I can see it as somewhat 'bloating' `common_panel_components`. But again, I wanna hear other's thoughts on it.

ktw1016:
So the only change I'd like to request is to directly import each modules instead of importing * as util components. Other than that, should be good to go !

ktw1016 (approved):
Looks good!🎉

PR #957, opened 2021-03-19
Service inventory apollo hooks

ktw1016:


client/src/models/populate_services.js
```diff
@@ -7,322 +7,160 @@ import { lang } from "src/core/injected_build_constants.js";
...
+  const { id } = query_options;
+  const is_gov = id === "gov";
+  const variables = {
+    lang,
+    id,
+  };
+
+  const query = services_query(query_options);
+  const res = useQuery(query, { variables });
+  const { loading, error, data } = res;
```
AlexCLeduc, line 148: fetchServices is a hook, right? It took me a while to understand that, it should be called something like `useServices` or `useServicesData`

AlexCLeduc:
I'm confused when we use the new hook vs. what is provided by `requires_services`. Is the idea to transition fully to the hook? 

It's not a big deal if we stick to the old API for now, hooks and batching can be implemented later if they're causing issues right now

> Prepare summary data ...

You're talking about the server computing aggregations like # of services by type, right? I thought we already did this? 

If not, I think this is essential for the gov level service panels. Making the client load every single GC service is overkill when just showing counts. We can optionally do the same for departments, the benefit there wouldn't be performance, but maybe consistency with the gov panels, e.g. we can write the aggregation logic once on the server 🤷

ktw1016:
New PR: #1002

PR #958, opened 2021-03-19
Smoothen pinned content scrolling

### LINKED ISSUE #952: "some things to keep in mind" sticky positioning is a little janky
It would be nice if my the panel's Y coordinates didn't shift when the element goes from static to fixed positioning. My guess is that the element being removed from static positioning is shifting the panel upwards. 

If we can't use real sticky positioning, maybe we can leave a copy of the element (or another element with equivalent height)  in static positioning, so nothing else has to move up 

ib-sticky-header-janky

GarvinH:
closes #952 

It's still janky if the user scrolls and open/closes the accordion at the same time (but I wonder how often this would really happen?). With that being said, I mostly found it to be like this only if closing the accordion.

### DIFF
--- a/client/src/components/PinnedContent.js
+++ b/client/src/components/PinnedContent.js
@@ -17,85 +17,107 @@ import text from "./PinnedContent.yaml";
 
 const text_maker = create_text_maker(text);
 
-export const PinnedContent = withRouter(
-  class extends React.Component {
-    constructor(props) {
-      super(props);
+class _PinnedContent extends React.Component {
+  constructor(props) {
+    super(props);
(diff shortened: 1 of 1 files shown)


ktw1016 (approved):
I think this works! Maybe just add a little comment on line 105 to remind how the div is making the scrolling smoother (so contents don't move around when sticked). Also just super minor comment, is `stickied` a word? I think just `sticky` works

GarvinH:
Yea it should be a word. I did a quick google search to double check, there's a link to collins dictionary 👍

PR #960, opened 2021-03-22
Back to top storybook

GarvinH:
Also fixed formatting of glossary on mobile devices (most notably for ipad size devices)


sranc046 (commented):
Looks good! Only thing I would suggest adding is a top border of the footer section in storybook to add visual clarity

sranc046 (approved):

PR #961, opened 2021-03-22
Add docs on adding tmux config file

GarvinH:


ktw1016:
This is actually in the docs! Take a look at step 3

GarvinH:
😂🤦‍♂️

I did a search on `.tmux.conf` and didn't find anything. I can't believe i didn't see it when it was right in front of me

GarvinH:
good catch! thanks

ktw1016:
😂

PR #964, opened 2021-03-22
User focus btt storybook

### LINKED ISSUE #959: Let user of BackToTop component handle focus
Right now the BackToTop component handles focus internally using the `querySelector` function which is in a way restrictive (handling search containing selectors `>` and `[]` properties).

I think we should let the parent component handle the focusing event (say in a function). That way, the focus can be more versatile.

In the case of glossary, it will use a `querySelector`. For the storybook, it can use `getElementById`. It would also allow us to handle using refs later if we wanted to.

ChetanMandur:
closes #959

### DIFF
--- a/client/src/components/BackToTop/BackToTop.stories.js
+++ b/client/src/components/BackToTop/BackToTop.stories.js
@@ -5,7 +5,6 @@ import React, { Fragment } from "react";
 import { BackToTop } from "./BackToTop.js";
 
 function focusBTT () {
-  document.body.scrollTop = document.documentElement.scrollTop = 0;
   document.getElementById("#test").focus();
 }
 
--- a/client/src/components/BackToTop/BackToTop.stories.js
+++ b/client/src/components/BackToTop/BackToTop.stories.js
@@ -4,6 +4,11 @@ import React, { Fragment } from "react";
 
 import { BackToTop } from "./BackToTop.js";
 
+function focusBTT () {
+  document.body.scrollTop = document.documentElement.scrollTop = 0;
+  document.getElementById("#test").focus();
+}
+
 const BackToTopTemplate = ({ focusId }) => (
   <Fragment>
     <div id="ib-site-header-area">Header</div>
@@ -12,7 +17,7 @@ const BackToTopTemplate = ({ focusId }) => (
     <div id="wb-info" style={{ height: "287px" }}>
       Footer
     </div>
-    <BackToTop focus={focusId} />
+    <BackToTop focus={focusBTT} />
   </Fragment>
 );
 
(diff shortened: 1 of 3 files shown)


client/src/components/BackToTop/BackToTop.stories.js
```diff
@@ -4,6 +4,11 @@ import React, { Fragment } from "react";
 
 import { BackToTop } from "./BackToTop.js";
 
+function focusBTT () {
```
GarvinH, line 7: we generally name our variables using `underscore_notation`. So please excuse the functions where it follows camel casing (i need to go back and fix those).
With that said, generally we use camel notation when we're accepting props (but I've only noticed us using camel case notation either when the library we're using uses it, or if I do it out of habit and I forget to go back and fix it

But in this case, I don't think there's much efficiency to be gained by setting a function in memory so we don't really need to rename this function. With that, we can create an anonymous arrow function

```
focus={() => {
  //code here
}}
```

(If you could rename the bottom one i'd appreciate it!)

client/src/components/BackToTop/BackToTop.stories.js
```diff
@@ -4,6 +4,11 @@ import React, { Fragment } from "react";
...
+  document.body.scrollTop = document.documentElement.scrollTop = 0;
```
GarvinH: I think we can keep this in `handleClick()` since we don't really have a use case where we'd want it to scroll only partially up just yet.

Also to clarify, focus is purely for focusing on a certain item within the page after the `BackToTop` button is pressed, so it wouldn't make sense to leave it here anyways haha

client/src/glossary/glossary.js
```diff
@@ -166,6 +166,11 @@ const Glossary_ = ({ active_key, items_by_letter }) => (
   </div>
 );
 
+function focusBTT() {
```
GarvinH, line 169: just a minor thing, i'm not too sure using BTT would exactly be the clearest immediately to new readers of the code, so maybe it would be better to have the full name written? At the same time I feel like the meaning of `focusBTT` can be a bit ambiguous possibly? I was thinking `back_to_top_focus_item` might be a bit clearer but it's kinda long to say haha. Feel free to use something else that is clearer and shorter.

As a side note, in this case, we want to keep this as a defined function (non-anonymous) because in dev/prod, the function isn't gonna be run at every single change on the page, thus we don't want this function to be recreated on every render cycle.

if it wasn't like this, the alternative would be to creating a class method.

client/src/components/BackToTop/BackToTop.stories.js
GarvinH: Forgot to mention that there's not much efficiency to be gained because everything gets re-rendered when the code changes (in storybook) anyways which means that the memory would've been reset anyways

(I think that's how storybook works anyways)

GarvinH (approved):

PR #965, opened 2021-03-22
Storybook a11y

### LINKED ISSUE #927: Storybook a11y
We need to inject `is_a11y_mode` into storybook to test components in their a11y states.. So we may need different ways other than DefinePlugin to inject build constants..

GarvinH:
closes #927 

Didn't manage to find a better way to do this, but yet I still feel like there has to be a better way to handle this...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -171,7 +171,7 @@ jobs:
             - client/build
           key: prod-build-{{ .Branch }}-{{ .Revision }}
       
-      - run: cd client && npm run build-storybook -- -o $BUILD_DIR/InfoBase/storybook
+      - run: cd client && npm run build-storybook -- -o $BUILD_DIR/InfoBase/storybook && npm run build-storybook-a11y -- -o $BUILD_DIR/InfoBase/storybook-a11y
 
       - restore_cache:
           keys:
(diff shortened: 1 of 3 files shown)


ktw1016:
Hmmm yeah I was playing around with this one too and couldn't find a good solution. Honestly, I think it could just be having two separate commands like `storybook` and `storybook-a11y`.. We would also need to do that for the dev link of the storybook

ktw1016:
Was the a11y injected variable working locally? I don't see a11y version of the dev link changed

GarvinH:
Yea it was working locally. I think none of the current stories contain any components that appear differently on the a11y versions.

I'll go ahead and find a component which has a11y and see if it works on dev link

GarvinH:
Typo in the package.json 😂🤦‍♂️. Gonna double check and push

GarvinH:
Okay it should be good now!

ktw1016 (approved):
I'm good with this but let's see what others think of this

Stephen-ONeil (approved):
Hmm, thought through it a bit more. Since we set our constants at build time and import them from a module, this is only addressable as a webpack/build time problem, duh. We'd have to change how `injected_build_constants.js` works to be able to mock it. We'll need a good solution to that to that eventually for unit testing, and it should be something storybook could share, but this is the only incremental solution we have for now.

Stephen-ONeil:
@GarvinH If you merge this and close the current issue, open a new issue about the general handling of the build time constants. We want to (in dev mode only probably) be able to override them. Reference back to the original storybook issue since we'll want to revisit this solution once that exists.

PR #969, opened 2021-03-25
Re-show heading numbers for double counted values

### LINKED ISSUE #968: turn double-counted values back on
who-we-help tags should go back to having amounts, e.g. in the image below, "small and medium enterprises" should have spending/FTE values that are totaled from their children programs. 

image

We can do the same for how-we-help as well..

If we can think of a way to make those number visually more suspicious/less legit, that would be cool, a few ideas we can combine are
- applying an opacity
-  prepending an `<` or `≤` character before the amount (technically these are up-to amounts, but because of tagging limits I suppose it's possible certain tags get left behind)
- surrounding these numbers with parentheses
- rounding the numbers even more to reduce perceived precision

GarvinH:
closes #968

image

### DIFF
--- a/client/src/components/misc_util_components.js
+++ b/client/src/components/misc_util_components.js
@@ -32,22 +32,22 @@ class Format extends React.PureComponent {
       prefix,
     } = this.props;
 
-    const formatted_content = formats[type](content);
-
-    const prefixed_content = prefix
-      ? `<span>${prefix}${formatted_content}</span>`
-      : formatted_content;
-
-    const parenthesis_content = in_parenthesis
-      ? `<span>(${prefixed_content})</span>`
-      : prefixed_content;
+    const formatted_content = _.chain(content)
+      .thru(formats[type])
+      .thru((content) =>
+        prefix ? `<span>${prefix}${content}</span>` : content
+      )
+      .thru((content) =>
+        in_parenthesis ? `<span>(${content})</span>` : content
+      )
+      .value();
 
     return (
       <span
         style={style}
         className={className}
         dangerouslySetInnerHTML={{
-          __html: parenthesis_content,
+          __html: formatted_content,
         }}
       />
     );
--- a/client/src/components/misc_util_components.js
+++ b/client/src/components/misc_util_components.js
@@ -1,7 +1,6 @@
 import _ from "lodash";
 import React, { Fragment } from "react";
 
-
 import {
   run_template,
   trivial_text_maker,
@@ -24,13 +23,17 @@ const ExternalLink = ({ children, href, title }) => (
 
 class Format extends React.PureComponent {
   render() {
-    const { type, content, style, className } = this.props;
+    const { type, content, style, className, in_parenthesis } = this.props;
 
     return (
       <span
         style={style}
         className={className}
-        dangerouslySetInnerHTML={{ __html: formats[type](content) }}
+        dangerouslySetInnerHTML={{
+          __html: !in_parenthesis
+            ? formats[type](content)
+            : `(${formats[type](content)})`,
+        }}
       />
     );
   }
--- a/client/src/components/misc_util_components.js
+++ b/client/src/components/misc_util_components.js
@@ -23,16 +23,31 @@ const ExternalLink = ({ children, href, title }) => (
 
 class Format extends React.PureComponent {
   render() {
-    const { type, content, style, className, in_parenthesis } = this.props;
+    const {
+      type,
+      content,
+      style,
+      className,
+      in_parenthesis,
+      prefix,
+    } = this.props;
+
+    const formatted_content = formats[type](content);
+
+    const prefixed_content = prefix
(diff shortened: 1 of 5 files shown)


client/src/explorer_common/resource_explorer_common.js
```diff
@@ -115,6 +114,16 @@ export const get_col_defs = ({ year }) => [
     get_val: (node) => _.get(node, "data.resources.spending"),
     val_display: (val) =>
       _.isNumber(val) ? <Format type="compact1" content={val} /> : null,
+    first_node_m2m_display: (val) =>
+      _.isNumber(val) ? (
+        <Format
+          type="compact1"
+          content={Math.ceil(val / 10000000) * 10000000}
```
Stephen-ONeil, line 121: The `compact1` formatter applies rounding, why the extra rounding here?

client/src/explorer_common/explorer_components.js
```diff
@@ -167,24 +167,34 @@ export const ExplorerNode = ({
...
+                      : computed_col_styles[id]
+                  }
+                >
+                  {node.data.is_m2m &&
+                  _.isFunction(first_node_m2m_display) &&
+                  depth === 0
+                    ? first_node_m2m_display(get_val(node))
+                    : _.isFunction(val_display)
+                    ? val_display(get_val(node))
+                    : get_val(node)}
```
Stephen-ONeil, line 194: IMO, `explorer_components.js` doesn't need to know about these weird special case displays for the first node in a many-to-many hierarchy. Bit of a concept leak. A developer just reading this code might have no idea what a `first_node_m2m_display`. In fact, the explorers using it could all go away or be refactored, and then it'd be a totally mysterious orphaned name inside this module.

I'd suggest `explorer_components.js` could pass `node` as a second argument to `val_display`. Then consumers of the explorer components could, optionally, implement additional `val_display` logic based on things like node depth, without the common explorer components needing to know anything about it.

client/src/components/misc_util_components.js
```diff
@@ -24,13 +23,32 @@ const ExternalLink = ({ children, href, title }) => (
...
+      ? `<span>(${prefixed_content})</span>`
+      : prefixed_content;
 
     return (
       <span
         style={style}
         className={className}
-        dangerouslySetInnerHTML={{ __html: formats[type](content) }}
+        dangerouslySetInnerHTML={{
+          __html: parenthesis_content,
```
Stephen-ONeil, line 50: Real minor nitpick, but at this point `parenthesis_content` would be more accurately named something like `potentially_parenthesised_potentially_prefixed_content`. 

It's a bit fiddly to name things well when making a series of nested conditional modifications. Could instead consider the prefix and parentheses to be part of the formatting, maybe even collapse the application of formatting in to a single chain and just have a formatting pipeline like
```
const formatted_content = _.chain(content)
  .thru(formats[type])
  .thru((content) => (prefix ? `<span>${prefix}${content}</span>` : content))
  .thru((content) => (in_parenthesis ? `<span>(${content})</span>` : content))
  .value();
```

Stephen-ONeil (changes requested):
I like the display! The text colour doesn't meet AA contrast though.

Bit fiddly to check because chrome dev tools still aren't great at this, but try setting the style directly on some text to have the explorer background grey and a color of `rgba(51, 51, 51, 0.5)`. Then, when you click on the color style, chrome will tell you the contrasts in the colour picker menu.

E.g.
image

client/src/explorer_common/resource_explorer_common.js
GarvinH: I think we wanted to have extra rounding to reduce the perceived precision of the estimation

client/src/explorer_common/explorer_components.js
```diff
@@ -181,7 +181,9 @@ export const ExplorerNode = ({
                 }
               >
                 {_.isFunction(val_display)
-                  ? val_display(get_val(node))
+                  ? depth === 0
+                    ? val_display(get_val(node), node)
+                    : val_display(get_val(node))
```
Stephen-ONeil, line 186: No reason `depth === 0` should be special here, always pass `node` as the second arg. Up to `val_display` to check depth and own any related logic.

Stephen-ONeil (changes requested):

client/src/search/SubjectWrappedTypeahead.js
```diff
@@ -0,0 +1,3 @@
+import React from "react";
+
+class SubjectWrappedTypeahead extends React.Component {}
```
Stephen-ONeil, line 3: Haha, I see you're continuing the typeahead cleanup (oops, delete)

Stephen-ONeil (approved):
Approved, but before merging see comment below.

client/src/search/SubjectWrappedTypeahead.js
GarvinH: whoops 😂 thanks

PR #970, opened 2021-03-25
Create storybook for pinned content

GarvinH:


client/.storybook/preview.js
```diff
@@ -26,3 +28,7 @@ export const parameters = {
 };
 
 addDecorator((storyFn, context) => withConsole()(storyFn)(context));
+
+addDecorator((story) => (
+  <MemoryRouter initialEntries={["/"]}>{story()}</MemoryRouter>
```
ktw1016, line 33: Why did we need this?

ktw1016 (approved):

client/.storybook/preview.js
GarvinH: PinnedContent uses `withRouter` which throws an error if there is no `Router` used

PR #971, opened 2021-03-25
Remove redundant height clipper

GarvinH:
`<Details>` already shortens the height


ktw1016 (approved):
Uhh yeah maybe it's for really long footnotes? But I think it's fine

PR #972, opened 2021-03-26
Drop grid system

### LINKED ISSUE #915: Drop grid-system.scss
Fairly legacy. The idea was something like a text-layout specific version of the bootstrap grid system, probably for similar reasons to our use of `fcol` (because we wanted flex layouts instead of Bootstrap 3's grid system). I think most of the text-specific pieces of this are unused though, it looks like it's more or less just acting as a confusing alternate to the other grid system. Still used in a couple obscure components, but I think it's worth refactoring all of those and dropping the whole thing.

Additional weirdness:
  - styles for the `KeyConceptList.js` live in `grid-system.scss`, just something to keep in mind while refactoring the grid system out of that component in particular
  - welcome mat uses it's own  copy of most of the grid system, namespaced via BEM... that can stay as is, refactoring the welcome mat as a whole will happen in its own time

GarvinH:
closes #915


ktw1016 (approved):
I think it's good!

PR #974, opened 2021-03-26
Remove top margin from SomeThingsToKeepInMind

GarvinH:
Previously, the top margin was used to put some space on between the cards and the `SomeThingsToKeepInMind`. Now the filter panel dropdown provides some margin of its own

before
image

with changes
image


ktw1016 (approved):
👍

PR #976, opened 2021-03-26
Fixed jumpy drop down animation

ChetanMandur:
Just something I noticed while using the website. [Specifically this page](http://localhost:8080/build/InfoBase/index-eng.html#orgs/gov/gov/infograph/financial).

Before, when clicking the 'FAQ for this data' dropdown, the animation displayed smoothly for the first half before jumping straight to the end.

### DIFF
--- a/client/src/components/Accordion/Accordions.js
+++ b/client/src/components/Accordion/Accordions.js
@@ -89,6 +89,10 @@ class AccordionEnterExit extends React.Component {
   }
 }
 
+AccordionEnterExit.defaultProps = {
+  max_height: "80vh",
+};
+
 const StatelessPullDownAccordion = ({
   max_height,
   title,
--- a/client/src/components/Accordion/Accordions.js
+++ b/client/src/components/Accordion/Accordions.js
@@ -24,7 +24,6 @@ function FirstChild(props) {
   return childrenArray[0] || null;
 }
 
-const defaultMaxHeight = "500px";
 class AccordionEnterExit extends React.Component {
   constructor() {
     super();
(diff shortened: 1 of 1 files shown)


client/src/components/Accordion/Accordions.js
```diff
@@ -24,7 +24,7 @@ function FirstChild(props) {
   return childrenArray[0] || null;
 }
 
-const defaultMaxHeight = "300px";
+const defaultMaxHeight = "500px";
```
GarvinH, line 27: Just some food for thought for the next person reviewing. Is there a reason we shouldn't have this at `80vh` or greater?

As far as I know, this height is only used for the `Transition` component/the actual animation occurring. With that being said, we know that any accordion according to the following class `StatelessPullDownAccordion` caps the child height (before overflowing) to `80vh`. Thus our default animation height would never exceed `80vh` anyways (well we'd have to add the accordion header and footer size as well).

With that being said, we could probably get rid of this `defaultMaxHeight` like so (untested, may need slight adjustment if we choose to do this):

```
const StatelessPullDown = ({max_height, ...existing_props}) => {
  <existing divs>
    <AccordionEnterExit
          className="pull-down-accordion-body"
          style={{ paddingTop: "5px" }}
          expandDuration={600}
          collapseDuration={600}
          max_height={max_height}
        >
          <div style={{ maxHeight: max_height, overflowY: "auto" }}>{children}</div>
        </AccordionEnterExit>
  </existing divs>
}
StatelessPullDownAccordion.defaultProps = {
  max_height: "80vh"
}
```

Now I guess we didn't really remove `defaultMaxHeight` in a sense, but we get the added control over the max content height of the accordion so it could be smaller or larger if we really wanted it to be the case

ktw1016: yeah `defaultMaxHeight` can prob go, and have defaultProps 80vh for both `StatelessPulldownAccordion` and also `AccordionEnterExit` (because it seems like it's used in explorer_common as well?)

ChetanMandur: I just pushed a change that (I think?) properly removes the `defaultMaxHeight` variable and uses the new `max_height` variable.

ktw1016:
Yeah looks good, I think just one more thing is that `AccordionEnterExit` should also have defaultProps for maxheight because it's being exported out, and being called in `explorer_common`

ktw1016 (approved):
👍

Issue #456, opened 2020-01-10
FTEs, and probably some spending, should be displayed as floats, not ints


Stephen-ONeil:
The FTEs are treated at integers by their Table class code, that for sure needs to change to decimal1 or decimal2. Should double check that other data that might contain decimals isn't also being rounded and that the formats used for their display are appropriate

GarvinH:
We could close this after #1001?

Stephen-ONeil:
Quick check over, seems good 👍

PR #982, opened 2021-03-29
Table decimal2 and dollar formatter

### LINKED ISSUE #977: Decimal2 and Dollar not being parsed into numbers
Currently `DisplayTable` is treating decimal2 and dollar as string values (since having 2 decimals requires a string format) causing the text align to stay on the left and total not summing up properly (also causing the table to be gigantic)

Should be done before #456 

Waiting for #921

GarvinH:
closes #977 
related #456


Stephen-ONeil (approved):
👍 

I hate the `Table` class and all it's related code, can't wait till it's all dropped.

PR #986, opened 2021-03-31
Cypress Setup

sranc046:
- [x] Install and configure Cypress
- [x] Test the InfoBase using Cypress
- [x] Convert existing tests to Cypress
- [x] Setup CI


sranc046:
Finally taking the time to write down a few thoughts I have for this branch...

1. As it stands, CI integration works fine (although will need some minor testing and/or updates to keep up with all the work being done regarding CircleCI and memory issues)
2. The initial route tests were supposed to be a direct reflection of what we currently have (we wait for the page to load and check if the spinner is gone). In theory this "should" work since when you visit a page with Cypress, it waits for the onLoad of the page to finish before continuing with the other tests. However in practice this seems unreliable. You can have a passing test when confirming the spinner exists or doesn't exist. Seems to be a weird timing issue when the test is being triggered. Although the test does work as intended when verifying the spinner no longer exist but the page does not completely load (ex: the server is not running and won't load all the data).
3. Using an arbitrary timeout time like our current tests to wait for the page to load is not recommended with Cypress and will throw an error.

Looking for any suggestions and/or thoughts on how to proceed. I had done some initial testing with server responses but will need a more in-depth understanding of it if we decide to go in that direction.

For Storybook testing, I think there still needs to be some discussions about how we go about storybooking our components so that we can develop useful tests for them.

sranc046:
Finally taking the time to write down a few thoughts I have for this branch...

1. As it stands, CI integration works fine (although will need some minor testing and/or updates to keep up with all the work being done regarding CircleCI and memory issues)
2. The initial route tests were supposed to be a direct reflection of what we currently have (we wait for the page to load and check if the spinner is gone). In theory this "should" work since when you visit a page with Cypress, it waits for the onLoad of the page to finish before continuing with the other tests. However in practice this seems unreliable. You can have a passing test when confirming the spinner exists or doesn't exist. Seems to be a weird timing issue when the test is being triggered. Although the test does work as intended when verifying the spinner no longer exist but the page does not completely load (ex: the server is not running and won't load all the data).
3. Using an arbitrary timeout time like our current tests to wait for the page to load is not recommended with Cypress and will throw an error.

Looking for any suggestions and/or thoughts on how to proceed. I had done some initial testing with server responses but will need a more in-depth understanding of it if we decide to go in that direction.

For Storybook testing, I think there still needs to be some discussions about how we go about storybooking our components so that we can develop useful tests for them.

There was also one random file that had failed the prettier --check some time during my commits that I had to add. No idea where that came from.

Lastly, I could probably squash (or collapse, not sure which is the right term) some or most of my commits as they are useless small changes I was doing to test the CI.

.circleci/config.yml
```diff
@@ -245,9 +252,11 @@ jobs:
   # Note: the image used for this job is built from a docker file found in ../dockerfiles/test
   test_end_to_end: # slower but more in-depth than the test job. Wait until these have passed before deploying the client
     docker:
-      - image: eaadtbs/ib-ci-test:1.4
+      - image: eaadtbs/ib-ci-cypress-test:latest
```
Stephen-ONeil, line 255: Was there actually something we needed to change in the image, in the end?

dockerfiles/test/Dockerfile
```diff
@@ -20,20 +25,21 @@ ENV PATH $SERVICE_HOME/google-cloud-sdk/bin:$PATH
...
+
+RUN wget -qO - https://www.mongodb.org/static/pgp/server-4.4.asc | apt-key add -
+RUN echo "deb http://repo.mongodb.org/apt/debian buster/mongodb-org/4.4 main" | tee /etc/apt/sources.list.d/mongodb-org-4.4.list
+RUN apt-get update && apt-get install -y mongodb-org
 
 # create default mongodb data directory and give the $USER access
 RUN mkdir -p /data/db && chown -R ${SERVICE_USER} /data/db
 
+RUN npm install --save-dev cypress
+RUN $(npm bin)/cypress verify
```
Stephen-ONeil, line 41: Do we need cypress globally installed?

package.json
```diff
@@ -19,14 +21,18 @@
   "license": "MIT",
   "dependencies": {},
   "devDependencies": {
+    "axe-core": "^4.2.0",
     "babel-eslint": "^10.1.0",
+    "cypress": "^6.9.1",
+    "cypress-axe": "^0.12.2",
     "eslint": "^7.12.1",
+    "eslint-plugin-cypress": "^2.11.2",
```
Stephen-ONeil, line 29: I think we should move the cypress dependencies/install in to `/client`, it won't be needed for the other sub-repos.

client/src/handlebars/register_helpers.side-effects.js
```diff
@@ -211,52 +211,37 @@ Handlebars.registerHelper("will_change_to", function (val, formatter, context) {
...
+  function (val, genre, nombre, formatter, context) {
+    return fr_master_change(val, formatter, "to", "past", genre, nombre);
+  }
+);
+Handlebars.registerHelper(
+  "fr_changed_by",
+  function (val, genre, nombre, formatter, context) {
+    return fr_master_change(val, formatter, "by", "past", genre, nombre);
+  }
+);
```
Stephen-ONeil, line 225: Did prettier decide to reformat this file for some reason? Do you know why?

Stephen-ONeil (changes requested):
First pass, just asking a bunch of question, haven't looked in to the tests themselves too closely but will soon.

.circleci/config.yml
```diff
@@ -104,6 +104,8 @@ jobs:
       - image: eaadtbs/ib-ci-build:1.4
         <<: *dockerhub_auth
     working_directory: "~/InfoBase"
+    environment:
+      CYPRESS_CACHE_FOLDER: '~/InfoBase/.cache/Cypress'
```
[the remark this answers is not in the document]
sranc046: Yeah, as implied in the name, it specifies where to save the cypress cache folder. The default location was `~/` which caused issues when trying to persist it to workplace (didn't have access to it).

.circleci/config.yml
```diff
@@ -350,10 +359,12 @@ jobs:
       - run: if [[ $CIRCLE_BRANCH = master ]]; then ./scripts/ci_scripts/store_bundle_stats.sh; fi # only saving master stats right now since they're somewhat large files (~1 mb each)
 
 
+orbs:
+  cypress: cypress-io/cypress@1
+
```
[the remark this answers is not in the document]
sranc046: Yes, didn't do a full clean up yet

.circleci/config.yml
```diff
@@ -117,6 +119,7 @@ jobs:
       - save_cache:
           paths:
             - node_modules
+            - .cache
```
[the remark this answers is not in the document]
sranc046: Yes, it is made during the npm ci command

.circleci/config.yml
sranc046: Yeah the image uses cypress:base as the base image and everything else is built on top of it. Without a cypress base image there were some missing dependencies.

dockerfiles/test/Dockerfile
sranc046: Can double check, this might be a leftover from docker image testing

client/src/handlebars/register_helpers.side-effects.js
sranc046: Yeah at some point I had encountered a similar error that Christina (I think) was having with prettier and husky or something. The file needed to be reformatted and committed so that I could push, no clue why it popped up when it did.

.circleci/config.yml
Stephen-ONeil: Ok. We were using `ib-ci-test` for multiple jobs. If the cypress version works for all of those and isn't much of a spin-up time difference, we can reuse it for other jobs too. Otherwise, we'll need to keep the old test image dockerfile around (currently the cypress image replaces it).

cypress/integration/ignoredTestFiles/route_tests.spec.js
```diff
@@ -0,0 +1,40 @@
...
+  route_load_tests_config.map((routes) => {
+    describe(`${routes.name} route`, () => {
+      routes.test_on.map((app) => {
+        it(`Tested on ${app}`, () => {
+          cy.visit(
+            `http://localhost:8080/build/InfoBase/index-${app}.html#${routes.route}`
+          );
+          cy.injectAxe();
+          cy.get(".leaf-spinner-inner-circle").should("not.exist");
+          cy.checkA11y(null, null, terminalLog, true);
```
Stephen-ONeil, line 35: Does cypress properly fail tests when an error is thrown, or do we still need to explicitly check that the error route hasn't been loaded here? Something to double check after we solve the issue of making routes wait until all spinners are gone consistently, probably.

Stephen-ONeil (changes requested):
> In theory this "should" work since when you visit a page with Cypress, it waits for the onLoad of the page to finish before continuing with the other tests. However in practice this seems unreliable.

The `load` event isn't going to be reliable for us. It fires when all resources are done loading, including resources requested by scripts... unless those scripts request those resources after the `load` event has fired. Can happen pretty regularly on a lot of routes with multiple layers of async loading. At some point it sees that everything's in and fires the even, not knowing (or really being responsible for) if those requests ending were about to trigger further requests.

Try adding something like this in the top level of root.js (start of the js entry point, will be in the initial bundle loaded through the html entry so guaranteed to be pre-load event triggering).
```
window.onload = () => {
  debugger;
};
```
See where it pauses on a few different routes, whether there's an active spinner at that time/one appears again after you continue. Can also switch to the network tab while paused and see what requests are pending (the `onLoad` event goes in the queue, I believe other code gets ahead of it and makes further requests before it flushes, but I could be wrong about something here).

This was why there was a hacky timed-out retry checking for the existence of spinners in testcafe code, to catch loading that started only after the main loading/spinners were done. We'll need to solve the same problem with cypress, although I imagine we're far from the first to have it. No immediate thoughts, not knowing cypress well, but if you want help later let me know.

sranc046 (commented):
Steps towards dynamic waiting and intercepting graphql responses before asserting if page is loaded. Work around for caching (can't intercept cached data). No built in function to wait for all aliases of same name to finish (right now the assertion is triggered after a first instance of the alias is intercepted), working on a solution for that. Not applicable to routes that don't require graphql queries.

client/.eslintrc.json
```diff
@@ -4,7 +4,8 @@
     "../.eslintrc.json",
     "plugin:import/typescript",
     "plugin:react/recommended",
-    "plugin:react-hooks/recommended"
+    "plugin:react-hooks/recommended",
+    "plugin:cypress/recommended"
```
Stephen-ONeil, line 8: Could scope this to just the `/cypress` dir by adding a `/cypress/.eslintrc.json` file that extends with this plugin.

.circleci/config.yml
```diff
@@ -94,40 +94,57 @@ jobs:
...
+      - image: markhobson/node-chrome:latest
+        <<: *dockerhub_auth
+    steps:
+      - checkout:
+          path: "~/InfoBase"
+      - run: ./scripts/ci_scripts/setup_lighthouse_env.sh
+      - run: ./scripts/ci_scripts/cleanup_lighthouse_db.sh
+      - run:
+          name: lhci autorun
+          command: node ./node_modules/@lhci/cli/src/cli.js autorun --config ./scripts/ci_scripts/lighthouserc.js
```
Stephen-ONeil, line 109: Was bringing this back this (or adding it?) intentional? I believe we dropped the lighthouse job on master quite a while ago.

Stephen-ONeil (changes requested):
I don't expect the GraphQL-watching approach to work out. In particular, it feels like a deal breaker that it would only make sense on routes that use server data. Some routes like that about page never will, but we still want them to be part of this basic test coverage. At most, checking for in-progress gql queries could be an optional part of the test logic, but it can't be the only way we try to detect that the page is "done".

Did something rule out just re-checking for the spinner component's existence on some timeout?

dockerfiles/cypress/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+    curl \
+    python
+
+USER ${SERVICE_USER}
+
+RUN curl -sSL https://sdk.cloud.google.com | bash && \
+  exec sh && \
+  gcloud init
+
+ENV PATH $SERVICE_HOME/google-cloud-sdk/bin:$PATH
```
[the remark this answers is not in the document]
Stephen-ONeil: Unless this is related to https://github.com/TBS-EACPD/infobase/pull/986#discussion_r640005254

If there _is_ a good enough reason to reuse this image for multiple jobs that's fine, might just need to give it a more generic name at some point. Probably best to have separate images though.

cypress/integration/ignoredTestFiles/route_tests.spec.js
```diff
@@ -0,0 +1,40 @@
+const {
```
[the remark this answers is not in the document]
Stephen-ONeil: Still curious.

client/cypress/integration/ignoredTestFiles/route_tests.spec.js
```diff
@@ -31,7 +273,9 @@ describe("Route tests", () => {
             `http://localhost:8080/build/InfoBase/index-${app}.html#${routes.route}`
           );
           cy.injectAxe();
-          cy.get(".leaf-spinner-inner-circle").should("not.exist");
+          // eslint-disable-next-line cypress/no-unnecessary-waiting
+          cy.wait(5000);
```
sranc046, line 277: This "works" but was trying to avoid as per: https://docs.cypress.io/guides/references/best-practices#Unnecessary-Waiting. Since multiple routes have a number of different things being loaded after the initial page load, we need to wait longer to assert that the spinner is no longer present, so maybe it's an okay rule break? But this is why I've been trying other methods of delaying the spinner assertion such as intercepting some requests.

sranc046: Just now noticing the wait is at the wrong place, should be after `cy.visit()` command. Mix up happened after some clean up before committing... but concept doesn't change. The file is being ignored from CI for now regardless.

Stephen-ONeil (changes requested):
From skimming around the linked docs, seems their point is about using arbitrary `wait` calls where they specifically are not necessary. E.g. 
  - `visit()` already waits till the `load` event fires, so _probably_ an additional arbitrary wait isn't necessary (but as previously discussed, `load` is not solid for a single page app with dynamic imports; let me know if you want me to clarify this)
  - the example with unnecessary `wait` and get is about triggering some event, waiting, then asserting that the expect change occurred because `cy.get(...).should(...)` has a built in wait/retry cycle (good to know, similar to TestCafe's equivalent, bet it has some built in time-out, hopefully we can customize that)
etc.

I agree with them on that. Were the `wait`s necessary to make the tests work here? If so, I'm going to guess that's because, thanks to dynamic loading throwing off `visit`, `cy.get(".leaf-spinner-inner-circle").should("not.exist")` fires pretty early and times out over our relatively long load times. If I'm right, and if the API lets us configure that, then we don't need `wait` and just need to give a (still arbitrary) longer time for the `should("not.exist")` assertion to come true.

Leaves us with the gotcha of sub-route loading spinners. Possible we'll have shaky timing on if follow-up spinners will spawn before or after the assertion. Took a glance and not sure the existing route tests account for this (but maybe it's something that testcafe covers, possible that it tries until the assertion _stays_ passing or something?). For all I know, cypress covers this out of the box :shrug:, read the docs and let me know!

While I'm looking at it, the old route load tests have some additional checks that'd be good to port over. Checking that the `'#app` element is non-empty for instance. The `#app` element has no children in the html templates, so this asserts that our JS actually loaded and ran. Additionally, the first thing the JS renders in to the DOM is a loading spinner, so this check just so happens to (well, by design but I didn't clearly document this) work as a pre-condition against the loading spinner test failing positive.

There's also a small wait time post-loading assertion. That one I documented at least, encountered bugs that occurred shortly after all loading finished, so holding the test open just a little bit past loading makes it more robust.

sranc046:
Will keep an eye on https://github.com/component-driven/cypress-axe/pull/91. Cypress-axe isn't fully supported by cypress 7 yet (but causes no issue from our use so far).

sranc046:
Running axe through Cypress on the footnote-inventory route causes a complete system crash locally and a timeout failure within CircleCI... might want to avoid that one.

.circleci/config.yml
```diff
@@ -274,7 +282,7 @@ jobs:
           command: (cd client && npm run ci_serve)
           background: true
       - run: (cd client && npm run ci_route_load_tests)
```
Stephen-ONeil, line 284: The old route load tests don't work in the current docker image, but that's fine. Cypress seems to be working out so probably about time to go around deleting all the old route load test stuff.

In theory it'd be best to run the old and new tests alongside eachother for a bit, but the old tests were flaky anyway so that would just be noisy. Better to just re-test the cypress equivalents on their own (e.g. go around introducing different types of error-throwing bugs in to different routes and re-running the tests to make sure they're properly caught).

Stephen-ONeil (changes requested):
> Running axe through Cypress on the footnote-inventory route causes a complete system crash locally and a timeout failure within CircleCI... might want to avoid that one.

Ouch. You said something about axe support for cypress 7 being unfinished, right? Is this related?

Maybe add a flag for `skip_axe` or something to the config file, slap it on the footnote inventory (and any other non-user-facing routes we might test, like hte panel inventory maybe). Lucky it's an internal route that's crashing since it's not an a11y concern and would have just been spitting out noisy warning anyway. Hopefully all the user-facing routes work with axe still.

Side note, just looked and saw that the cypress tests are much slower. Seen any tips on how to speed them up? TestCafe had a concurrency option, but a quick google makes it look like the official cypress parallelization solution is... just run it across multiple VMs :roll_eyes:. Might be other things we can do though, maybe see if you can run the test browser in headless mode or something (don't bother actually rendering, we aren't doing visual tests).

Stephen-ONeil:
Oh right, meant to comment this. Looking at `227471a` and `f52f9df`, I think that second commit was supposed to be part of/an edit to the first commit, right? Going by their names and content. If you need to add more changes to your previous commit, you can use `git commit --amend` :+1:

.circleci/config.yml
sranc046: Yeah it clicked that the old tests weren't going to work as soon as I pushed, forgot about the image🤦‍♂️

sranc046:
> If you need to add more changes to your previous commit, you can use `git commit --amend` 👍

That was the plan, it was just getting late and I messed up the ammend😅

Stephen-ONeil (changes requested):
Mostly just cleanup stuff left, close! Also, please rebase off master before the next (last?) review.

client/package.json
```diff
@@ -49,6 +51,7 @@
     "@nivo/line": "^0.72.0",
     "@nivo/pie": "^0.72.0",
     "@nivo/treemap": "^0.72.0",
+    "axe-core": "^4.3.1",
```
Stephen-ONeil, line 54: Showing up in both dependencies and dev-dependencies. Just manually delete it from regular dependencies (which are packages that are used by the run-time client).

Stephen-ONeil (approved):
A couple of the cypress files have comment blocks about them being example files. IF that's true (and the file is unused), please delete it. If that's false and the file is actually used/has been customized from the example, please drop any of those comments that could be misleading/aren't specifically relevant. Devs in the future should refer to the docs rather than comments in those committed files, to be update-proof.

Other than that and the one other cleanup comment, approved :tada:! Go ahead and merge once you've handled that.

PR #987, opened 2021-03-31
Changed dropdown animation in tag explorer

ChetanMandur:
[The page in question](http://localhost:8080/build/InfoBase/index-eng.html#tag-explorer/dept/planned)
Currently, the dropdown menu has a fade effect/animation that causes text to overlap when opening and closing. To fix this, I just made both the speed faster and the opacity lower.

I experimented with getting a 'blanket' effect working (like on the [FAQ](http://localhost:8080/build/InfoBase/index-eng.html#orgs/gov/gov/infograph/financial) drop-down, where the content gets more exposed as the dropdown moves) but I couldn't seem to get it working. I'm still open to getting that feature added, I just wanted to make a PR of what I currently have.

### DIFF
--- a/client/src/components/Accordion/Accordions.js
+++ b/client/src/components/Accordion/Accordions.js
@@ -32,19 +32,19 @@ class AccordionEnterExit extends React.Component {
     const node = ReactDOM.findDOMNode(component);
     const initialHeight = node.offsetHeight;
     select(node)
-      .style("opacity", 1)
+      .style("opacity", this.props.closing_opacity)
       .style("max-height", initialHeight + "px")
       .transition()
       .ease(easeLinear)
       .duration(this.props.collapseDuration)
-      .style("opacity", this.props.opacity)
+      .style("opacity", this.props.opening_opacity)
       .style("max-height", "1px");
   };
   onEntering = (component) => {
     const node = ReactDOM.findDOMNode(component);
     select(node)
       .style("max-height", "0px")
-      .style("opacity", this.props.opacity)
+      .style("opacity", this.props.opening_opacity)
       .transition()
       .ease(easeLinear)
       .duration(this.props.expandDuration)
@@ -90,7 +90,8 @@ class AccordionEnterExit extends React.Component {
 
 AccordionEnterExit.defaultProps = {
   max_height: "80vh",
-  opacity: 1e-6,
+  opening_opacity: 1e-6,
+  closing_opacity: 1,
 };
 
 const StatelessPullDownAccordion = ({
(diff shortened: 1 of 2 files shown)


client/src/explorer_common/explorer_components.js
```diff
@@ -181,15 +181,20 @@ export const ExplorerNode = ({
                 }
               >
                 {_.isFunction(val_display)
-                  ? val_display(get_val(node), node)
```
GarvinH, line 184: I'm guessing this was a mistake haha

ChetanMandur: Good catch haha. I have no idea why I even changed that line. It must have been by mistake.

That explanation makes more sense, thank you! I was just now playing with it trying to understand what it was doing 😅

GarvinH:
Jeez that took me a quite a bit to isolate... Here's the culprit into making the "blanket" effect happen

```
<AccordionEnterExit
          expandDuration={500}
          collapseDuration={300}
+         style={{ overflowY: "hidden" }}
        >
+       <div style={{ overflowY: "auto" }}>
            {get_children_content({
              node,
              depth: depth + 1,
              explorer_context,
            })}
+       </div>
</AccordionEnterExit>
```

With that, I think you could (probably) remove the rest of the changes?

ChetanMandur:
🙏🏽 Thank you @GarvinH for finding that. It was the final thing still bugging me with the animation.

After playing around with the blanket effect, I decided to restore the opacity values back to default. I kept the speed changes as I feel it makes the UI a bit more intuitive.

client/src/explorer_common/explorer_components.js
```diff
@@ -99,7 +99,7 @@ const get_children_content = ({
           typeName="ul"
           className="ExplorerNodeContainer__ChildrenList"
           staggerDurationBy="0"
-          duration={500}
+          duration={400}
```
GarvinH, line 102: Do we still need this change?

If i'm correct, that duration is for the `FlipMove`

client/src/components/Accordion/Accordions.js
```diff
@@ -91,6 +90,8 @@ class AccordionEnterExit extends React.Component {
 
 AccordionEnterExit.defaultProps = {
   max_height: "80vh",
+  opening_opacity: 1e-6,
+  closing_opacity: 1,
```
GarvinH, line 94: I'm not sure if we'd want to have the animation opacities be custom (or what use-cases would that be)

I'll probably let someone else provide some input on this haha

client/src/explorer_common/explorer_components.js
ChetanMandur: I forgot to note this in the PR, but I added this change on purpose. I just thought `500` was a bit too slow. 
More of a preference-based change than anything. Open to change.

ktw1016 (approved):
👍🎉

PR #991, opened 2021-04-05
Apollo batching

AlexCLeduc:
This involved moving from express-graphql to apollo-server. 

- [x] Try getting this to work with our special GET requests (we'll probably have to move back to GET + body, hopefully we can get that to cache properly)
- [x] See if batching works in dev-links
- [x] ~~cloudflare considerations?~~ (we weren't cloudflare caching anything before)
- [x] graphiql considerations (apollo-server seems to replace graphiql with apollo playground)
    - infobase's `#graphiql` route still works, but since introspection is turned off in prod, it won't support docs or autocomplete unless you're developing locally
- [x] turn `validationRules: [depthLimit(15)]` security back on using apollo-server options 
- [x]  Cross browser testing
    - [x] IE11 and iOS both make requests and get responses properly
    - [x] unfortunately we need to add a `String.prototype.replaceAll` polyfill for IE11 

Screen Shot 2021-04-05 at 12 23 30 PM

### DIFF
--- a/client/src/graphql_utils/graphql_utils.js
+++ b/client/src/graphql_utils/graphql_utils.js
@@ -57,24 +57,23 @@ export const query_length_tolerant_fetch = async (uri, options) => {
   // important, this regex lazy matches up to and including FIRST ? occurence, which (in a URI)
   // should be where the query string starts. I've complicated it slightly just in case there's ever a ? IN
   // the query string (well, that'd be an encoding error anyway)
-  // const url_encoded_query = uri.replace(/^(.*?)\?/, "");
+  const query = options.body;
+  const query_hash = string_hash(query);
 
-  // const query_string_hash = string_hash(url_encoded_query);
-
-  // const short_uri = `${await get_api_url()}?v=${sha}&queryHash=${query_string_hash}`;
-  const short_uri = `${await get_api_url()}?v=${sha}`;
+  const uri = `${await get_api_url()}?v=${sha}&queryHash=${query_hash}`;
 
   const new_options = {
     ...options,
+    method: "GET",
+    body: undefined,
     headers: {
       ...options.headers,
-      // "encoded-compressed-query": compressToBase64(
-      //   decodeURIComponent(url_encoded_query)
-      // ),
+      // "encoded-compressed-query": compressToBase64(query),
+      "gql-query": query,
     },
   };
 
-  return fetch(short_uri, new_options);
+  return fetch(uri, new_options);
 };
 
 let client = null;
--- a/client/src/graphql_utils/graphql_utils.js
+++ b/client/src/graphql_utils/graphql_utils.js
@@ -53,9 +53,6 @@ export const get_api_url = async () => {
 
 // Makes our GET requests tolerant of long queries, sufficient but may not work for arbitrarily long queries
 export const query_length_tolerant_fetch = async (uri, options) => {
-  // important, this regex lazy matches up to and including FIRST ? occurence, which (in a URI)
-  // should be where the query string starts. I've complicated it slightly just in case there's ever a ? IN
-  // the query string (well, that'd be an encoding error anyway)
   const query = options.body;
   const query_hash = string_hash(query);
 
@@ -67,7 +64,6 @@ export const query_length_tolerant_fetch = async (uri, options) => {
     body: undefined,
     headers: {
       ...options.headers,
-      // "encoded-compressed-query": compressToBase64(query),
       "gql-query": query,
     },
   };
--- a/client/src/graphql_utils/graphql_utils.js
+++ b/client/src/graphql_utils/graphql_utils.js
@@ -6,7 +6,6 @@ import {
 import { BatchHttpLink } from "@apollo/client/link/batch-http";
 
 import _ from "lodash";
-import { compressToBase64 } from "lz-string";
 import React from "react";
 
 import string_hash from "string-hash";
--- a/client/src/graphql_utils/graphql_utils.js
+++ b/client/src/graphql_utils/graphql_utils.js
@@ -60,7 +60,7 @@ export const query_length_tolerant_fetch = async (uri, options) => {
   const query = options.body;
   const query_hash = string_hash(query);
 
-  const uri = `${await get_api_url()}?v=${sha}&queryHash=${query_hash}`;
+  const uriWithVersionAndQueryHash = `${await get_api_url()}?v=${sha}&queryHash=${query_hash}`;
 
   const new_options = {
     ...options,
@@ -73,7 +73,7 @@ export const query_length_tolerant_fetch = async (uri, options) => {
     },
   };
 
-  return fetch(uri, new_options);
+  return fetch(uriWithVersionAndQueryHash, new_options);
 };
 
 let client = null;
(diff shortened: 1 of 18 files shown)


AlexCLeduc:
Got it to work with GET requests, and it seems to cache locally

Screen Shot 2021-04-08 at 9 33 29 PM
Screen Shot 2021-04-08 at 9 34 15 PM

client/src/graphql_utils/graphql_utils.js
```diff
@@ -56,32 +56,32 @@ export const query_length_tolerant_fetch = async (uri, options) => {
...
     ...options,
+    method: "GET",
+    body: undefined,
     headers: {
       ...options.headers,
-      "encoded-compressed-query": compressToBase64(
-        decodeURIComponent(url_encoded_query)
-      ),
+      // "encoded-compressed-query": compressToBase64(query),
+      "gql-query": query,
```
AlexCLeduc, line 71: Note that I stopped encoding it to header that holds the query. It's less annoying to debug, and as far as I can tell HTTP headers don't need to be encoded. 

That being said, HTTP headers should be ASCII, and might not support UTF8. This might cause an issue when we start using the API for search queries, and someone searches for something with accents. If this causes a bug, we can just `_.deburr` queries. All the searching should ignore diacritics/accents anyway.

client/src/models/covid/populate.js
```diff
@@ -23,7 +23,7 @@ export const api_load_years_with_covid_data = (subject) => {
         return {
           is_loaded: subject_is_loaded(subject.id),
           level: "dept",
-          id: subject.id,
+          id: String(subject.id),
```
AlexCLeduc, line 26: Apollo server doesn't cast integers to strings the way express-graphql does, so I did that manually. 

We could alternatively make the department queries in the schema take integer arguments 🤷

server/src/app.js
```diff
@@ -33,29 +34,30 @@ app.use(
...
   } else {
     console.log(`Request type: ${req.originalUrl}, ${req.method}`);
   }
 
-  process.env.USE_REMOTE_DB &&
-    console.log(JSON.stringify(get_log_object_for_request(req)));
+  if (process.env.USE_REMOTE_DB) {
+    get_log_objects_for_request(req).forEach((log_obj) => {
+      console.log(JSON.stringify(log_obj));
+    });
```
AlexCLeduc, line 59: Since a request can now contain multiple queries, we log each query. Part of the log objects is a batch_hash, which can be used to identify which query logs were part of the same request

client/src/graphql_utils/graphql_utils.js
```diff
@@ -56,32 +56,32 @@ export const query_length_tolerant_fetch = async (uri, options) => {
   // important, this regex lazy matches up to and including FIRST ? occurence, which (in a URI)
   // should be where the query string starts. I've complicated it slightly just in case there's ever a ? IN
   // the query string (well, that'd be an encoding error anyway)
```
Stephen-ONeil, line 58: This comment no longer applies

client/src/graphql_utils/graphql_utils.js
Stephen-ONeil: I believe this, or a similar post, was what informed my previous decision to compress: https://stackoverflow.com/questions/3326210/can-http-headers-be-too-big-for-browsers#:~:text=My%20Conclusion%3A%20If%20you%20want,256KB%20for%20all%20headers%20together
  
Don't know if Firefox still has the same 10KB per-header limit. That's still possibly up to 10k characters of query though, so it _sounds_ like we're totally safe even with batching growing query sizes? I'll try and dig up the old PR where I added this just to check whether I was adding it out of extreme caution, or if I'd actually had issues in the wild.

I do agree it'd be nice to stop it, it really is a huge pain when debugging.

client/src/models/covid/populate.js
Stephen-ONeil: Ouch. Yeah, the inconsistency in id types between int and string bugs me (and creates bugs) a lot. Org ids should probably just always be strings, doesn't really matter that they happen to only have int values. Other subjects pretty much all have non-int ids so it's only the org subjects that have the ambiguity headache.

server/package.json
```diff
@@ -10,14 +10,16 @@
...
   "dependencies": {
+    "apollo-server": "^2.22.2",
+    "apollo-server-express": "^2.22.2",
     "body-parser": "^1.19.0",
     "compression": "^1.7.4",
     "cors": "^2.8.5",
     "d3-dsv": "^1.1.1",
     "dataloader": "^1.4.0",
     "express": "^4.17.1",
     "express-graphql": "^0.6.12",
```
Stephen-ONeil, line 21: We can drop the express-graphql and compression dependencies now, right? Any other packages we've stopped using here?

server/src/app.js
```diff
@@ -19,10 +20,10 @@ connect_db().catch((err) => {
   // just logging, not trying to recover here. DB connection is reattempted per-request below
 });
 
-const schema = create_schema();
 const app = express();
 
-app.use(body_parser.json({ limit: "50mb" }));
+//TODO: figure out if we need this anymore, since apollo seems to work without it?
+// app.use(body_parser.json({ limit: "50mb" }));
```
Stephen-ONeil, line 26: Hmm, don't even remember what this was for. Maybe introspection queries got too big for the express defaults? Probably should have only been on in dev if that was the case, oops.

Stephen-ONeil (commented):
Awesome! Mostly minor comments here. Going to test out in a bunch of browsers and try and sniff out any gotchas before approving.

Stephen-ONeil:
Follow up on what I asked about in the dev meeting, can confirm that (as expected) Apollo still de-dupes queries in batches and skips queries that are already in-flight/cached. 

Well, hard to say for certain about in-flight requests, timing being what it is, but pretty confident about that either way. No reason for Apollo to lose the ability to be smart about that just because batching is on. 

If you want to look at that your self, a good place to do it is in the covid infographic (with `COVID_EXPENDITUES_FLAG` changed to true in covid_config.js). Both the covid authorities and covid expenditures panels make a query for `query_gov_covid_summaries` on mount, but it only shows up once in all the batched queries.

client/src/graphql_utils/graphql_utils.js
Stephen-ONeil: Original API GET requests PR: https://github.com/TBS-EACPD/infobase-archive/pull/753

Didn't leave much explanation (either of the problem or solution), possible I never figured out what was actually wrong. Looks like I ruled out the per-header limit being the issue at least. Sounds like certain longer queries were somehow throwing CORs errors, and I don't mention any particular browser so it must have been a fully server side issue. Don't see it happening anywhere now though 🤷 .

Stephen-ONeil (approved):
Everything seems solid to me! Pre-approved, pending
  - [x] any action you want to take from the minor comments above
  - [x] polyfilling `replaceAll`
  
If you're busy on Titan, I can wrap this up for you 👍

server/package.json
AlexCLeduc: Good point, forgot to remove express-graphql. 

We're still using compression to gzip responses, which are going to become even more redundant with batching.

server/src/app.js
AlexCLeduc: nice find, got rid of it

ktw1016 (commented):
I looked through and mostly seems good! Would just need to clean up the express graphql stuff before merging. I tried testing it out with covid and results but difference was pretty minimal as they only make 2-3 queries. I'll experiment it with service inventory queries and let you know once it merges

Issue #377, opened 2019-11-25
hover tooltip improvements


ktw1016:
![Screen Shot 2019-11-25 at 2 42 50 PM](https://user-images.githubusercontent.com/29746626/69572320-e5e6bc80-0f91-11ea-9f55-6a8cac0d358a.png)
Seems like a valid point

Stephen-ONeil:
1) is more about reviewing glossary content. Not going to put popups in popups.

2) is possibly valid some times, but is very case-by-case. Not something I want to force a full text review for, we just need to be applying our best judgement any time we write new/update old text.

PR #992, opened 2021-04-06
Add tooltips to (previously tool-tipped) terms on different panels

### LINKED ISSUE #989: Tippy.js bolded not actting like from bootstrap 3
Bolded text causes text to shift down a line.

Bolded text also causes description to be bolded

image

image

GarvinH:
closes #989 
related #377


ktw1016 (approved):
Had a quick look over, looks good!

PR #995, opened 2021-04-07
Fix jumping to panel on filter change with url options

### LINKED ISSUE #994: infographic panel filter should clear the panel URL
there's a bug that can easily be reproduced:

1. visit this url that has a panel parameter: https://www.tbs-sct.gc.ca/ems-sgd/edb-bdd/index-eng.html#orgs/gov/gov/infograph/financial/.-.-(panel_key.-.-'in_year_estimates_split)
2. scroll up to the panel filter
3. select program spending, filtering occurs as expected. In particular this filters out the panel identified by the url param
4. select tabled estimates, **this will cause the page to scroll to the in-year-estimates panel**

I think we can fix this if we clear the panel key from the URL  (probably using `replaceState`, to avoid extra history entries) when a user uses the filter feature

GarvinH:
closes #994
related #988

### DIFF
--- a/client/src/infographic/Infographic.js
+++ b/client/src/infographic/Infographic.js
@@ -172,7 +172,7 @@ class InfoGraph_ extends React.Component {
     this.scroll_to_panel_when_all_loading_done.cancel();
   }
   render() {
-    const { subject, active_bubble_id } = this.props;
+    const { subject, active_bubble_id, url_replace } = this.props;
     const {
       loading,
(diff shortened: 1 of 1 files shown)


GarvinH:
With the `#` it's only considered a hash rather than a route.

The `createHashHistory` in `bootstrapper` automatically strips the `/` for us and leaves a `#`

ktw1016 (approved):
👍

PR #996, opened 2021-04-07
hotfix add related tags for wwh tags

ktw1016:


AlexCLeduc (approved):
thanks for the quick fix

PR #1004, opened 2021-04-09
Service Inventory w/ apollo hooks

ktw1016:
Had so much trouble trying to rebase #957, so I just started a new branch and manually copied over the changes.. What a headache. I'll need to merge the whole service inventory PR sooner or later with service feature flag turned off


client/src/graphql_utils/graphql_utils.js
```diff
@@ -75,9 +75,8 @@ let client = null;
 export function get_client() {
   if (!client) {
     client = new ApolloClient({
-      link: new BatchHttpLink({
+      link: createHttpLink({
```
ktw1016, line 78: Haven't switched back to `compressToBase64`, let me know if that's necessary here

Stephen-ONeil (changes requested):
- [x] Panel level spinners within the service panels are a different size from other infographic panels that handle their own loading. Please make that consistent. No strong opinion which size of spinner you standardize on though
 - [ ] just a reminder to disable the services feature flag (and make sure everything that should be disabled is, e.g. section in the datasets page)
 
Otherwise approved, so that it gets out of the way of rebasing.

I'll be coming along with a follow up PR of my own to try and standardize client side query management between the COVID and services data (maybe even finally update the results loading too, but I expect that will justify a whole separate PR of its own). Things I'll probably be doing:
  - centralizing and identifying all possible service queries
    - no more panels passing in fragments
    - giving all service queries unique names for both client side logging and server side logging (via the `_query_name` query variable)
  - expanding the existing `query_logging_wrapper` utility to return both a loading promise and a loading hook. Shouldn't need to write a custom hook for every query, re-implement the client side logging logic, etc
  
As I said before, I don't think any components should be writing/using their own custom queries. Queries/query hooks should be centrally defined in the relevant sub-dir of `/models`. Those queries/hooks should also contain some initial response processing steps. Ideally, we should be able to refactor the server side schemas without changing anything on the client outside of the query files in `/models`.

client/src/models/populate_services.js
```diff
@@ -113,217 +223,137 @@ export function api_load_subject_has_services(subject) {
...
-  return { services, serviceStandards };
+  `;
+  const res = useQuery(query, {
+    variables: {
+      lang,
+    },
+  });
+  const { loading, error, data } = res;
+  if (error) {
+    const resp_time = Date.now() - time_at_request;
```
Stephen-ONeil, line 284: Just pointing this out, I'm trying to find a good pattern for timing the `useQuery` hook on my own branch right now, but this isn't actually the time it took to fetch this data. When execution reaches this line, `time_at_request` is actually more like "time at response received" and `resp_time` is more like "execution time for lines 268 to 284".

ktw1016:
> * [x]  Panel level spinners within the service panels are a different size from other infographic panels that handle their own loading. Please make that consistent. No strong opinion which size of spinner you standardize on though

I think the only actual spinner equivalent to how `inline_panel` config size is used is in `tp_by_region`. The other spinners are used for tabs, inside result drilldown, indicator panel, etc. So I made change to `tp_by_region` only as I think each of the sizes in different places serve different purposes

> centralizing and identifying all possible service queries

For this, this is going against the reason for the switching to hooks. The best practice (I've been trying to find link to prove this but couldn't find any for the time being😓) is to keep data fetching to where it's needed inside the components, hence to actually decenteralize it. I think we should however, have a wrapper for a hook wrapper to "centralize" the logging piece and other general things

So to summarize, queries should be happening inside individual panel components and it should pass in its own query fragments to only request exactly what it needs.

Stephen-ONeil:
Each component being responsible to fetch its own data, relying on the apollo cache rather than managing our own local stores, is the pattern I suggested way back/implemented with the COVID data. I think that's the useful part. Developers writing components/components code itself doesn't need to know anything about the app's caching state or what any other parts of the code base may or may not be doing with the same data.

I'm not sure about the idea of scattering queries, or pieces of queries, throughout the client though. Please try to find an article promoting that pattern, or try and list out some benefits of it yourself.

Writing out queries in full from each component would be one thing. I could see some reasons why that would be a positive pattern. It's having central base query fragments stitched together with far-flung inner query fragments by functions stuffed full of branching conditionals which is, I think, going to be a big problem in the future. I'm sure it all makes sense and looks clean to you, having just finished writing it and having complete knowledge of both client and server implementations, but how will a dev in the future complete any of these tasks?
  1) looking at a server side log for a query, find where that query originated from client-side
  2) after updating the services schema, find and update all the corresponding client side fragments
I think they'd need to search for small fragments of what they actually want to find, hope they get lucky and those turn up in-code, and then they'd have to manually trace back a bunch of functions/do some speculative debugging to actually find the constructed queries.

That's mostly a problem with constructing queries from scattered fragments. My approach shares some of those problems, _but_ in the worst case there's only one file per subject to look in client side _and_ all final queries are identified by searchable names (both in logging and where imported by other modules).

Maybe a middle ground would be to write, fully inline, all queries where ad as needed. Easier than either of our current approaches to search for query strings without being foiled by templating... but otherwise a mess of repetition and possible de-syncs. Could still be a big mess to make schema changes with.

As for the question of writing totally use-case-specific queries vs writing more general queries, in the best case the former does provide some potential room for optimization... but only at the level of per-request transfer size. It has the potential to actively reduce the effectiveness of both Apollo's in-session caching and our cross-session GET request caching (admittedly, already a little compromised by query batching). Either way, my proposed approach of centralizing all query+resolver pairs for a given topic to one module in `/models/<topic>` doesn't mean you can't write a ton of highly scoped queries that are each only used in one place (although that'd be a bit awkward), it just means that there'd be only a single actual source-code interface between the server and client per topic, and that it'd be easier to enforce standards on them and do things like review the actual in-practice effectiveness of the schema, etc.

PR #1005, opened 2021-04-09
GraphQL RPB (service inventory only for now)

ktw1016:
for discussion roughly as all the details are temporary to discuss if this is the way we should do it for graphql rpb

Copied from #973

### DIFF
--- a/client/src/InfoBase/bootstrapper.js
+++ b/client/src/InfoBase/bootstrapper.js
@@ -13,6 +13,7 @@ import {
 } from "connected-react-router";
 import _ from "lodash";
 import { createHashHistory } from "history";
+import { ApolloProvider } from "@apollo/client";
 import React from "react";
 import ReactDOM from "react-dom";
 import { Provider } from "react-redux";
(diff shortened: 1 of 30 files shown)


Stephen-ONeil:
Keeping this as a branch off of the apollo hooks branch is basically a full on blocker. I'd suggest either decoupling this (cherry picking over to master and dropping the reliance on hook based loading), or dropping this branch for now and just focusing on hooks.

Regardless, if you want discussion on this, please be the one to kick it off. I think we were on the same page with the initial issue/sketch of an idea discussed in the dev meetings. Only you've spent any amount of time developing those ideas since so balls in your court if you want to have a conversation about the approach/architecture!

ktw1016:
Will come back to this once Apollo hooks, batching, summary data is stable

Stephen-ONeil:
> Will come back to this once Apollo hooks, batching, summary data is stable

Isn't batching done? Is there something in apollo hooks that's going to upset things?

Do you expect that to be soon? This PR existing is going to keep anyone else from taking a stab at putting GraphQL data in to the RPB their own way, if it's deep in your stack then I'd say give it the axe.

ktw1016:
With multiple queries (using hooks), batching is broken. I can close this one out though, I can replicate the changes if needed later on

PR #1006, opened 2021-04-12
initial basic concept of summary data for services_types for gov

ktw1016:


server/src/models/services/schema.js
```diff
@@ -11,6 +14,15 @@ const schema = `
     services: [Service]
     has_services: Boolean
   }
+  type ServiceSummary{
+    id: String
+    service_type_summary: [ServiceTypeSummary]
```
ktw1016, line 19: Idea is basically extending this summary per panel

server/src/models/services/schema.js
```diff
@@ -101,8 +113,26 @@ export default function ({ models, loaders }) {
...
+        value,
+      }))
+      .value();
+  };
 
   const resolvers = {
+    Gov: {
+      service_summary: (_x, __x, { lang }) => ({
+        id: "service_summary",
+        service_type_summary: get_service_type_summary(lang),
```
ktw1016, line 133: Could performing this inside a resolver instead of `populate` be a performance bottleneck? It's pretty annoying to compute this inside populate because services data is mostly bilangual string.

AlexCLeduc: I can see how doing everything in populate would be annoying. 

At the very least, using dataloaders would help, though. following this pattern you're calling`Model.find()` per aggregation type (you've only got one right now, so it's not a huge deal) but you've got other aggregation types to do (interaction points, top 10 services, etc.) and probably don't want to fetch every single service 6 times per infographic. You've already got the org_id based loader, you can add a government dataloader too. 

Doing this stuff in populate would be safer, though. Is everything really just en/fr strings? In titan we made an effort to give all the lookups a code-friendly identifier called `code`, because sometimes we need to build rules that get triggered when a service is external. We should probably be adding those to the pipeline

client/src/panels/panel_declarations/services/services_digital_status.js
```diff
@@ -41,26 +36,28 @@ const colors = scaleOrdinal().range([
...
-      key,
-      [can_online]: get_current_status_count(key, true),
-      [cannot_online]: get_current_status_count(key, false),
-      [not_applicable]: get_current_status_count(key, null),
-    }))
-    .sortBy(can_online)
-    .value();
+  const processed_data = _.map(data, (row) => ({
+    ...row,
+    key_desc: text_maker(row.key_desc),
```
ktw1016, line 56: this is where key to text conversion happens

server/src/models/services/populate.js
```diff
@@ -166,11 +181,110 @@ export default async function ({ models }) {
...
+    .map(({ code, en, fr }) => [code, { en, fr }])
+    .fromPairs()
+    .value();
+  const gov_type_summary = _.chain(service_rows)
+    .flatMap("service_type_code")
+    .countBy()
+    .map((value, type_code) => ({
+      id: `${type_code}_${value}`,
+      subject_id: "gov",
+      label_en: service_types_lookup[type_code].en,
```
ktw1016, line 204: this is where I needed that lookup @AlexCLeduc (I made my own csv, not coming from pipeline), is there a way to avoid having it?

server/src/models/services/populate.js
```diff
@@ -166,11 +181,110 @@ export default async function ({ models }) {
...
+  const get_current_status_count = (services, key, value) =>
+    _.countBy(services, `${key}_status`)[value] || 0;
+  const populate_digital_summary_key = (services, subject_id, level, key) => ({
+    id: `${level}_${subject_id}_${key}`,
+    key_desc: `${key}_desc`,
+    key,
+    subject_id,
+    can_online: get_current_status_count(services, key, true),
+    cannot_online: get_current_status_count(services, key, false),
+    not_applicable: get_current_status_count(services, key, null),
```
ktw1016, line 250: (1) Potential red flags: these `key_desc`, `can_online`, `cannot_online`, `not_applicable` are keys that is expected and assumed to be translated over in client side. Should this be taken care of on the server side?

server/src/models/services/populate.js
AlexCLeduc: you can compute the lookup during populate, something like

```js
_.chain(service_rows)
    .uniqBy("service_type_code")
    .map(({service_type_code,service_type_en,service_type_fr}) => [
        code, {en: service_type_en,fr: service_type_fr}
     ])
    .fromPairs()
    .value()
```

If the lookup is many-to-many, with `<>` or `,` separated list, it's a little more complicated, maybe something like

```js
_.chain(service_rows)
    .map(s => _.pick(s, ['service_type_code', 'service_type_en', 'service_type_fr']) )
    .uniqBy("service_type_code") // optional first uniq grouping to cut down on processing
    .flatMap( ({service_type_code,service_type_en,service_type_fr})  => {
        const groups = _.zip(
          service_type_code.split(","),
          service_type_en.split(",")
          service_type_fr.split(",")
        )
        return groups.map( ([code, en, fr]) => ({ code,  en, fr }) )
     })
    .uniqBy("code")
    .map(({code,en,fr}) => [code, {en, fr }])
    .fromPairs()
    .value()
```

This works if the codes, en and fr values are all sorted consistently. I've checked that this is the case, and I'll tell Ross to ensure it is always the case with other lookups that have their codes added.

AlexCLeduc: Honestly, though, having a json file of every single lookup isn't a bad idea either, something like 

```json
ServiceType: {
  "acq": { "en": "Acquisition Services", "fr": "Services d'acquisition" },
  ...
}
StandardType: {
  "access": { "en": "Access", "fr": "Accès"},
  ...
}
```

It should be easy for titan to spit out a json file like this alongside the big excel sheet

server/src/models/services/populate.js
AlexCLeduc: I think doing that on the client makes more sense

data/service_types_lookup.csv
```diff
@@ -0,0 +1,18 @@
+code,en,fr
```
ktw1016, line 1: This was created manually by me, I think it'd be much cleaner to have a lookup file from pipeline (json or csv) rather than trying to programatically create one (since it's many-to-many so it has to be assumed to be sorted correctly)

let me know any thoughts

server/src/models/services/populate.js
```diff
@@ -167,10 +202,362 @@ export default async function ({ models }) {
...
+  const dept_top_website_visits_summary = _.chain(service_rows)
+    .groupBy("org_id")
+    .flatMap(get_top_website_visits_summary)
+    .thru(post_process_website_visits)
+    .value();
+  const program_top_website_visits_summary = _.chain(service_rows)
+    .reduce(group_by_program_id, {})
+    .flatMap(get_top_website_visits_summary)
+    .thru(post_process_website_visits)
+    .value();
```
ktw1016, line 519: All of the above computations could be broken up into each of files.. let me know any suggestions on organization

AlexCLeduc (approved):
Nice, this is way faster! The server's populate code is getting long and complex, but I don't have any better ideas at the moment. 

Ross' changes to the GCSI extract were merged into titan. The lookup CSVs seem a little bit like overkill, I suggest we start exporting a JSON file from titan with all the lookups. 

```json
"ServiceType": {
  "acq": { "en": "Acquisition Services", "fr": "Services d'acquisition" },
  ...
}
"StandardType": {
  "access": { "en": "Access", "fr": "Accès"},
  ...
}
```

I've opened a [titan issue](https://github.com/TBS-EACPD/results/issues/1416) about it, so let's keep that in mind going forward. We'll probably keep the `,`-separated en/fr values in the data despite having a new lookup, at least for a while.

PR #1007, opened 2021-04-12
Line graph overlapping segments with dashes

### LINKED ISSUE #557: Make line graphs smart about overlapping line segments
I've got some code in the `auth_exp_planned_spending` panel that makes it automatically style overlapping line segments as dashed. I think it's prime for hoisting up to the `NivoResponsiveLine` component as a default setting. Only thing lacking right now is to differentiate the dash array value when more than two line segments overlap (cut that corner in the `auth_exp_planned_spending` implementation since that graph only has two series that could potentially overlap).

GarvinH:
closes #557 

Handle overlapping segments for all line graphs and change code to support overlaps of more than 3 lines

### Live case
[Dev link](http://dev.ea-ad.ca/master/index-eng.html#orgs/dept/186/infograph/financial)
[Local](http://localhost:8080/build/InfoBase/index-eng.html#orgs/dept/186/infograph/financial)
image

### Test case
[Local](http://localhost:8080/build/InfoBase/index-eng.html#orgs/dept/326/infograph/financial)
image

### DIFF
--- a/client/src/charts/wrapped_nivo/WrappedNivoLine.js
+++ b/client/src/charts/wrapped_nivo/WrappedNivoLine.js
@@ -75,7 +75,6 @@ export class WrappedNivoLine extends React.Component {
       enableGridY,
       stacked,
       markers,
-      layers,
       disable_table_view,
       table_name,
       table_first_column_name,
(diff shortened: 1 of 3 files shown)


Stephen-ONeil (approved):
👍  rad. 

Occured to me, we should add storybooking for the charts to our TODOs (at least for the special features we've added on top of Nivo, can leave everything else to Nivo's own storybooks).

PR #1008, opened 2021-04-12
Table of contents cleanup

Stephen-ONeil:
Some refactoring, things I would have brought up if I'd been here last week. Let me know if you want me to go deeper on the "why" for any of these changes.

### DIFF
--- a/client/src/components/PinnedContent/PinnedContent.js
+++ b/client/src/components/PinnedContent/PinnedContent.js
@@ -101,7 +101,7 @@ export class PinnedContent extends React.Component {
           <InView>
             {({ inView, ref, entry }) => {
               const is_stickied =
-                !this.is_pinned &&
+                this.is_pinned &&
                 !inView &&
                 entry &&
                 entry.boundingClientRect.top < 0;
@@ -144,11 +144,11 @@ export class PinnedContent extends React.Component {
                             border: "none",
                           }}
                           aria-label={text_maker(
-                            this.is_pinned ? "pin" : "unpin"
+                            !this.is_pinned ? "pin" : "unpin"
                           )}
                           onKeyDown={this.handleKeyDown}
                         >
-                          {!this.is_pinned ? (
+                          {this.is_pinned ? (
                             <IconPin
                               height="25px"
                               width="25px"
--- a/client/src/components/PinnedContent/PinnedContent.js
+++ b/client/src/components/PinnedContent/PinnedContent.js
@@ -100,26 +100,27 @@ export class PinnedContent extends React.Component {
         {({ width }) => (
           <InView>
             {({ inView, ref, entry }) => {
-              const is_stickied =
+              const should_pin =
                 this.is_pinned &&
                 !inView &&
                 entry &&
                 entry.boundingClientRect.top < 0;
 
               return (
-                // this div with ref is for the intersection check
                 <div ref={ref}>
-                  {/* this conditional div acts as a placeholder to make scrolling smoother */}
-                  {/* before adding this, stickying by "position: fixed" would take it out of the DOM block
-                  flow which would bump up the window due to the total block content being shortened.
-                  By adding this placeholder div, the height of the total block content remains the same,
-                  thus no longer causing the window to jump  */}
-                  {is_stickied && <div style={{ height: content_height }} />}
+                  {/* 
+                    this conditional div with height height: content_height acts as a placeholder to make scrolling
+                    smoother. Before adding this, stickying by "position: fixed" would take it out of the DOM block
+                    flow which would bump up the window due to the total block content being shortened.
+                    By adding this placeholder div, the height of the total block content remains the same,
+                    thus no longer causing the window to jump
+                  */}
+                  {should_pin && <div style={{ height: content_height }} />}
                   {/* this div is for sticky styline, must be flex to include margins onto height */}
                   <div
                     style={{
                       display: "flex",
-                      ...(is_stickied && {
+                      ...(should_pin && {
                         position: "fixed",
                         top: 0,
                         zIndex: 2001,
@@ -148,7 +149,7 @@ export class PinnedContent extends React.Component {
                           )}
                           onKeyDown={this.handleKeyDown}
                         >
-                          {this.is_pinned ? (
+                          {!this.is_pinned ? (
                             <IconPin
                               height="25px"
                               width="25px"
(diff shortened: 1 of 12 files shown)


client/src/components/PinnedContent/PinnedContent.js
```diff
@@ -84,24 +84,6 @@ class _PinnedContent extends React.Component {
...
-      match: {
-        params: { options: prev_panel_link },
-      },
-    } = prev_props;
-
-    !prev_panel_link &&
-      panel_link &&
-      this.setState({ user_has_enabled_pinning: false });
-  }
-
```
GarvinH, line 104: Not sure if it's just me, but the pinned content now inconsistently decides to pin itself when using the ToC?

client/src/infographic/TableOfContents.js
```diff
@@ -2,105 +2,68 @@ import _ from "lodash";
...
-      initial_open: !prev_state.initial_open,
-    }));
-  };
-
+  constructor(props) {
+    super(props);
+    this.state = {
+      is_open: false,
+    };
+  }
```
GarvinH, line 23: Just wondering why we prefer to do it like this.

(I'd assume it's because of consistency, and it allows us to add props into the initial state though?)

client/src/infographic/TableOfContents.js
```diff
@@ -2,105 +2,68 @@ import _ from "lodash";
...
 export default class TableOfContents extends React.Component {
-  state = {
-    initial_open: false,
-  };
-
-  handle_click = () => {
-    this.setState((prev_state) => ({
-      initial_open: !prev_state.initial_open,
-    }));
-  };
```
GarvinH, line 27: I might've asked about this before but:

Since state change is asynchronous, wouldn't it be preferable to be using the `prev_state` instead of directly accesing `this.state`? Although, I guess it's next to impossible for this to fail considering that there's no other state change called in this component haha

client/src/components/PinnedContent/PinnedContent.js
ktw1016: Hmm I've been trying and never happened for me

ktw1016 (approved):
👍

client/src/components/PinnedContent/PinnedContent.js
Stephen-ONeil: Not inconsistent (might appear to be if you're editing the url without reloading vs with reloading). I did drop the functionality of disabling pinning when using the table of contents though.

Was weighing taking the time to make a non-hacky (or at least less hacky) way for other components to disable pinned content, since that was a pre-existing TODO in `PinnedContent.js`. Could circle back to that in this PR and bring back that table of content feature.

GarvinH: Oh I know what's causing it

If you scroll down so that the pinned content enters sight, then select a panel to go to, it will keep itself as stickied.
But if you don't have it in sight then click on a panel, it won't sticky.

client/src/infographic/TableOfContents.js
Stephen-ONeil: My mental model is as follows (it's been a while since I fact checked it though, I could be wrong, bullets starting with ?'s are things I'm going to go try and confirm after this):
  - `setState` calls are applied async, yes
  - React will, timing dependant, batch multiple `setState` calls together
  - the batched `setStates` only result in one change to the actual state (`this.state`)
  - ? but as each is applied they will create a new "pending" state (need to double check)
  - ? this "pending" state is what's actually passed as the `prev_state` arg to any callback `setState`s in the batch (need to double check)
  
The shakeout of that, assuming I'm correct, is a little tricky. It's going to be case dependant, and possibly not clear, on whether to use a callback with `prev_state` or directly access `this.state`. In this case though, I think any user interaction with a `Details` drawer that is rendered as open  (i.e. `this.state.is_open === true`) should ONLY be able to result in it closing and vice-versa. Updating the state based on the _actual_ state value forces that (even if multiple `setState`s end up in the same batch). Using the (incrementally updated/pending) `prev_state` value would cause the ultimate state to appear unchanged if an even number of calls end up in the same batch.

Make sense? I'll get back to you if it turns out those assumptions I listed were wrong though.

client/src/infographic/TableOfContents.js
Stephen-ONeil: Pretty much. Consistency, and the minimal number of line changes needed if props need to factor in to the initial state in the future.

Same idea as adding trailing commas to array/object definitions (at least multi-line ones). They're optional, but the fiddly little reason for linters/formatter rules to always enforce them is so that a future commit adding a new item to the array will only touch the new line and won't have to touch an old line just to add a new comma. This is most specifically for the sake of diffs and git blames, cutting down on the noise and keeping it just to the substantial change.

client/src/infographic/TableOfContents.js
GarvinH: Yea I think that makes sense 👍

client/src/infographic/TableOfContents.js
GarvinH: Makes sense, thanks 👍

client/src/infographic/TableOfContents.js
Stephen-ONeil: Everything seems to check out except my assumption that batching is mostly timing dependant. Haven't dug all the way down on how things specifically work in React 17, but all `setState` calls are batched and in at least some cases that batching is based on the scope they're kicked off from (e.g. all `setState` calls within the same react DOM event callback go in the same batch). If there's any exceptions/specific cases with dependence on timing, it's probably introduced by schedulers like `setTimeout`, `_.debounce`, etc.

Stephen-ONeil:
See 54a0aa9, `PinnedContent` rework; not strictly a refactor, some slight behavioural changes.

GarvinH:
The `is_pinned` state wasn't matching what was being shown by the DOM, but it was a small fix so I went ahead and did it haha

GarvinH (approved):
👍

Stephen-ONeil:
> The `is_pinned` state wasn't matching what was being shown by the DOM, but it was a small fix so I went ahead and did it haha

Aha, oops, I renamed the state a couple times, always leads to miss-matches.

Stephen-ONeil:
Actually, it's broken now. The setter on scroll to panel is doing the opposite of what we'd wanted. I think we've got opposite ideas about what pinned means. Mostly to disambiguated it from `is_stickied`, I think I'd switched it from what it used to mean ("pinned to top of screen") to "pinned to top of infographic". I'll sort it out to standardize on what it used to be ("pinned to top of screen").

Stephen-ONeil:
Wait, the icon might be reversed from the internal meaning on master. The since it's a button, the icon is supposed to display what will happen when it's clicked, not what its current state is.

PR #1015, opened 2021-04-15
remove equality from package.json

### LINKED ISSUE #931: Drop explicit @wry/equality dependency (in a month or so)

ktw1016:
closes #931


Stephen-ONeil (approved):
👍  should be good at this point

PR #1017, opened 2021-04-15
Fix styling for graphs with one data row in sobjs detailed spending

### LINKED ISSUE #1013: Sobjs Bar chart height abnormally small when single bar
image

image

GarvinH:
closes #1013

### DIFF
--- a/client/src/panels/panel_declarations/finances/detailed_program_spending_split/detailed_program_spending_split.js
+++ b/client/src/panels/panel_declarations/finances/detailed_program_spending_split/detailed_program_spending_split.js
@@ -348,6 +348,7 @@ class DetailedProgramSplit extends React.Component {
                 style={{
                   display: "block",
                   margin: "10px auto",
+                  width: "100%",
                 }}
                 className="form-control"
               />
@@ -371,7 +372,7 @@ class DetailedProgramSplit extends React.Component {
                   bottom: 30,
                   left: 215,
                 }}
-                graph_height={divHeight}
+                graph_height={divHeight > 100 ? divHeight : "90px"}
                 colors={(d) => colors(d.id)}
                 bttm_axis={{
                   tickSize: 5,
--- a/client/src/panels/panel_declarations/finances/detailed_program_spending_split/detailed_program_spending_split.js
+++ b/client/src/panels/panel_declarations/finances/detailed_program_spending_split/detailed_program_spending_split.js
@@ -372,6 +372,10 @@ class DetailedProgramSplit extends React.Component {
                   bottom: 30,
                   left: 215,
                 }}
+                // HACKY: ideally we'd not be using a ternary to set a minimum height
+                // but the library used by nivo sets unpredicatble heights
+                // which for some reason causes graphs with only one bar to be a line instead
+                // TODO: refactor this
                 graph_height={divHeight > 100 ? divHeight : "90px"}
                 colors={(d) => colors(d.id)}
                 bttm_axis={{


client/src/panels/panel_declarations/finances/detailed_program_spending_split/detailed_program_spending_split.js
```diff
@@ -371,7 +372,7 @@ class DetailedProgramSplit extends React.Component {
                   bottom: 30,
                   left: 215,
                 }}
-                graph_height={divHeight}
+                graph_height={divHeight > 100 ? divHeight : "90px"}
```
Stephen-ONeil, line 375: Took a look, `divHeight` is used for two things, right? `graph_height` and also `legendOffsetY`. In the case that `divHeight <= 100`, isn't this desyncing `graph_height` and `legendOffsetY`? I assume they should stay in sync, seeing as they're both derived from `divHeight`.

Especially if I'm right about that, but more or less in general too, this feels more like it should be a minimum value baked in to the declaration of `divHeight`. Declaring `divHeight` a bunch of lines up and then enforcing a minimum non-locally, down where it's actually used, could obscure from future developers that the value has a lower bound.

Stephen-ONeil (changes requested):

client/src/panels/panel_declarations/finances/detailed_program_spending_split/detailed_program_spending_split.js
GarvinH: Unfortunately keeping it in sync throws the legend off even more haha 
image

Gonna keep playing around with it...

Stephen-ONeil: Strange, must be some magic in those calculations, ugh.

GarvinH: Yea this is gonna be a really annoying fix, we'd have to make the chart treat all bars consistent in height (is this even possible?) and we'd also have to go in and manually render each bar after that. But theoretically, we if could do the former, the latter would be solved automatically. I tried to use the layers prop but then I realized that the bars would probably start overlapping since i'm guessing the chart itself isn't gonna scale to fit the bars

I think this method would be a good temporary fix on the small charts. I've left a comment stating that it's hacky and left a todo to refactor it.

Stephen-ONeil (changes requested):
Might sound a bit backwards but, IMO, this isn't a big enough bug to be worth a hacky fix. 

A hacky fix might become a future bug. If nothing else it's a good indicator that there's something here we don't understand, and using things we don't understand could also indirectly make future bugs. Or maybe this is just a symptom of some bigger hidden bug and we'd just be making things worth by patching over it. Point is we can't say if we don't get what's going on. 

Not fixing this right now, on the other hand, leaves us with a known but rather minor bug. If we let it linger until someone finds a non-hacky fix, then we won't have to worry about the fix itself being a future bug and we might just learn an important quirk about these graphs in the process.

A hacky fix with a TODO comment is somewhere in between these (probably exaggerated) extremes... but let's be honest, we've got too any TODO comments and we aren't going around addressing them. It's much more likely to actually get properly fixed if we don't close the issue with this right now.

Stephen-ONeil:
Oh, just saw your latest comment. So is the a general `WrappedNivoHBar` bug on receiving a single bar, not a specific bug with the props this panel sets/calculates itself? If so, please update the original issue (or close it and start a new one referencing it) with any additional information you've uncovered!

GarvinH:
Looking at it a second time, I'm only feeling half confident about it haha

I think I've uncovered a symptom rather than the cause. I'll probably not leave a comment in the issue, but since this PR is referenced, the next person can always look back here

PR #1018, opened 2021-04-15
Append tooltip to app instead of parent

### LINKED ISSUE #1016: Glossary tooltip definition not reading on VoiceOver
Some tooltip hover over definitions are being read but some are not

GarvinH:
Without this, voice over kinda jumps all over the place once a tooltip opens up.

(at least for me it does)

### DIFF
--- a/client/src/glossary/TooltipActivator.js
+++ b/client/src/glossary/TooltipActivator.js
@@ -18,7 +18,7 @@ const create_tooltip = (node) =>
     duration: 0,
     allowHTML: true,
     hideOnClick: false,
-    appendTo: "parent",
+    appendTo: app,
   });
 
 const TooltipActivator = _.isUndefined(MutationObserver)


GarvinH:
example: 
image

I went from the definition to the next paragraph, skipping the rest of the sentence

ktw1016:
In my case, this didn't fix for me. I tried out https://atomiks.github.io/tippyjs/v5/accessibility/ (by putting `<span id="parent"> for all gl_tt handlebars), which then reads out the text by manually tabbing into it. Although when you hover over, text elements move around to resize so it's super annoying.

Others should try it out

Stephen-ONeil:
Is this #1016 or not? I think you referenced it at some point but then removed the reference. It's not clear to me if you and Taewan are discussing the same problem here (sounds like Taewan had glossary definitions not being read and you had the navigation not working as expected).

GarvinH:
I thought it would fix it, but I didn't, so I decided to remove the reference haha

But yea, the navigation wasn't working as expected for me

GarvinH:
I've confirmed with Chetan that he also has navigation troubles, and that this PR smoothens it out

Stephen-ONeil (approved):

Issue #1021, opened 2021-04-15
Storybooking components


GarvinH:
## client/src/components

- [x] Accordion
- [x] BackToTop
- [x] CheckBox
- [x] DisplayTable
- [x] HeightClipper
- [x] IconGrid 
- [x] FixedPopover
- [x] FocusLockedModal
- [x] ModalButton
- [x] StatelessModal
- Typeahead
  - [x] Basic Typeahead
  - [x] Typeahead with headers + options
- [x] Alert banner
- [x] Card background image
- [x] Card centerer image
- [x] Card left image
- [x] Card list
- [x] Card top image
- [x] Countdown
- [x] Countdown circle
- [x] Debounced Text input
- [x] Details
- [x] Dropdown menu
- [x] Fancy UL
- [x] Filter table
- [x] Footnote list
- [x] Header notification
- [x] Key concepts
- [x] labelled box
- [x] labelled table
- [x] panel
- [x] radio button
- [x] select
- [x] share button
- [x] sort direction
- [x] spinner wrapper (Different formats/sizes)
- [x] tabbed content
- [x] tab loading wrapper
- Tombstones
  - [x] Unlabelled tombstone
  - [x] Labelled tombstone
- [x] Two Level Select
- [x] Write to clipboard

Edit: This is just a general guideline. Feel free to have multiple stories for each component to show how different props can change the interface/functionality!

ktw1016:
👍😃

PR #1024, opened 2021-04-16
Create basic typeahead for storybook

GarvinH:
related #1021

### DIFF
--- a/client/src/components/Typeahead/Typeahead.stories.js
+++ b/client/src/components/Typeahead/Typeahead.stories.js
@@ -1,0 +1,113 @@
+// List.stories.js
+
+import _ from "lodash";
+import React from "react";
+
+import { memoized_re_matchers } from "src/search/search_configs.js";
+
+import { Typeahead } from "./Typeahead.js";
+
+const NA = "North America";
+const SA = "South America";
+const EU = "Europe";
+const AS = "Asia";
+
+const data = [
+  {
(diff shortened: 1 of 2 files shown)


Stephen-ONeil (changes requested):
Nice!

General advice here, when writing test\* code:
  - it's extra important to keep things orthogonal (perpendicular? same metaphor, haha)
    - to clarify that, I mean keeping the test and your code base from interfacing at more than one point (that point being the module you are explicitly testing)
    - thinking of the Story as a test, it is a noisy signal for developers if a refactor in `search_configs.js` breaks the tests on `Typeahead,js`, especially when they aren't explicitly coupled in the src code itself
  - it's far less important to keep things DRY
    - partially a takeaway from the first point, lots of repetition is necessary to keep test files from adding new coupling

  \* lumping stories in with tests, practically and conceptually close enough

Not a hard rule, in a complicated enough code base you might end up wanting reusable test patterns/utils/mocks etc... but ideally not. You don't have to want to write tests for your tests!

With that in mind, instead of importing `memoized_re_matchers` (or even whole sale copying in the code for `memoized_re_matchers`), the preference would be to write the dumbest possible filter function here. Heck, it could even just be an object keyed on a fixed set of "query" values. The behaviour of `Typeahead`, the thing we actually want to test here, should be totally `on_query` agnostic.

PR #1026, opened 2021-04-16
Replace circle ci format check with git hook

### LINKED ISSUE #975: Reconsider handling format check during CI
We should reconsider how we handle the format check going forward.

Reasons to remove it altogether:

> To be honest, I kind of question the use of a format check step in CI. I think we don't care if some random file isn't formatted right. Rebasing, find-and-replace, code mods etc can always bypass the on-save prettier run... but it doesn't matter much since the file will reformat the next time it's opened and re-saved
> 
> All we really want to enforce is that devs actually have prettier running on-save locally, right? This is flaky for that purpose, since other changes are more likely to trigger it
> If we REALLY care, the right solution is to add a global prettier check/run as a git-hook pre-push or something (edited) 
> Easy enough to do, we've already got an example of one pre-push hook in the archive branch name convention check
> ... but might be kinda annoying, having to wait for a prettier run on every push attempt

Reasons to keep it:
> - Previously the formatter check was in individual test jobs within the CI check but they would never fail for some reason. Keeping it means the CI check would run as (previously) intended
> - Format check occurs in other projects (such as zeus)
> - New devs will know right away if prettier isn't working locally
> - Wiki says non-prettier compliant code will fail in CI (If we drop it from CI, remember to update wiki!)

Summary of current choices:
- Remove it altogether
- Keep it as is
- Use git-hooks to ensure formatting pre-push

Feel free to add suggestions

GarvinH:
closes #975

### DIFF
--- a/scripts/dev_scripts/hooks/pre-push
+++ b/scripts/dev_scripts/hooks/pre-push
@@ -15,3 +15,5 @@ if [[ ($local_branch_name =~ ^__ || $local_branch_name =~ ^archived__ ) && $remo
   echo "Double underscore and archived branches can only be pushed to the archive (private) remote!"
   exit 1
 fi
+
+npm run check_format .
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,25 +112,6 @@ jobs:
           command: node ./node_modules/@lhci/cli/src/cli.js autorun --config ./scripts/ci_scripts/lighthouserc.js
 
   # Note: the image used for this job is built from a docker file found in ../dockerfiles/build
-  format_check:
-    docker:
-      - image: eaadtbs/ib-ci-build:1.3
-        <<: *dockerhub_auth
-    working_directory: "~/InfoBase"
-    steps:
-      - checkout:
-          path: "~/InfoBase"
-
-      - restore_cache:
-          keys:
-            - top-level-dependencies-{{ checksum "package-lock.json" }}
-      - run: 
-          command: |
-            [ -e "node_modules" ] ||  npm ci
-
-      - run: npm run check_format .
-
-  # Note: the image used for this job is built from a docker file found in ../dockerfiles/build
   build:
     docker:
       - image: eaadtbs/ib-ci-build:1.3
@@ -384,9 +365,6 @@ jobs:
 
 workflows:
   version: 2
-  format_check:
-    jobs:
-      - format_check
 
   test_email_backend:
     jobs:
--- a/.prettierignore
+++ b/.prettierignore
@@ -4,6 +4,7 @@ node_modules
 client/app/
 client/build/
 client/build_prod/
+client/storybook-static
 */transpiled_build/
 */coverage/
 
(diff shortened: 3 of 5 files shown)


scripts/dev_scripts/hooks/pre-push
```diff
@@ -14,4 +14,6 @@ local_branch_name=$(echo "$local_ref" | sed "s/refs\/heads\///")
 if [[ ($local_branch_name =~ ^__ || $local_branch_name =~ ^archived__ ) && $remote != "archive" ]]; then
   echo "Double underscore and archived branches can only be pushed to the archive (private) remote!"
   exit 1
-fi
\ No newline at end of file
+fi
+
+npm run check_format .
```
Stephen-ONeil, line 19: Something to check, because I don't know off the top of my head, what's the working directory of a hook? The root of the git repo, or the location of the bash shell that ran the hooked command?

Stephen-ONeil:
Just the one comment off the top of my head, will get back to reviewing this next week.

scripts/dev_scripts/hooks/pre-push
GarvinH: It checks `client`, `email_backend`, and `server`, even when I'm in the `client` directory when I run the `git push`. So I'm guessing that it checks from root

Stephen-ONeil: `check_format` is only defined in the root `package.json`, so if hooks weren't always run in the root context it'd be throwing errors instead.

Stephen-ONeil (changes requested):
Noticed that `client/storybook-static` is getting format checked. Takes a good chunk of time, and pretty surprised it's passing at all. Please update the `.prettierignore` with that, and maybe double check there's no other build directories etc. missing from there.

GarvinH:
Was there a setting you could insert into CLI that would make it print all files that it checks? There doesn't seem to be a verbose option on the prettier website, so it seems like the only way to verify if to be very quick at reading the cli.

For example, I'd imagine `local_mongo` in the `email_backend` is being checked (I'm guessing we don't need to check it too), but it's too quick to actually verify if it really is getting checked or not

Stephen-ONeil:
Might be an option for that, check the docs. If not, you can monkey patch the eslint source itself! I was just watching the command run and noticed it from the line showing the file currently being scanned.

Only checks .js files, so there's nothing in `local_mongo` to scan.

GarvinH:
Hm that's kinda interesting haha, the prettier ignore was saying how prettier seems to want to format everything, but I guess it still skips certain files/folders

Unless I'm missing some build folders in my project folder, it doesn't seem to be formatting many (if any) unnecessary files

Stephen-ONeil (approved):
👍

Stephen-ONeil:
Ok, let's merge it in and all use it for a while. We can evaluate whether it's too slow, etc, after using it for a bit before spending more time debating when/where it should run or trying to optimize it it up.

PR #1027, opened 2021-04-19
Removing flexbox_grid.scss

### LINKED ISSUE #916: Revert from custom fcol's to bootstrap col's
I just remembered why we had out own `fcol` grid system... Bootstrap 3 mostly used floats for its grid layout, and we wanted flexbox. Bootstrap 4 uses flex for its column system though!

If I'd remembered, this potentially could have been part of the bootstrap 4 migration. On the other hand, I'm sure it won't be entirely straight forward, there's lost of differences between the two and we may have lots of fiddly little style rules spread about that could be tied to the current `fcol` implementation. Would require a lot of careful checking.

Note that this would essentially be a hard-decision to stick to using bootstrap. The fact that we were mostly switched to an alternate grid system was a big part of why dropping bootstrap altogether was (and is) an option for us. Just something to keep in mind.

GarvinH:
closes #916 

I also removed all the other styles in the `flexbox_grid.scss` in favour of the bootstrap equivalents and removed the file


Stephen-ONeil (changes requested):
Found a bug pretty quick. Please fix then spot test the site at various widths your self, could be other issues.

image

GarvinH:
Turns out bootstrap and our breakpoints name is differently so I'm gonna have to re-do all of this 🤦‍♂️
(we have extra-small, and bootstrap has extra-large). But at least it will be consistent!

client/src/common_css/flexbox-grid.scss
```diff
@@ -1,988 +0,0 @@
...
-  .fcol-lg-offset-9,
-  .fcol-lg-offset-10,
-  .fcol-lg-offset-11,
-  .fcol-lg-offset-12 {
-    box-sizing: border-box;
-    -webkit-box-flex: 0;
-    -ms-flex: 0 0 auto;
-    flex: 0 0 auto;
-    padding-right: 0.5rem;
-    padding-left: 0.5rem;
```
Stephen-ONeil, line 795: I think we lost this general padding, might be places where we counted on it in our designs.
image

Probably want to bring it back for bootstrap cols in `bootstrap-wet-fixes-extensions.scss`

client/src/common_css/flexbox-grid.scss
```diff
@@ -1,988 +0,0 @@
...
-    max-width: 100%;
-  }
-
-  .fcol-md-offset-0 {
-    margin-left: 0;
-  }
-
-  .fcol-md-offset-1 {
-    margin-left: 8.33333333%;
-  }
```
Stephen-ONeil, line 657: Did a random spot check of dropped class names, found this one being used in `tags_related_to_subject_panels.js` (don't see that being updated in this PR). Please double check for any deprecated `flexbox-grid.scss` classes that you may have missed, either consider recreating them in `bootstrap-wet-fixes-extensions.scss` or refactoring the layouts that used them.

Stephen-ONeil (changes requested):

client/src/common_css/flexbox-grid.scss
Stephen-ONeil: Wait, scratch that, `tags_related_to_subject_panels.js` was using the col version, not fcol. Weird, thought we'd switched all cols to fcols during the bootstrap 4 upgrade.

client/src/common_css/flexbox-grid.scss
GarvinH: `.row` has a negative margin, so I just removed those margins instead. Should have about the same impact I think?

Stephen-ONeil: I took a closer look, I guess our `frow`s were using -0.5rem margins for some reason (computed value of -5px). Bootstrap `container`s have 15px (and their `row`s have -15px), so our `frow`s all effectively had a 10px margin.

Looks like the -15px margins are important to bootstraps grid layout concept though ([link](https://stackoverflow.com/questions/51282923/bootstrap-negative-margin-on-rows)). Since we're committing to Bootstrap, we shouldn't be messing with that. Proper solution would be to manually review and add back in util classes to explicitly add the desired padding on rows (like the above home page example) where the padding is important for the site to look good. I'd guess that most grid system content doesn't actually look all that different with the implicit padding of the `frow`s effectively removed.

Stephen-ONeil (changes requested):
See response in comment chain above.

client/src/charts/canada/canada.js
```diff
@@ -200,8 +200,8 @@ export class Canada extends React.Component {
       })
     );
     return (
-      <div className="frow no-container">
-        <div className="fcol-md-3">
+      <div className="row no-container">
```
Stephen-ONeil, line 203: Pretty sure `.no-container` doesn't exist anymore, it was from `flexbox-grid.scss`. Did you double check the classes tossed out with the rest of `flexbox-grid.scss` and clean them up as needed?

Stephen-ONeil (changes requested):
I think it's all good, caught one deprecated classname though so just want to make sure you've double checked for those. Site all looks good, so should be very close to approved 👍

client/src/charts/canada/canada.js
GarvinH: Whoops, I think this was the only one I accidentally skipped haha

I did check all the other one's during the first commit, so I doubt there's another one that managed to sneak past me.
The only one i would imagine that got past is `reverse` since that one is used all over the project with different contexts (ex: the reverse method). But I triple checked the find function on VSCode, so I'm pretty sure there was none

Stephen-ONeil (approved):
👍

PR #1028, opened 2021-04-19
Add no-throw-literal lint rule

Stephen-ONeil:
Pretty simple, mostly for consistency (but thrown literals can also produce more opaque error messages in older browsers).

### DIFF
--- a/client/src/core/ErrorBoundary.js
+++ b/client/src/core/ErrorBoundary.js
@@ -65,7 +65,7 @@ export class ErrorBoundary extends React.Component {
       testing_for_stale_client: false,
     });
 
-    throw this.state.error;
+    throw new Error(this.state.error);
   }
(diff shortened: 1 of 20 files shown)


GarvinH (approved):
👍

client/src/core/ErrorBoundary.js
```diff
@@ -65,7 +65,7 @@ export class ErrorBoundary extends React.Component {
       testing_for_stale_client: false,
     });
 
-    throw this.state.error;
+    throw new Error(this.state.error);
```
GarvinH, line 68: Wait I think this one is throwing Error(Error())? But I don't think that's a big deal

Stephen-ONeil: 👍

PR #1029, opened 2021-04-19
Typescript in client

AlexCLeduc:
Sort of working off of #803

- [x] builds work
- [x] VSCode problems matches the webpack output
- [x] storybook works
- [x] ES features (import, spread, etc.) work
    - [ ] have not tested decorators which are a special babel opt-in

I included a few example modules that were converted in #803, I'll have nested PRs for other examples. The Diffs tend to get huge because `import "x.js"` has to become `import "x"`

## How to review this

- check that vscode still highlights stuff appropriately, in js and in TS files
- check that you can build stuff locally
- look at your network tab and make sure I haven't blown up any bundles (or does CI do this?)


client/package.json
```diff
@@ -33,6 +39,7 @@
     "clipboard-polyfill": "^2.8.1",
     "connected-react-router": "^6.5.0",
     "core-js": "^3",
+    "crypto-browserify": "^3.12.0",
```
AlexCLeduc, line 42: this package, along with `stream-browserify`, are required by storybook. `stream` and `crypto` are node std-lib packages for which webpack 4 automatically provided polyfills. Webpack 5 doesn't include those polyfills, hence the need to get a browserify'd version.

client/package.json
```diff
@@ -121,12 +133,13 @@
...
+    "@storybook/addon-a11y": "^6.2",
+    "@storybook/addon-actions": "^6.2",
     "@storybook/addon-console": "^1.2.3",
-    "@storybook/addon-essentials": "^6.1.21",
-    "@storybook/addon-links": "^6.1.21",
-    "@storybook/react": "^6.1.21",
+    "@storybook/addon-essentials": "^6.2",
+    "@storybook/addon-links": "^6.2",
+    "@storybook/builder-webpack5": "^6.2",
+    "@storybook/react": "^6.2",
```
AlexCLeduc, line 142: storybook 6.1 uses webpack 4, which had [compatibility problems](https://github.com/storybookjs/storybook/issues/14157) with `ts-loader`. After adding `@storybook/builder-webpack5` and upgrading the others to 6.2 (and adding some polyfills above), I got storybook to work.

client/src/types/global.d.ts
```diff
@@ -0,0 +1,17 @@
+declare const IS_A11Y_MODE: boolean | undefined;
```
AlexCLeduc, line 1: Looks like we have yet another list of globals to maintain 😞 

vscode's eslint is still acting up on this file, claiming these are all unused variables. I tried adding `*.d.ts` to eslintignore but it's still complaining.

client/src/types/types.d.ts
```diff
@@ -0,0 +1,12 @@
+//TODO: figure out why this file isn't parsing in VSCode but still successfully prevents type errors when importing yaml
+interface TextBundle {
```
AlexCLeduc, line 2: Yaml and CSV imports all follow these types. It's not perfect, we'll probably have more work to do once we convert `text.js`

client/tsconfig.json
```diff
@@ -0,0 +1,35 @@
+{
```
AlexCLeduc, line 1: This file is super fragile, if you touch it, you'll need to test a bunch of stuff and look for VSCode problems

client/deploy_scripts/build_all.sh
```diff
@@ -39,6 +39,9 @@ function print_captured_output {
 
 
 scratch=$(mktemp -d -t captured_build_output.XXXXXXXXXX)
+
```
AlexCLeduc, line 42: TODO:  remove this, test more concurrent scenarios

client/.eslintignore
```diff
@@ -1,3 +1,4 @@
 patches/*
 build/*
-build_prod/*
\ No newline at end of file
+build_prod/*
+/**/*.d.ts
```
AlexCLeduc, line 4: this ignore might not actually work 🤔

Stephen-ONeil:
> ## How to review this
...
> * look at your network tab and make sure I haven't blown up any bundles (or does CI do this?)

http://dev.ea-ad.ca/typescript2/bundle-stats.html

Worst change is probably the bootstrapper
image

Note, might want to rebase off master then re-check the bundle stats. The base point in that comparison (right column) is the current master state when this branch last went through CI. Could be a bit misleading based if this branch wasn't up to date with master at that point.

Stephen-ONeil:
Hmm, ran an `npm ci` in client and restarted by `npm run IB_q`, but the resulting build was complaining about the use of `interface` in `TextMaker.tsx` ("interface is a reserved keyword..." error, and a suggestion that some loader was missing). Poked around and nothing looked wrong, tried double checked I'd run `npm ci` and restarted the build process/tried an `IB_q_both` and... at some point it was suddenly working 🤷 

Haven't tried to recreate, unclear what was wrong or where to start. Maybe the old webpack process from my previous branch ran right when I checked out the typescript branch and poisoned the webpack cache? Working now anyway, just a heads up for anyone else who tries checking this out locally.

Stephen-ONeil:
Built in language support/code highlighting works 👍 

Needed to restart the eslint plugin before it'd load the new typescript plugins, so all good there.

client/src/components/TextMaker.tsx
```diff
@@ -24,16 +34,25 @@ const TextMaker = ({
...
-
+interface TMProps {
+  k: string;
+  el?: string;
+  args?: any;
+  tmf?: any;
+  style?: any;
+  className?: string;
+  template_str?: any;
+}
```
Stephen-ONeil, line 45: Shouldn't these types be the same as `TextMakerProps` (with short hand names)? Any reason for the `any`s here?

Stephen-ONeil (approved):
Awesome, thanks for chewing through all these config issues!

Everything seems to work, and totally on board for approving this as the first step in an incremental process. 

Couple questions here, and one as a comment below, that don't necessarily need to get in the way of this merging:
  - on that earlier comment about bundle size, any idea why the bootstrapper js would grow a bunch?
  - would you have anything to start noting down in the docs on typescript migration? Gotchas, or some of the things you left warning comments on from this PR?

client/src/components/TextMaker.tsx
AlexCLeduc: Good point, fixed

PR #1030, opened 2021-04-19
Client package cleanup

Stephen-ONeil:
Noticed some vestigial packages in there. Managed to identify a handful that could be dropped. Decided to clean it up in general while I was at it.


ktw1016 (approved):
👍

PR #1031, opened 2021-04-20
Stop pinned content from accessing null ref in text-only

GarvinH:


Stephen-ONeil (approved):
Approved, although I think the difference between the standard and a11y version could be made easier to see at a glance.

What if the exported component was a wrapper that just checked `is_a11y_mode` and then fell through to either rendering the children directly or rendering an inner version with the actual pinning behaviour? Then it would be the most obvious that none of the state, etc, is used when in text-only mode.

GarvinH:
Hm... Sounds like a pretty good idea! I'll give it a go

Stephen-ONeil (approved):
👍

PR #1033, opened 2021-04-21
Use Husky instead of init hooks

### LINKED ISSUE #1032: Consider managing git hooks with a node package
E.g. https://www.npmjs.com/package/husky

Currently, there's an `init_hooks` script in the root `package.json`, it symlinks `/scripts/dev_scripts/hooks/` to `.git/hooks` and adds a merge/rebase hook that reapplies the symlinking (to catch new hooks in the future). This has worked consistently, but it does require the one initial setup step.

Something like husky would remove that initial manual step, should guarantee all developers have the expected hooks.

GarvinH:
closes #1032


ChetanMandur:
This only seems to work after running `npm ci`. Without running this, Husky will never install/work.

package.json
```diff
@@ -9,7 +9,8 @@
     "init_hooks": "sh scripts/dev_scripts/init_hooks.sh",
```
Stephen-ONeil, line 9: Delete this

.husky/pre-push
```diff
@@ -1,3 +1,7 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+
 #!/bin/sh
```
Stephen-ONeil, line 5: This only means anything if it's the first line of the file (in which case it hints what shell to use if executing the file without being explicit).

.husky/pre-push
```diff
@@ -1,3 +1,7 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
```
Stephen-ONeil, line 2: What's the story on this and the `_` dir? Did a ctrl + f in the docs and didn't see an explanation.

Stephen-ONeil (changes requested):
Couple minor notes and a question

Stephen-ONeil:
> This only seems to work after running `npm ci`. Without running this, Husky will never install/work.

Running `npm ci` in the root directory is necessary for a lot of the dev environment to work. Much harder to miss than running the previous separate `init_hooks` command.

Stephen-ONeil:
If we don't manually clean up our `.git/hooks` dirs, does this result in hooks running twice? If so, make sure everyone's aware of that (and to re-run `npm ci` in root) before this merges.

.husky/pre-push
GarvinH: Whoops, blindly copy pasted haha

.husky/pre-push
GarvinH: So that line gets generated automatically when using their `husky add` command when creating a new hook.

For the `_` dir, its added automatically on the initial install (`npm ci`)
image

GarvinH:
> Running npm ci in the root directory is necessary for a lot of the dev environment to work. Much harder to miss than running the previous separate init_hooks command.

Yea, just important to have all the branches rebased and `npm ci`'d if `.git/hooks` gets deleted.

Yup! Will make sure to let everyone know about the .git/hooks

Stephen-ONeil (approved):
👍

PR #1037, opened 2021-04-23
statically type business constants

AlexCLeduc:
another example of a PR that migrates stuff to TS. We'll merge `typescript2` first, and then I'll cherry-pick this over the resulting master 

I removed the `businessConstants.yaml` and put it all into the TS file because
1. there weren't any multi-line strings or handlebar/markdown transforms, so there's no syntactic advantage of using yaml
2. Because translations were in nested entries, the yaml-lang-loader wasn't excluding the other language from bundles anyway

Note that I statically typed businessConstants' dependencies as well. Although there may be small benefits in sometimes statically typing files without their dependencies, we'd still have to manually declare typings for those dependencies. It's therefore more efficient to target the "leaves" of our dependency tree for conversion first, so we avoid having to type things multiple times.


client/src/models/businessConstants.ts
```diff
@@ -0,0 +1,723 @@
...
+interface BusinessConceptEntrySet {
+  [key: string]: BusinessConceptTextValue;
+}
+
+function normalizeBilinguals(
+  entries: RawBusinessConceptEntrySet
+): BusinessConceptEntrySet {
+  return _.mapValues(entries, (entry) => getNormalizedEntry(entry));
+}
+
```
AlexCLeduc, line 42: Everything outside this block is basically noise. What's going on here is that we have two ways of defining a business constant, either bilingual `{en, fr}` dicts or `{text}` dicts. This code deals with converting both types to `{text}` dicts.

Note the use of a common TS pattern of checking properties via `hasOwnProperty` and the use of the `as` keyword.

There's a good chance there's a cleaner pattern I could have used here. [these docs](https://www.typescriptlang.org/docs/handbook/2/narrowing.html) are relevant. This PR is a bit of a conflict magnet, though, so I'd say we can explore this internal refactor after merging.

docs/js-tips-style-guide.md
```diff
@@ -76,7 +76,7 @@ Our (un)official (and somewhat incomplete) rules for code style:
 * Globals
   * Currently, all app globals are set as a constant value which is injected by `WebPack`. We don't do this in general, preferring modular code.
   * If you really think there is a valid reason to set a global, it must be injected from `build_code/webpack_common.js` inside the `DefinePlugin` function. From there, a runtime constant must be created and exported from `src/core/inject_build_constants.js`
-  * In order to access the globals, you can import the constants from `src/core/inject_build_constants.js` like so: `import { const_name } from "src/core/injected_build_constants.js"`
+  * In order to access the globals, you can import the constants from `src/core/inject_build_constants.js` like so: `import { const_name } from "src/core/injected_build_constants"`
```
GarvinH, line 79: really really minor, but it's `.ts` now 😆

client/src/models/businessConstants.ts
```diff
@@ -0,0 +1,723 @@
...
+function getNormalizedEntry(
+  obj: RawBusinessConceptInputTextValue
+): BusinessConceptTextValue {
+  if (obj.hasOwnProperty("text")) {
+    return { text: (obj as BusinessConceptTextValue).text };
+  } else if (obj.hasOwnProperty(lang)) {
+    return { text: (obj as BilingualBusinessConceptTextValue)[lang] };
+  } else {
+    throw "shouldn't happen";
```
GarvinH, line 26: I thought we added a lint for not throwing text. Maybe it only works for JS?

Should probably use the Error class

docs/js-tips-style-guide.md
GarvinH: 😂

GarvinH (approved):
Other than the error thing, I think this should be good?

GarvinH:
oof, I found an instance where it's not working

http://dev.ea-ad.ca/business-constants-ts/index-eng.html#orgs/dept/326/infograph/financial

client/src/models/businessConstants.ts
```diff
@@ -0,0 +1,723 @@
...
+  entries: RawBusinessConceptEntrySet
+): BusinessConceptEntrySet {
+  return _.mapValues(entries, (entry) => getNormalizedEntry(entry));
+}
+
+export const sos: {
+  [key: number]: BusinessConceptTextValue;
+} = normalizeBilinguals({
+  1: {
+    text: "Personnel",
```
GarvinH, line 47: we'll have to add `so_num` to these

client/src/models/businessConstants.ts
```diff
@@ -0,0 +1,723 @@
...
+    en: "Result to be achieved in the future",
+    fr: "Résultat à atteindre dans le future",
+  },
+});
+
+export const months = normalizeBilinguals({
+  0: {
+    text: "",
+  },
+  1: {
```
GarvinH, line 491: these need `ix`

GarvinH (changes requested):

client/.eslintrc.json
```diff
@@ -13,6 +13,15 @@
   "commonjs": true,
   "es6": true
   },
+  "overrides": [
+    {
+        "files": ["*.ts", "*.tsx"],
+        "rules": {
+            "no-undef": "off",
+            "import/extensions": "off"
```
AlexCLeduc, line 21: These 2 rules are handled via typescript, sometimes the logic is different than eslint and they go out of sync, better to keep typescript in charge here

client/src/models/businessConstants.ts
```diff
@@ -0,0 +1,739 @@
...
+    21: {
+      en: "External Revenues",
+      fr: "Revenus externes",
+    },
+    22: {
+      en: "Internal Revenues",
+      fr: "Revenus internes",
+    },
+  },
+  (val: BilingualBusinessConceptTextValue, key) => ({
```
Stephen-ONeil, line 108: Just curious, seeing some places where you declare (some) types for arguments to lodash predicates and other places where you don't. Got a pattern for that or is it something you're playing by ear?

Stephen-ONeil (approved):
Nice! Had one question for my own curiosity, LGTM though :+1:

client/src/types/global.d.ts
```diff
@@ -1,20 +1,25 @@
...
+
+const IS_A11Y_MODE: boolean | undefined;
+const IS_DEV_LINK: boolean | undefined;
+const IS_CI: boolean | undefined;
+const IS_DEV: boolean | undefined;
+
+const APPLICATION_LANGUAGE: LangType | undefined;
+const SHA: string | undefined;
+const PREVIOUS_DEPLOY_SHA: string | undefined;
+const BUILD_DATE: string | undefined;
```
AlexCLeduc, line 23: turns out this cleaner syntax works the same

client/src/types/global.d.ts
```diff
@@ -1,20 +1,25 @@
...
-declare const BUILD_DATE: string | undefined;
-declare const CDN_URL: string | undefined;
-declare const LOCAL_IP: string | undefined;
 
+// Global types:
+type LangType = "en" | "fr";
+interface LangDict<T> {
+  en: T;
+  fr: T;
+}
```
AlexCLeduc, line 8: This'll be a useful generic type we use a bunch

client/src/models/businessConstants.ts
AlexCLeduc: Just fiddling with it until TS stops complaining. The first time I went through all the business constants, I wasn't aware that mapValues had generics, so I didn't use them. I'm not entirely sure in which cases we'll need them or not, but they were useful the 2nd time

PR #1038, opened 2021-04-23
Implicit chaining

### LINKED ISSUE #522: Performance implications of using explicit lodash chains over implicit chains
The explicit syntax is `_.chain(...)...value()`, implicit is `_(...)...` (note: doesn't `.value()` call). Turns out only the implicit version is [lazy/benefits from "shortcut fusion"](orgs/tag/HI1029/infograph/intro).

We should benchmark some pages, run a code mod to switch to all implicit calls, and see if there's any significant performance boost! In the future, if we take the time to learn the set of optimizable utilities, we could write (marginally?) more performant lodash chains.

GarvinH:
closes #522

### DIFF
--- a/client/build_code/write_footnote_bundles.js
+++ b/client/build_code/write_footnote_bundles.js
@@ -71,7 +71,7 @@ function populate_stores(parsed_models) {
 
     //ESTIMATES footnotes bundle
     const topics = obj.topic_keys.split(",").map((key) => key.replace(" ", ""));
-    if (!_.chain(topics).intersection(estimate_topics).isEmpty().value()) {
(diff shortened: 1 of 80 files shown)


Stephen-ONeil (changes requested):
Hard to comment on the code changes themselves, but as a meta comment this is something that'll need to be documented. Should try and cover the how and why of implicit chains in the context of the InfoBase, gotchas and tips. E.g. some sort of idea of when to break chains v.s. wrap a breaking method in `.thru` (if that's not actually an anti pattern? any good cases for that?) v.s. falling back to an explicit `_.chain`.

GarvinH:
closing because I closed #522

PR #1041, opened 2021-04-27
Change contact us page to tsx

GarvinH:


Stephen-ONeil (approved):
Haha, nice velocity on this 👍 

Before merging, please let CI finish and drop a screen shot of any changes in bundle stats. Want to see if there's any sort of inherent weight to even a trivial TypeScript port.

GarvinH:
image

Seems like an extra .2 mb

Stephen-ONeil:
Doesn't seem to be Typescript's fault. In fact, the contact related bundles might be a bit smaller. Instead, it seems there's some new(?) redundancy introduced with `html2canvas` showing up in two bundles now. A mystery to look in to, but not a big deal because that stuff's only reachable though clicking the button to generate a pdf for a table or panel.

GarvinH:
Gonna merge in then!

PR #1043, opened 2021-04-27
Drop lighthouse CI job

Stephen-ONeil:
I think it's been broken for a while, at least [this link](https://infobase-lhci.herokuapp.com/app/projects) from the readme doesn't work. I believe the heroku free tier is always running out of storage space on us/not being maintained. The job also flakes in CI.

Since no one checks (or can check?) the output, and since we ignore the noisy CI failures, it's not doing anything but taking up CI resources.

Stop me if I'm wrong!


Stephen-ONeil:
> So, at every stage, if a package is already installed in an ancestor node_modules folder, then it is not installed at the current location.
https://docs.npmjs.com/cli/v7/configuring-npm/folders#cycles-conflicts-and-folder-parsimony

Good to know, little gotcha to watch out for when dropping packages from the top level `package.json`.

ktw1016 (approved):
byebye lighthouse😢

PR #1045, opened 2021-04-28
Displaytable to typescript

GarvinH:
Putting this on pause (incomplete).

Lodash type checks don't help the webpack checker for determining the types being used and I'm not sure why.
> var : string | Function
> _.isString(var) ? var()

string var is not callable????

A couple of hacky workarounds used because of type definition overrides.

Gonna prioritize working on the lower level files and moving those to TS first.
Also gonna work on the docs for migrating to TS.

If there's no drastic changes that occur to displaytable, this PR may be a good point to continue from. Otherwise, probably just start from the latest master commit and use this as inspiration if needed.


client/src/components/DisplayTable/DisplayTable.tsx
```diff
@@ -520,50 +574,56 @@ export class DisplayTable extends React.Component {
...
-                            col_configs_with_defaults[col_key].formatter(
+                              ) ? (
+                                <Format
+                                  type={
+                                    col_configs_with_defaults[col_key].formatter
+                                  }
+                                  content={row[col_key]}
+                                />
+                              ) : (
+                                col_configs_with_defaults[col_key].formatter(
```
ktw1016, line 608: Like Garvin was saying, getting `  No constituent of type 'string | Function' is callable` here

client/src/components/DisplayTable/DisplayTable.tsx
```diff
@@ -331,12 +377,14 @@ export class DisplayTable extends React.Component {
           key="columnToggleUtil"
           columns={
             <LegendList
-              items={_.map(all_ordered_col_keys, (key) => ({
-                id: key,
-                label: col_configs_with_defaults[key].header,
-                active: _.includes(visible_ordered_col_keys, key),
-              }))}
-              onClick={(clicked_key) => {
+              items={_.map(all_ordered_col_keys, (key: string) => {
```
ktw1016, line 380: Another bug here, not sure why Typescript thinks `items` is `boolean[]`

client/src/charts/legends/LegendList.tsx
```diff
@@ -2,20 +2,34 @@ import classNames from "classnames";
...
 
 import { is_a11y_mode } from "src/core/injected_build_constants.js";
 
 import "./LegendList.scss";
 
+export interface LegendItemProps {
+  active: boolean;
+  id: string | number;
+  label: string;
+  color?: string;
```
ktw1016, line 15: Might make a separate ts file for all legends to use. Will take care of all legends and include that change later

client/src/components/DisplayTable/DisplayTable.tsx
```diff
@@ -297,13 +343,14 @@ export class DisplayTable extends React.Component {
...
+    ) as string[];
+
     const csv_string = _.chain(visible_ordered_col_keys)
-      .map((key) => col_configs_with_defaults[key].header)
+      .map((key: string) => col_configs_with_defaults[key].header)
       .thru((header_row) => [header_row])
       .concat(
-        _.map(sorted_filtered_data, (row) =>
-          _.map(visible_ordered_col_keys, (key) =>
+        _.map<any, string[]>(sorted_filtered_data, (row: DisplayTableData) =>
```
ktw1016, line 352: To be honest, I don't know how this generic (`<any, string[]`) fixed it if someone can enlighten me on that, please do

AlexCLeduc: this gets matched against the lodash type declarations (cmd-click on `_.map` in vscode):

```typescript
        map<T, TResult>(collection: T[] | null | undefined, iteratee: ArrayIterator<T, TResult>): TResult[];
        /**
         * @see _.map
         */
        map<T, TResult>(collection: List<T> | null | undefined, iteratee: ListIterator<T, TResult>): TResult[];
        /**
         * @see _.map
         */
        map<T>(collection: Dictionary<T> | NumericDictionary<T> | null | undefined): T[];
        /**
         * @see _.map
         */
        map<T extends object, TResult>(collection: T | null | undefined, iteratee: ObjectIterator<T, TResult>): TResult[];
        /**
         * @see _.map
         */
        map<T, K extends keyof T>(collection: Dictionary<T> | NumericDictionary<T> | null | undefined, iteratee: K): Array<T[K]>;
        /**
         * @see _.map
         */
        map<T>(collection: Dictionary<T> | NumericDictionary<T> | null | undefined, iteratee?: string): any[];
        /**
         * @see _.map
         */
        map<T>(collection: Dictionary<T> | NumericDictionary<T> | null | undefined, iteratee?: object): boolean[];

```

In this case, I think it's matching the first rule, 

```typescript
map<T, TResult>(collection: T[] | null | undefined, iteratee: ArrayIterator<T, TResult>): TResult[];
```

Which means that it's mapping a list of `any` into  a list of `strings`.  It looks like the mapper function is expecting `DisplayTableData` elements, though. So this should really be `_.map<DisplayTableData, string>`. If you try and make that switch, though, it complains that `sorted_filtered_data` isn't the right type, which probably means we need some more annotations in `sorted_filtered_data`'s derivation above.

AlexCLeduc (changes requested):
We should try and get rid of the `any`, because `any` removes all type safety

PR #1046, opened 2021-04-28
Add docs for typescript tips

GarvinH:


AlexCLeduc (approved):
Thanks! And nice progress on lodash, that'll be super helpful.

PR #1050, opened 2021-04-28
Node 14

Stephen-ONeil:
Move entire project to Node 14, both as the expected dev environment and as the intended target for Google Cloud Function deploys. Move node-targeted code to esModules [where possible](https://github.com/TBS-EACPD/infobase/pull/1050#issuecomment-830222733).

Have yet to see client builds hitting memory limits after switching to Node 14 🤞. **Edit:** still true locally, where it seems the problem had always been the baked in memory limits of the older V8 I was still on. Still very easy to burn through the CI container's 4GB, but I've re-re-fixed the prod build memory capping in this branch and I think I've got CI surviving the build jobs consistently.


Stephen-ONeil:
New CI images a much larger, fingers crossed the impact on CI job spin up isn't too great.

Stephen-ONeil:
... had hoped this would be the time we could fully standardize on esModules, but:
  - in `/server` and `/email_backend`:
    - even with Node 14... turns out GCF specifically _do not_ support esModules, so even if we can now run locally from `src` without babel we still need to transpile pre-deploy 
    - Jest only semi-supports esModules, and doesn't do a good enough job of it to work out of the box. Fine though, since it transpiles using your local .babelrc by default anyway (and if we still need to transpile for GCF, it's preferable to run tests against the transpiled version anyway)
  - in `/client`:
    - while I got builds running with all `/build_code` moved to esModules, the resulting site had some confusing React errors, didn't dig too deep in to that because of the following issue
    - Storybook doesn't support an esModule `.storybook/main.js`. Extra painful, because our storybook config imports from `build_code/webpack_common.js` to share webpack rules... So for now Storybook also forces our `webpack_common.js` to remain CommonJS one way or another (tried some workarounds, but they failed to work-around)

Stephen-ONeil:
Well, minor victory, since the local `server` and `email_backend` can be run without transpilation, debugging them will be a nicer experience!

ktw1016 (approved):
It's a shame that storybook doesn't support es moudle.. but I'm a fan that server code support fully es module

PR #1053, opened 2021-04-30
Changing core files, etc. to TS if very little dependencies

GarvinH:
in particular

- /core where there's no dependencies to js
- /icons

Edit: this was built on-top of #1037 as it had the business constants changes (should rebase with master once that branch gets merged in)


client/src/core/format.ts
```diff
@@ -94,22 +94,33 @@ const result_percent_formatter = {
...
+const compact = (
+  precision: number,
+  val: number,
+  lang: LangType,
+  options: OptionsProps
+) => {
   precision = precision || 0;
 
-  const abbrev = {
+  const abbrev: { [key: number]: { [key: string]: string } } = {
```
AlexCLeduc, line 111: `key` should be `LangType` instead of `string`. Even better: create a generic `LangDict<T>` type of the form `{ en: T, fr: T}`

ktw1016 (approved):
👍

PR #1054, opened 2021-05-03
Remove eslint import extension error for tsx

ktw1016:
Just noticed this new rule. From what I understand, you cannot explicitly put file extensions in import statements, so we can't have this rule. Correct me if I'm wrong!

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -66,8 +66,7 @@
         ]          
       }
     ],
-    "import/no-commonjs": 2,
-    "import/extensions": ["error", "always"]
+    "import/no-commonjs": 2
   },
   "settings": {
     "import/resolver":{


Stephen-ONeil:
Hm? That a TSX thing? Were you getting errors from the rule you couldn't fix?

It's a necessary rule for the two node projects, as node 14 with esModules requires import file paths to be fully qualified. I made it global and ran it against `/client` too because I briefly tried moving the client build scripts to esModules too (thwarted by storybook as mentioned). I didn't have any problems on my end during that.

Moving to esModules aside, I prefer the module extensions to specified anyway, especially now that we have both JS and TS.

Stephen-ONeil (commented):
See comment above, I don't know what you mean by "you cannot explicitly put file extensions in import statements". 

Are you talking about imports for node modules?
  a) this rules shouldn't complain about importing from them _unless_
  b) you have imports like `package/sub-package`, in which case you might just need to change it to `package/sub-package/index.js` or something along those lines (depends on the package)

ktw1016:
Lint will give you an error that it needs to specify file extension, but you cannot for ts or tsx. E.g.: on about.tsx, check the import on LabeledTable

Stephen-ONeil (changes requested):
Ugh, no good way to do it. We _want_ the ts extensions when importing it from a js module, but we _can't_ have it when importing from another TypeScript module. The `import/extensions` rule can be set per-extension but has no "sometimes" option (it's all or nothing, either always have the extension or never have the extension). Will have to turn it off in `/client`.

Leave it in root though, just disable it in the client rules please.

Stephen-ONeil:
For the record, our build process can handle an import for a `.ts`/`.tsx` whether or not it's from inside of a TypeScript module. Seems a good amount of developers want an option to disable the error/ignore the dumb "no TypeScript extension" convention, but TypeScript devs are sticking to their guns 🤷

AlexCLeduc:
I'm addressing this in #1037, what I'm doing is adding an eslint override in TS/TSX files so that they can import files without extensions. This unfortunately allows them to import non-TS/TSX sources without extensions.

```json
  "overrides": [
    {
        "files": ["*.ts", "*.tsx"],
        "rules":
            "import/extensions": "off"
        }
    }
  ]


```

ktw1016:
Closing in favor of #1037

PR #1056, opened 2021-05-03
dropdownmenu component js to tsx

ktw1016:


client/src/components/DropdownMenu.tsx
```diff
@@ -1,22 +1,44 @@
...
+  is_open: boolean;
+};
+
+export class DropdownMenu extends React.Component<
+  DropdownMenuProps,
+  DropdownMenuState
+> {
+  dropdown_ref = React.createRef<HTMLDivElement>();
+  toggle_dropdown_button = React.createRef<HTMLDivElement>();
+  dropdown_area = React.createRef<HTMLDivElement>();
```
ktw1016, line 29: A bit of annoyance. You can't straight `this.dropdown_ref = ...` into constructor

client/src/components/DropdownMenu.tsx
```diff
@@ -1,22 +1,44 @@
...
+
+  constructor(props: DropdownMenuProps) {
     super(props);
     this.state = {
       is_open: false,
     };
```
ktw1016, line 35: Another bit of annoyance: can't use that experimental syntax to initalize state outside of constructor so we have worst of both worlds😔

ktw1016:
I think there could be more of these annoying little bits going forward. Also community seems to have gone more of using hooks with functional components + typescript. I think that could be more ideal for cleaner code as I'm looking more into this

client/src/components/DropdownMenu.tsx
```diff
@@ -1,22 +1,44 @@
...
-
-    this.dropdown_ref = React.createRef();
   }
-  handleWindowClick = (e) => {
+
+  handleWindowClick = (e: MouseEvent) => {
     const { is_open } = this.state;
-    if (is_open && !this.dropdown_ref.current.contains(e.target)) {
+    const dropdown_node = this.dropdown_ref.current;
+    if (is_open && dropdown_node && !dropdown_node.contains(e.target as Node)) {
```
ktw1016, line 41: https://github.com/Microsoft/TypeScript/issues/15394#issuecomment-297495746
```
Basically EventTarget is the most general type, of which Element is a subtype, and HTMLElement is a subtype of that. If you get back a thing from the DOM, we generally have no idea which it is, and you should add a type assertion to "add in" the specific external knowledge that you have about the structure of your particular DOM layout.

It's possible, for example, that the .target of an event is not an Element (you can add event listeners to XMLHttpRequest, for example, and XMLHttpRequest does not have a getBoundingClientRect method).
```
Basically: target may or may not be a Node, so TS can't infer. Hence, need to assert `as Node`

client/src/components/DropdownMenu.tsx
AlexCLeduc: When I saw this, I was worried that different instances would share the same ref, but it turns out that that `React.createRef` is called during instance construction


```js
createObj = () => ({a:5}}
class A {
    prop = createObj()
}
a1 = new A()
a2 = new A()
a2.prop === a3.prop // False! I guess `prop = createObj()` gets run for each instance

```

.eslintrc.json
```diff
@@ -66,8 +66,7 @@
         ]          
       }
     ],
-    "import/no-commonjs": 2,
```
AlexCLeduc, line 69: I don't think we need to touch the lint config here?

client/src/components/DropdownMenu.tsx
```diff
@@ -58,7 +87,9 @@ export class DropdownMenu extends React.Component {
 
     const aria_label = dropdown_a11y_txt
       ? dropdown_a11y_txt
-      : _.isString(dropdown_trigger_txt) && dropdown_trigger_txt;
+      : _.isString(dropdown_trigger_txt)
+      ? dropdown_trigger_txt
+      : undefined;
```
AlexCLeduc, line 92: super nitpick: can we just use a `let` here? nested ternaries hurt my brain

AlexCLeduc (approved):
If the eslint isn't an accident (although I guess you'll have to rebase anyway), LGTM

.eslintrc.json
ktw1016: oh yea, this was temporary change for development purposes lol

client/src/components/DropdownMenu.tsx
Stephen-ONeil: How about an `if` statement inside an immediately invoked function (or a  `get_aria_label` declared elsewhere or something)?

ktw1016: yup, I already did iif in my commit! https://github.com/TBS-EACPD/infobase/pull/1056/commits/db235bd5e3966c504c0db8e65cd01f6435599154

PR #1058, opened 2021-05-03
Revive treemap

Stephen-ONeil:
At Val's request, revive the treemap subapp. It gets to live for now, but if it's still here in a year and hasn't been polished/still receives minimal traffic then it'll go for good.


ktw1016 (approved):
it's back😆

Stephen-ONeil:
:pensive:

PR #1060, opened 2021-05-06
Refine pattern for defining GraphQL queries/hooks in client

Stephen-ONeil:
Refine pattern from `client/graphql_utils/graphql_utils.js`'s `query_logging_wrapper`. Introduce (tentatively named) `query_maker`, to return both promise and hook based query options, with standard logging built in, based on a common query configuration. Enforces naming of queries (for logging) and strongly suggest pairing each query with a response resolver (so that the query itself encapsulates as much knowledge of the actual query structure/server schema as it can).

Note, the hooks, being based on Apollo's `useQuery`, aren't actually usable until #1004 adds the necessary `ApolloProvider`. This was originally a branch off of `SI_apollo_hooks`, but I split it off when I decided it had reached a reviewable state without getting in to potentially touching a bunch of service inventory code.

### DIFF
--- a/client/src/graphql_utils/graphql_utils.js
+++ b/client/src/graphql_utils/graphql_utils.js
@@ -106,9 +106,7 @@ export function get_client() {
   return client;
 }
 
-const make_query_promise = (query_name, query, response_resolver) => (
-  variables
-) => {
+const make_query_promise = (query_name, query, resolver) => (variables) => {
   const time_at_request = Date.now();
 
   return get_client()
@@ -119,7 +117,7 @@ const make_query_promise = (query_name, query, response_resolver) => (
         _query_name: query_name,
       },
     })
-    .then(response_resolver)
+    .then(resolver)
     .then((resolved_response) => {
       log_standard_event({
         SUBAPP: window.location.hash.replace("#", ""),
@@ -142,9 +140,7 @@ const make_query_promise = (query_name, query, response_resolver) => (
     });
 };
 
-const make_query_hook = (query_name, query, response_resolver) => (
-  variables
-) => {
+const make_query_hook = (query_name, query, resolver) => (variables) => {
   const [time_at_request, set_time_at_request] = useState(Date.now()); // eslint-disable-line no-unused-vars
 
   const { loading, error, data } = useQuery(query, {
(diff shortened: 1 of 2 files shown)


client/src/models/covid/queries.js
```diff
@@ -3,286 +3,230 @@ import _ from "lodash";
...
+  resolver: (response) =>
+    _.get(response, "data.root.gov.years_with_covid_data"),
+});
+export const {
+  query_org_years_with_covid_data,
+  useOrgYearsWithCovidData,
+} = query_maker({
+  query_name: "org_years_with_covid_data",
+  query: gql`
+    query($lang: String! = "${lang}", $org_id: String!) {
```
Stephen-ONeil, line 38: For inline query variable defaults, like `$top_x` below, I think the query definition is definitely the place to put them. Actually a bit conflicted about adding the default `lang` value inline though. Most likely going to lift passing a default lang value back up in to the `query_maker`.

client/src/graphql_utils/graphql_utils.js
```diff
@@ -97,70 +106,98 @@ export function get_client() {
...
       });
+
       throw error;
     });
 };
 
-const InnerLoadingHoc = ({ Component, data_to_props }) => (props) => {
-  if (props.data.loading) {
-    return <div> Loading ... </div>;
+const make_query_hook = (query_name, query, resolver) => (variables) => {
```
AlexCLeduc, line 143: Hooks should always follow the `use*` naming pattern, otherwise it's not clear that this is indeed a hook

client/src/graphql_utils/graphql_utils.js
```diff
@@ -97,70 +106,98 @@ export function get_client() {
...
-  variables,
-}) =>
-  apollo_connect(query, {
-    options: variables ? { variables } : {},
-  })(InnerLoadingHoc({ Component, data_to_props }));
-
-assign_to_dev_helper_namespace({
-  query_api: (query) => client.query({ query }),
-});
+export const query_maker = ({ query_name, query, resolver = _.identity }) => {
```
AlexCLeduc, line 184: this is also a hook, and should use the `use*` naming convention

Stephen-ONeil: Same as above. `query_maker` returns `{query_<query_name>, use<query_name, in pascal case>}`. The first being a promise based query, the second being the hook equivalent.

PR #1069, opened 2021-05-07
Drop currency code from French dollar formats

Stephen-ONeil:
Before
image

Now
image


Stephen-ONeil:
Note, I added `currencyDisplay: "narrowSymbol",` to both en and fr configs, but it's equivalent to the default in en. Was just keeping them consistent (probably could have gone further and cleaned this up, but cleaning up `format.ts` is a rabbit hole).

Stephen-ONeil:
Hey, come on
image

PR #1071, opened 2021-05-10
Typescript migrations

ktw1016:
Files completed:
- `IconArray.js`
- `index_data.js`
- `graphRegistry.js` (kinda hacky)
- `AuthHeightVirtualList`
- `Details.js`
- `TabbedContent.js`
- `HeightClipper.js`
- `KeyConceptList.js`
- `Accordion.js`
- `BackToTop`


client/src/components/BackToTop/BackToTop.tsx
```diff
@@ -9,21 +9,34 @@ import { is_mobile } from "src/core/feature_detection";
 
 import "./BackToTop.scss";
 
-export class BackToTop extends React.Component {
-  constructor(props) {
+interface BackToTopProps {
+  focus: Function;
```
Stephen-ONeil, line 13: Could be `() => void` I think. That should make it clear that the `focus` function is just for side effects.

client/src/components/AutoHeightVirtualList.tsx
```diff
@@ -0,0 +1,77 @@
+import React from "react";
+import { List, CellMeasurerCache } from "react-virtualized";
+
+interface AutoHeightVirtualListProps {
+  list_ref: React.RefObject<any>;
```
Stephen-ONeil, line 5: Ugh, I tried with `<List>` instead of any, but `@types/react-virtualized`'s typing for `Grid` is missing the `getTotalRowsHeight` method.
image

Wonder if in a case like this it's worth hitting `@types/react-virtualized` with the ol'patch-package?

client/src/components/AutoHeightVirtualList.tsx
```diff
@@ -0,0 +1,77 @@
+import React from "react";
+import { List, CellMeasurerCache } from "react-virtualized";
+
+interface AutoHeightVirtualListProps {
```
Stephen-ONeil, line 4: Shouldn't there be a way to import and extend `ListProps` from `@types/react-virtualized`? Would clean this up and get an implementation of `rowRenderer` that doesn't use an `any`.

client/package.json
```diff
@@ -45,7 +45,11 @@
...
     "@nivo/treemap": "^0.67.0",
     "@types/d3-dsv": "^2.0.1",
+    "@types/d3-dispatch": "^2.0.0",
+    "@types/d3-ease": "^2.0.0",
     "@types/d3-scale": "^3.2.2",
+    "@types/d3-selection": "^2.0.0",
+    "@types/d3-transition": "^2.0.0",
     "@types/dompurify": "^2.2.1",
     "@types/js-yaml": "^3.12.5",
     "@types/lodash": "^4.14.162",
```
Stephen-ONeil, line 55: Thinking about it, types should be dev dependencies.

Stephen-ONeil (changes requested):

client/src/components/AutoHeightVirtualList.tsx
ktw1016: Yeah, there is a way and that was my initial approach as well ([you can see it in this commit](https://github.com/TBS-EACPD/infobase/pull/1071/commits/524db7b402d5cf9bdea4121936d261e4b5b52a93))
But `ListProps` require `height`, which requires `Typeahead` to pass in `height` when passing props into `AutoHeightVirtualList`. But obviously the sole purpose of `AutoHeightVirtualList` is to control `height` so I decided to manually define type for this component

client/src/components/AutoHeightVirtualList.tsx
ktw1016: Yup I ran into the same issue. I personally would rather not do any patching just for this. One workaround could be `(this.list_ref.current.Grid as any).getTotalRowsHeight();`
casting `as any` actually seem like common thing to do in typescript.. At least from what I've seen online

client/src/components/AutoHeightVirtualList.tsx
Stephen-ONeil: Googled up some options, I think `interface AutoHeightVirtualListProps extends Omit<ListPRops, "height">{` should do it?

https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys

Stephen-ONeil: Some alternate options seem to be overriding by extending with `height?: never`. Probably a use case for that patern, but the `Omit` util type seems better here.

client/src/components/AutoHeightVirtualList.tsx
Stephen-ONeil: That workaround looks good to me, definitely beats patching!

Stephen-ONeil: Hmm, well, it might actually be a bad idea to use a flawed type. Might confuse future devs who might try and use the other `react-virtualized` methods listed in the docs that are missing from the type. I guess just leave a comment on the `any` type ref mentioning that missing fields in `@types/react-virtualized`.

Stephen-ONeil (changes requested):
See responses to comments above. Two small things to do there, possibly. Otherwise all looks good to me!

client/src/components/AutoHeightVirtualList.tsx
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  render() {
+    return (
+      <List
+        {...this.props}
+        rowHeight={this.props.rowHeight}
+        rowCount={this.props.rowCount}
+        width={this.props.width}
+        rowRenderer={this.props.rowRenderer}
```
ktw1016, line 59: Pretty strange, it seems like Typescript doesn't know `props` extends from `ListProps`. It's like they don't exist. It complains that those 4 props (`rowHeight, rowCount, width, rowRenderer`) are not being passed in. Fixed by just explicitly passing them in

Stephen-ONeil (commented):
Pretty strange, I tried `type test = Omit<ListProps, "height">;` and it reports 
```
type test = {
  [x: string]: any;
  [x: number]: any;
}
```

Seems to be because `ListProps` is a union type with `GridCoreProps`, which itself has `[key: string]: any;` (a comment says it's there so you can use an arbitrary prop to force a pure component to rerendner... ouch, hope that's not a React-TS pattern). Did some tests with `Omit` on some simple types, and if there's a `[key: string]: any;` then no matter what you omit you get `{ [x: string]: any; [x: number]: any;}` back. Ouch.

I think I've dug up some additional utility type snipets we can use to fix this, will put up a PR with them in a minute.

Stephen-ONeil:
Ok, did a local test and everything works as expected using the `IndexTypeSafeOmit` util type added in#1121!

ktw1016:
Works!

client/src/components/Accordion/Accordions.tsx
```diff
@@ -112,10 +132,10 @@ const StatelessPullDownAccordion = ({
       </button>
     </div>
     <TransitionGroup component={FirstChild}>
-      {isExpanded && (
+      {isExpanded ? (
         <AccordionEnterExit
           className="pull-down-accordion-body"
-          style={{ paddingTop: "5px" }}
+          style={{ paddingTosp: "5px" }}
```
Stephen-ONeil, line 138: Typo

client/src/components/AutoHeightVirtualList.tsx
```diff
@@ -0,0 +1,66 @@
...
+      }
+    }
+  }
+
+  render() {
+    return (
+      <List
+        {...this.props}
+        {...{
+          ...this.props,
```
Stephen-ONeil, line 59: Redundant spread of `...this.props`?

client/src/components/TabbedContent/TabbedContent.tsx
```diff
@@ -6,7 +6,36 @@ import { lang } from "src/core/injected_build_constants";
...
+type tab_option = {
+  key: string;
+  label: string;
+  is_open: boolean;
+  is_disabled: boolean;
+};
+
+interface TabbedControlsProps {
+  tab_options: tab_option[];
+  tab_callback: Function;
```
Stephen-ONeil, line 18: I think this can be `(key: string) => void;`

Stephen-ONeil (changes requested):
Few small things

Stephen-ONeil (approved):
:+1:

PR #1077, opened 2021-05-11
SI using query factory

ktw1016:


ktw1016:
I felt the need for separating query promise and hook because each are used differently
- I don't see a use case where we need both promise and hook
- Promises have another layer with `populate` script, which in itself has a conditional to switch between subject level and also  has global like pattern with `subject.set_has_data` or `model.create_and_register` where as hooks are being directly called inside panel level

ktw1016:
Waiting for review before I apply the changes for all other service queries

Stephen-ONeil (changes requested):
> I felt the need for separating query promise and hook because each are used differently

Bzzzt, wrong. They are functionally equivalent, the hook is just a special case syntax for use in function components. Otherwise, they both make an async request for some data, apply the same resolver, make the same logs, and then return that data. Only difference is the general use promise syntax versus the special case hook syntax

> * I don't see a use case where we need both promise and hook

Correct-ish. A function component can't use the promise (without hookify-ing it anyway), and a class component can't use the hook. We'll never have reason to (or be able to) use both in the same place... but we're probably always going to have a mix of class and function components in the code base, let alone non-React code. Again, the point of this is that the single factory returns a functionally equivalent hook and promise from a common config. 

> * Promises have another layer with `populate` script, which in itself has a conditional to switch between subject level and also  has global like pattern with `subject.set_has_data` or `model.create_and_register` where as hooks are directly called inside panel level

The (hopefully soon to be legacy) "populate" scripts and `ensure_loaded`, etc, use promises, yes... ? Absolutely nothing stopping a class component from directly using a promise to load data too (look closer, it's what all the covid panels do). Opposite actually, the vast majority of our current panels _can't_ use the hooks, unless we decide to make a refactor away from class components part of the move to direct GraphQL loading (which would be kind of arbitrary to do).

> See [2f9c1a6](https://github.com/TBS-EACPD/infobase/commit/2f9c1a6e4a5a63b327c6724b12bf331691cea927) for isolated change of this. (Another train of thought in this change was to make props that go into each custom hooks more consistent.. For example `useServicesList` and `useSingleService` is taking different args)

Missing the point, the factory was the standard pattern that enforced consistency, you need to rewrite your existing hooks/query construction patterns to fall in line with it, not the other way around.

ktw1016:
> Bzzzt, wrong. They are functionally equivalent, the hook is just a special case syntax for use in function components. Otherwise, they both make an async request for some data, apply the same resolver, make the same logs, and then return that data. Only difference is the general use promise syntax versus the special case hook syntax

I never said they are functionally equivalent, and I'm well well aware that they are. What I meant here is exactly what you pointed out - syntax difference.

> Correct-ish. A function component can't use the promise (without hookify-ing it anyway), and a class component can't use the hook. We'll never have reason to (or be able to) use both in the same place... but we're probably always going to have a mix of class and function components in the code base, let alone non-React code. Again, the point of this is that the single factory returns a functionally equivalent hook and promise from a common config.

Yeah we're gonna have mix of class and functional components in the codebase, but that doesn't mean that a single factory returning both promise and a hook is the way to go. Each sections of code where promises or hooks are needed, can independently decide if it needs hooks or promises. There is no need to get less efficient, or under featured version of one for the sake of exactly duplicating hook in the same way as promise. The point of my comment and this PR is to change that single factory because it doesn't make sense when using a hook. If you'd like, I can revert back to one factory returning both promise and hook, I would just be adding what I did in this PR to both promise and hook.

> The (hopefully soon to be legacy) "populate" scripts and `ensure_loaded`, etc, use promises, yes... ? Absolutely nothing stopping a class component from directly using a promise to load data too (look closer, it's what all the covid panels do). Opposite actually, the vast majority of our current panels _can't_ use the hooks, unless we decide to make a refactor away from class components part of the move to direct GraphQL loading (which would be kind of arbitrary to do).

My bad for not explicitly clarifying. Despite nothing stopping from a class component using a promise to load data (like what all covid panels do), the boilerplate and unnecessary complexity that adds to each panel is far inferior to that of simply using hooks. (Having to deal with async yourself, more state management, unnecessary configs). Some more problems with how query factory was designed: Unnecessary configs and conditionals due to the fact that you cannot control the query at all at each panel level,  duplicated code for all of the queries inside the query code, as well as useless exports of unused promises/hooks.

Furthermore, I believe those vast majority of panels _should_ be refactored into functional components so they can use hooks, should they need to load data. But that's far in the future and is for another discussion.

> you need to rewrite your existing hooks/query construction patterns to fall in line with it, not the other way around.

In fact, that's what I was originally doing. Much easier for me to simply change one, copy and paste for all other queries but it was like over bloating my code for the sake to make my code work _just_ so it works with that single factory. I had to ask myself, for what gains? At the end of the day, I would have repetitive 25 query factory calls exporting 50 promises/hooks with half of them unused.

IMO, the value of of these wrappers are dealing with logging (These wrappers make logging consistent across and removing logging boilerplate code for each query construction.) - not returning _functionally equivalent promise and hook at the same time_. Please let me know area of improvement in my code or other reasons to revert this PR back.

Stephen-ONeil:
Ok, let's pop the stack back to the precipitating discussion [here](https://github.com/TBS-EACPD/infobase/pull/1004#issuecomment-833737712) and get on a call tomorrow. Seems we were not actually on the same page about this.

Stephen-ONeil:
> I would have repetitive 25 query factory calls exporting 50 promises/hooks with half of them unused.
>
> IMO, the value of of these wrappers are dealing with logging (These wrappers make logging consistent across and removing logging boilerplate code for each query construction.) - _not returning functionally equivalent promise and hook at the same time_.

The logging is a bonus, the mountain of (potentially unused) promise/hook pairs which are functionally equivalent and built from the same config _is_ the point. Like I said in #1004,  "...centralizing all query+resolver pairs for a given topic to one module in /models/<topic>..." was the point. This isn't about boilerplate code, this is about having an isolated and change resilient client-server interface.

Stephen-ONeil:
@ktw1016 mind shelving this for now? I'll try and develop the client side query pattern a bit more soon, we can return to the question then.

PR #1079, opened 2021-05-13
Implemented SortDirection component

csong202:


client/src/components/SortDirection/SortDirection.js
```diff
@@ -2,6 +2,7 @@ import React from "react";
 
 import { trivial_text_maker } from "src/models/text.js";
 
+// eslint-disable-next-line import/no-unresolved
```
ktw1016, line 5: It seems like this can be cleaned up

client/src/components/SortDirection/SortDirection.stories.js
```diff
@@ -0,0 +1,19 @@
...
+import React from "react";
+
+import { SortDirection } from "./SortDirection";
+
+export default {
+  title: "SortDirection",
+  component: SortDirection,
+};
+
+const Template = (args) => <SortDirection {...args} />;
```
ktw1016, line 11: Should give it a background color. Can't see the white arrows on white background!

client/src/components/SortDirection/SortDirection.stories.js
```diff
@@ -0,0 +1,19 @@
...
+
+export const Basic = Template.bind({});
+Basic.args = {
+  className: "",
+  style: {},
+  active: true,
+  sortDirection: "",
```
ktw1016, line 18: Actually, `SortDirection` doesn't have `className`, `style` props!

If you look at `SortDirection.js`, you can see the props it takes:
```
export const SortDirection = ({ sortDirection, active }) => ...
```
The above is "destructuring" the props. (you can read about it [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment)) So this is equivalent as writing:
```
export const SortDirection = (props) => {
  const { sortDirections, active } = props;
  ...
```
And as you can see, `SortDirection` isn't using any `className` or `style` from props

ktw1016 (commented):
I think it's fair to say that there will always be 2 SortDirection, one for ascending and one for descending.

You should try to make a pair of SortDirection, one as ascending, one as descending. Just simply toggle one if the other is clicked, using state!

ktw1016:
Oops, I just realized it wasn't waiting for review.. My bad haha

client/src/components/SortDirection/SortDirection.stories.js
```diff
@@ -0,0 +1,37 @@
...
+        { name: "navy blue", value: primaryColor },
+        { name: "blue", value: secondaryColor },
+      ],
+    },
+  },
+};
+
+const Template = (args) => {
+  const [is_active1, set_active1] = useState(false);
+  const [is_active2, set_active2] = useState(false);
```
ktw1016, line 23: Kinda minor but you can just use one state for this
```
const [is_ascending, set_is_ascending] = useState(false);
  return (
    <div
      onClick={() => set_is_ascending(!is_ascending)}
    >
      <SortDirection sortDirection="ASC" active={is_ascending} />
      <SortDirection sortDirection="DESC" active={!is_ascending} />
    </div>
  ); 
```

ktw1016 (approved):
Other than the minor comment, looks good for now!

client/src/components/SortDirection/SortDirection.stories.js
csong202: I had the two states so both buttons would inactive initially

ktw1016: okay, why should both buttons be inactive initially?

csong202: both buttons are inactive initially on the tables on InfoBase so I thought I should do the same

PR #1084, opened 2021-05-17
Npm 7 dependency cleanup

Stephen-ONeil:
Pave the way for us all to get together on the same page with npm version 7 (will need to happen before this merges). Bunch of package upgrading and a couple small config changes in the process.

TODO:
  - [x] roll back prettier update/formatting, that stuff should go in a different PR
  - [x] fix `IB_base`, something's broken with the latest `css-minimizer-webpack-plugin` (ironically, might be a missing peer dependency, oof)
  - [ ] local package in `/scripts/common_node_scripts` isn't installing correctly
  - [ ] same explosion of node processes/memory issues in CI as #1045, must be a problematic package change. Out of scope for this branch, but a blocker for sure
  - [ ] confirm that the new core-js polyfilling config works
    - note, currently using `"usage"` option which is nice but likely makes build memory usage worse. If CI is back to running out of memory than may need to try `"entry"` instead (temporarily)
  - [ ] some dependency update in `/server` had a bunch of `coreMongooseArray` objects appearing in the snapshot output. Investigate further
  - [ ] update docs to mention installing npm 7 specifically
  - [ ] walk everyone through updating to npm 7 (easy, just the one gotcha of needing to run install with the `--legacy-peer-deps` flag, maybe open an issue to revisit that in a few months when maintainers have had time to catch up)


Stephen-ONeil:
Paused to give various packages time to make npm-7 friendly releases. Probably want at least a partial fix to #1092 before this continues. May end up needing to restart this branch/make careful cherry-picks due to other package changes that have happened in the meantime.

PR #1085, opened 2021-05-18
Drop FCAC late DP resources banner

Stephen-ONeil:
The whole FCAC special case for results reporting is covered by a footnote now, there's requests to exempt them from the banner. Done, and added a check in `dynamic_footnotes.js` to catch any future listing of an exempt org in the late org properties for a results doc config.


ktw1016 (approved):
👍

PR #1086, opened 2021-05-18
Better CI build memory usage logging

Stephen-ONeil:
Output looks like this, much more information than before:
image

Note: `VSZ` (virtual memory size, includes swap) and `RSS` (resident set size, the non-swapped physical memory) are in megabytes (output in kb, coverted via some awk for easier reading). `%MEM` is the "ratio of the process's resident set size to the physical memory on the machine", but it's a bit confusing to interpret as a % because processes could be sharing memory etc.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -140,7 +140,7 @@ jobs:
             while true; do
               sleep 5
               echo "====="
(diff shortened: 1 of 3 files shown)


ktw1016 (approved):
nice

PR #1088, opened 2021-05-18
Drop redact_env_vars_from_logging.sh, trust CircleCI built-in "secret masking"

### LINKED ISSUE #1087: Drop redact_env_vars_from_logging.sh
Hahaha, never noticed this, but a few months after I hacked together the `redact_env_vars_from_logging.sh` pattern for us (July 2019), CircleCI added essentially the same behaviour as a built-in feature (October 2019).

https://circleci.com/blog/keep-environment-variables-private-with-secret-masking/

Might as well drop that script (after double checking that the built-in equivalent is what we want), will clean up some general script boilerplate.

Stephen-ONeil:
Closes #1087


Stephen-ONeil:
Hmm, I forgot, my script subs in the env var's name. CircleCI just redacts it with \*'s. Former is a little more friendly for debugging, but very slightly less secure (e.g. collision of some secret key with part of some otherwise known hash, probably redacted in either case but worse for it to be redacted with a meaningful name at that point).

On the other hand, my script holds up all output until the process exits. The built-in secret masking doesn't do this so it's got the edge if any devs want to watch some CI scripts run in real time.

Could go either way on this, will have to think about it more.

Stephen-ONeil:
Eh, at the end of the day, I prefer the small bit of cleanup in those shell scripts to any marginal/debatable benefits of keeping the script. The shell scripts are never going to be friendly to new devs, the less in them, and the less fancy tricks used by them, the better.

ktw1016 (approved):
Never really knew what redact things were until now🤷‍♂️ Good to hear CircleCI masks built in now

PR #1089, opened 2021-05-19
CircleCI safe terser config

Stephen-ONeil:
https://www.npmjs.com/package/terser-webpack-plugin#parallel

Latest versions of terser will spawn far too many worker processes in CircleCI (or potentially any environment where `os.cpus()` is inaccurate). Will play around with hard coded values when in CI till I find a good balance of speed vs memory usage.


Stephen-ONeil:
When I ran a clean npm install to recreate what lead to the bad terser plugin version on other branches, I ended up with a pretty broken storybook build as well. Looks like current storybook releases are giving some people issues with webpack 5. Couple of issues, full of work arounds, on the storybook repo (e.g. add a certain version of some other package as a direct dependency of your project to bump it up in storybooks dependency chain, etc). None of those workarounds fully fixed it for me, so I reverted the  full, clean, package install. 

Either way, this PR should fix both the memory issue caused by upgrading terser and the additional lint warnings from upgrading eslint, even if it doesn't necessarily roll forward to those versions. Also fixesthe jam ups with the lockfile V2's I'd committed earlier, as I've rolled them back to V1's on this branch.

ktw1016 (approved):
👍

PR #1090, opened 2021-05-20
add script to track TS migration progress

AlexCLeduc:
1. `npm install --global dependency-cruiser`
2. `depcruise -T json -f dependencies.json --exclude node_modules/  src/`
3. `node generate_typescript_progress_stats.js`
4. open up `typescript_progress_stats.csv`, give it a quick check and paste it in a NEW sheet on the google drive document. Name the sheet based on today's date.
  - new sheet is recommended because pasting it over the existing sheet might remove important existing metadata of who's working on what. We could periodically delete old sheets.

### DIFF
--- a/client/.gitignore
+++ b/client/.gitignore
@@ -37,4 +37,4 @@ gruntdocs
 npm-debug*
 /storybook-static
 yarn.lock
- 
+typescript_progress_stats.csv
--- a/client/generate_typescript_progress_stats.js
+++ b/client/generate_typescript_progress_stats.js
@@ -1,0 +1,65 @@
+/*
+
+This script is to be run in the client/ dir and requires a dependencies.json file to be created by dep-cruiser:
+
+1. npm install --global dependency-cruiser
+2. depcruise -T json -f dependencies.json --exclude node_modules/  src/
+3. node generate_typescript_progress_stats.js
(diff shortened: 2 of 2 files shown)


Stephen-ONeil (changes requested):
Should `dependencies.json` also go in .gitignore?

If you don't mind, would be a good time to reorganize the scripts slightly. I did in `/email_backend` a bit ago but hadn't bothered/needed to in `/client` or `/server` yet. Would be nice if `/client/deploy_scripts` was moved to `/client/scripts/deploy` and this script was moved to `/client/scripts/util`. Not necessary though, it might be fiddly to make sure all the script paths everywhere are updated properly, I can do it later if you're busy.

Any reason to install `dependency-cruiser` globally, rather than as a local dev dependency etc?

AlexCLeduc:
Generally I hope this script doesn't live too long, so I didn't want to find the perfect place for it. I probably can't shift the scripts around with the time I have left, I'll leave it as is for now.

I forgot about the json file, I'll have the script delete it instead of `.gitignore`-ing it. 

> Any reason to install dependency-cruiser globally, rather than as a local dev dependency etc?

I think it's a pretty heavy dependency, it might require some graphical dependencies, I wouldn't want CI to install it

Stephen-ONeil (approved):
:+1:

ktw1016 (approved):
Just tried it out, it works👍 For others who'll be using this, let's make sure to not touch the "status" column of the sheet

PR #1091, opened 2021-05-21
fixed bug: added new config to leaf_spinner_configs

csong202:
- Scale wasn't working because when there is no config_name the LeafSpinner used the first config (initial) in leaf_spinner_configs.
- Fixed bug by adding a new config before initial that uses default_spinner_config_form(spinner_scale)

### DIFF
--- a/client/src/components/LeafSpinner.js
+++ b/client/src/components/LeafSpinner.js
@@ -6,7 +6,7 @@ import { primaryColor, secondaryColor } from "src/core/color_defs";
 import leaf_loading_spinner from "src/svg/leaf-loading-spinner.svg";
 import "./LeafSpinner.scss";
 
-export const LeafSpinner = ({ config_name }) => {
+export const LeafSpinner = ({ config_name, spinner_scale }) => {
   const default_spinner_config_form = (scale) => ({
     outer_positioning: "default",
     spinner_container_style: {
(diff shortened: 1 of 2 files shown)


Stephen-ONeil (changes requested):
Can you give an example of where this occurs as a bug?

I think this may actually be more of a poorly designed API on `SpinnerWrapper`'s part than a bug with `LeafSpinner`. `SpinnerWrapper` has the (\*unused?) option to render a spinner using the non-react `spin.js` version of our spinner, which is the only case in which it uses the `scale` prop. When (default case) it's just rendering a `LeafSpinner`, the `scale` prop shouldn't have been used.

\* I don't see anywhere that the `use_leaf_spinner: true` default is overridden. Seems, in practice, this wrapper doesn't do anything. Going to guess the need for rendering a non-React spinner was some legacy case we've since moved past. Could find and replace `SpinnerWrapper` with `LeafSpinner` and delete `SpinnerWrapper` without change anythings (I think we should actually, since `SpinnerWrapper` is a weird concept with a bad API anyway :laughing:).

In the case where we just drop `SpinnerWrapper`, we could also drop the `scale` prop anywhere it was being used. I think we want things picking from a predefined set of configs rather than supplying ad hoc scale values, for consistency :+1:

csong202:
It was more something that I noticed while messing around with the storybook controls. Yeah, that approach makes a lot more sense.

Stephen-ONeil:
@csong202 is this something you are planning to return to, or should this be closed/maybe archived as an issue?

Stephen-ONeil:
Related #1140

PR #1094, opened 2021-05-21
Fix pre-push hook failing in a way that is obscured by husky

Stephen-ONeil:
Does the trick, don't know if I should be mad at git for running pre-push hooks _after_ rejecting the push (and not documenting that this changes the arguments `read` by the hook), or at husky for swallowing up git's rejection error if the hook script exits unexpectedly.


ktw1016 (approved):
👍

PR #1095, opened 2021-05-21
Legends to Typescript

ktw1016:


client/src/charts/legends/LegendList.tsx
```diff
@@ -6,25 +6,20 @@ import { CheckBox, CheckBoxProps } from "src/components/CheckBox/CheckBox";
...
-  color?: string;
-}
+import "./LegendList.scss";
 
-interface LegendListProps {
+export interface LegendListProps {
   isHorizontal?: boolean;
-  items: LegendItemProps[];
+  items: LegendItemType[];
   onClick: Function;
```
Stephen-ONeil, line 16: Haven't looked at this too closely, but I think this should be something like `onClick: (event: React.MouseEvent) => void`

Stephen-ONeil: Or `onClick: React.ReactEventHandler<element>`?

client/src/components/CheckBox/CheckBox.tsx
```diff
@@ -10,10 +11,10 @@ import "./CheckBox.scss";
 
 export interface CheckBoxProps {
   disabled?: boolean; // greys out checkbox and make it non-clickable if true
-  onClick: Function; // (id) => {}
-  id: string | number;
+  onClick?: Function; // (id) => {}
```
Stephen-ONeil, line 14: `(id: string | number) => void`?

client/src/components/CheckBox/CheckBox.tsx
```diff
@@ -10,10 +11,10 @@ import "./CheckBox.scss";
...
+  id?: string | number;
   label: string; // required text for checkbox
-  active: boolean; // required, equivalent of 'checked'
+  active?: boolean; // required, equivalent of 'checked'
   color?: string; // defaults to primaryColor
   isSolidBox: boolean; // defaults to false
   container_style?: Object; // style for div, containing checkbox and label
```
Stephen-ONeil, line 20: Looks like this should be `React.CSSProperties`

Stephen-ONeil (changes requested):

client/src/components/CheckBox/CheckBox.tsx
ktw1016: Didn't know that existed!

Stephen-ONeil (approved):
:+1:

PR #1113, opened 2021-06-01
Add ssh to docker images, general image cleanup

Stephen-ONeil:
RE: 
image

Finally found a case where the CircleCI native git client's non-standard behaviour was a problem. Added ssh to build and test images (we already had git), checkout now uses standard git. Resolved the issue.

While I was at it, I cleaned up some legacy lines in the build docker image and updated the test image to use the build image as its base.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,8 +31,8 @@ jobs:
           path: "~/InfoBase"
       - restore_cache:
           keys:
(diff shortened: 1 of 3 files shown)


sranc046:
Looks good! Once this is merged I'll rebase the Cypress branch and see what can be done with dockerfiles so that all the required dependencies are still there. Might be able to extend the test image with cypress's or maybe using it in build (depending if it affects ressources)

sranc046 (approved):

PR #1114, opened 2021-06-01
Prettier upgrade + target more files

### LINKED ISSUE #1055: Let prettier format css/scss
It can, we just exclude them via our current `.prettierignore`. Can't think of a reason not to my self, did we have one?

Stephen-ONeil:
Upgrade prettier to 2.3.0, which produces new formatting output in certain cases. Remove some file types from the ignore list while at it (css/scss most notably), so that we only have the one big awkward prettier branch/subsequent rebasing to deal with.

Closes #1055


Stephen-ONeil:
For the record, I experimented with letting prettier handle yaml files too, but it's changes were mostly newlines at ends of files and lines where, from the diff, I couldn't tell the difference. Might investigate further in case it is worth letting it control those files too, but it seems very low value and also low confidence (it doesn't know that we have markdown in our yaml, possible it would mess with meaningful whitespace at end of lines, etc).

Stephen-ONeil:
Once both this and #1121 are merged, remember to remove `client/src/types/util_types.d.ts` from the .prettierignore!

sranc046 (approved):

sranc046:
Not much to really look over here but agree with the decision to let prettier format css and scss files.

Stephen-ONeil:
Paused, to be revived once some of the larger open branches are merged, to avoid the worst of the rebase headaches this'll result in. Won't hold off forever though.

PR #1115, opened 2021-06-02
Sobjs bugfix

### LINKED ISSUE #1013: Sobjs Bar chart height abnormally small when single bar

IvanFeng1:
Regarding issue #1013 

The height of the horizontal bar is proportional to  the number of entries in the graph and the value for the height for when there is only one entry is very low. My fix is I added a flat number to the height so it is not too small when there is only one entry 

Screen Shot 2021-06-02 at 12 12 08 PM

closes #1013


client/src/panels/panel_declarations/finances/detailed_program_spending_split/detailed_program_spending_split.js
```diff
@@ -294,7 +294,7 @@ class DetailedProgramSplit extends React.Component {
 
     // Increase height of the graph region for y-axis labels to have sufficient room
     // This is required to corretly display the labels when too many programs are present
-    const divHeight = 650 * (graph_ready_data.length / 30) * 2;
+    const divHeight = 650 * (graph_ready_data.length / 30) * 2 + 40;
```
Stephen-ONeil, line 297: This calculation, even before the `+ 40` feels dangerously like a "magic number" situation. It's obviously kind of a hack, since it was necessary to explain it with a comment in the first place, and it's extra hacky because it's unclear if the individual numbers have special meaning or not. 

The fact that the previous author didn't simplify the equation and left the `650 *`, `/ 30`, and `* 2` as separate factors kinda makes it seem like those numbers are special/should mean something... Did you happen to notice if 650, etc, showed up anywhere else in the file? Maybe in the horizontal bar graph component itself? May or may not be a mystery we can solve, they could have really just been the "magic numbers" that happened to make it look right at the time...

Anyway, even if we can't figure out and clean up the existing code, I wanted to suggest using `Math.min` to set the lower bound for the `divHeight`. Something like
```
const divHeight = Math.min(
  650 * (graph_ready_data.length / 30) * 2,
  80, // "magic" minimum height for the graph to render as desired
);
```
We still have the old hacky/magic number calculation for the non-minimum case, but it fixes the bug without making the problem worse by introducing more unexplained numbers to the existing calculation.

Stephen-ONeil (changes requested):
Fix works, nice! One fiddly comment on the line, mostly the fault of whoever wrote it before you but would be good to address it a bit here.

Small tip, if you include the phrases "closes #<issue number>" in a PR or commit message, then once that's merged in to master GitHub will automatically close the issue for you and tag it back to the closing PR/commit.

client/src/panels/panel_declarations/finances/detailed_program_spending_split/detailed_program_spending_split.js
Stephen-ONeil: My choice of 80 there is still magic, because it's just rounding off `650 * (1 / 30) * 2 + 40`. The point is to not distract/confuse future devs further by leaving it up to them to try and figure out what the `+ 40` means, haha.

IvanFeng1: One of the values used in calculating divHeight is used once again for calculating legendOffsetY. I'll update the divHeight calculation with your suggestion.

client/src/panels/panel_declarations/finances/detailed_program_spending_split/detailed_program_spending_split.js
```diff
@@ -294,8 +294,11 @@ class DetailedProgramSplit extends React.Component {
 
     // Increase height of the graph region for y-axis labels to have sufficient room
     // This is required to corretly display the labels when too many programs are present
-    const divHeight = 650 * (graph_ready_data.length / 30) * 2;
-
+    // const divHeight = 650 * (graph_ready_data.length / 30) * 2 + 40;
```
Stephen-ONeil, line 297: Delete please

client/src/panels/panel_declarations/finances/detailed_program_spending_split/detailed_program_spending_split.js
```diff
@@ -402,6 +405,7 @@ class DetailedProgramSplit extends React.Component {
                             fontSize: "11px",
                           }}
                         >
+                          {console.log("")}
```
Stephen-ONeil, line 408: Delete please

Stephen-ONeil (changes requested):
Two stray lines to clean up, should be good to approve after that :+1:

Stephen-ONeil (approved):
:+1:  :tada:

PR #1117, opened 2021-06-03
Graph toggling bug

### LINKED ISSUE #1103: Bar graph legend interactions broken
I.e. toggling items on/off on graph. Still works on the line graphs I checked.

Examples:
  - http://dev.ea-ad.ca/master/index-eng.html#orgs/gov/gov/infograph/financial/.-.-(panel_key.-.-'gocographic)
  - http://dev.ea-ad.ca/master/index-eng.html#orgs/gov/gov/infograph/covid/.-.-(panel_key.-.-'covid_estimates_panel)
  
Note: just on the standard bars, the version used by the people graphs, that can switch between multiple graph types (both line and bar) still work.

sranc046:
Turns out there was only a slight issue with the StandardLegend props which wasn't properly using LegendCheckBoxProps so isSolid wasn't being checked and was defaulting to false. This made every bar graph that was not intended to be toggled seem like it should (because of the checkmark appearing in the legends rather than a solid box). 

I still converted the covid graph with a hacky lodash fix (would appreciate some insight on how to better handle this since my lodash is still rusty, was having issues getting the proper data format) but overall we might decide to just leave it as a non-toggling graph.

Closes #1103


client/src/panels/panel_declarations/covid/covid_estimates.js
```diff
@@ -512,6 +340,209 @@ class CovidEstimatesPanel extends React.Component {
...
+              items={legend_items}
+              isHorizontal={true}
+              onClick={(label) => {
+                !(
+                  this.state.selected.length === 1 &&
+                  this.state.selected.includes(label)
+                ) &&
+                  this.setState({
+                    selected: toggle_list(this.state.selected, label),
+                  });
```
Stephen-ONeil, line 481: Just checking my understanding, `SummaryTab` and `tab_content_configs` were moved to being properties of `CovidEstimatesPanel` so that `SummaryTab` could call `setState` on `CovidEstimatesPanel`, right? Were there other reasons?

sranc046: Yeah that’s all it was

client/src/panels/panel_declarations/covid/covid_estimates.js
```diff
@@ -512,6 +340,209 @@ class CovidEstimatesPanel extends React.Component {
       })
     );
   }
+
+  SummaryTab = ({ args: panel_args, data }) => {
```
Stephen-ONeil, line 344: I find it is a little awkward that `tab_content_configs` and `SummaryTab` are properties of `CovidEstimatesPanel` now, especially since the other tab components are not. I think this is something of a code smell to look out for, when you have to move big blocks of code around to change their scope access. Usually going to be a better way.

It's pretty common to need a child component to have both read and write access to some state from some parent component. The standard way to do this is to pass down the state alongside some sort of setter callback. Some benefits:
  - the callback limits the child component's control of the parent's state, can enforce validation etc
  - keeps components composable
    - parent and child only interact through the child's API. Either could potentially be used with any other parents or children that implement an equivalent API
    
The pattern is in use here, you can see it with the `YearSelectionTabs` component and `CovidEstimatesPanel`'s `on_select_year` method.

client/src/panels/panel_declarations/covid/covid_estimates.js
```diff
@@ -434,69 +311,20 @@ const ByMeasureTab = wrap_with_vote_stat_controls(
...
   constructor(props) {
     super(props);
     this.state = {
       loading: true,
       summary_by_fiscal_year: null,
       selected_year: _.last(props.panel_args.years),
+      selected: [
+        text_maker(`covid_estimates_stat`),
+        text_maker(`covid_estimates_vote`),
+      ],
```
Stephen-ONeil, line 324: See the other comment first, because it gives some useful general advice (but it overall a bit undercut by this comment).

Since only `SummaryTab` cares about this state (not needed across tabs like the year), it'd be easier to make `SummaryTab` a class component (or use hooks) and make it own this state directly. Right? Or am I missing something?

Stephen-ONeil (changes requested):
I didn't leave any comments on the lodash specifically just yet, will look over it more after these first suggestions are addressed :+1:

Stephen-ONeil (approved):
One small thing, but otherwise approved!

Before you merge, I'd prefer moving `SummaryTab` back inside `covid_estimates.js`. Mostly for consistency, since the other tab contents are still declared inside `covid_estimates.js` (and similar for `covid_expenditures.js`).

Stephen-ONeil:
Oh, and a quick reminder, if you include "closes #1103" in your PR, or in a git commit, GitHub will close that issue (and add reference links to/from it) once it merges in to master.

PR #1118, opened 2021-06-03
Converting js files to typescript

IvanFeng1:
Converted TagCloud.js and glossary_components.js to typescript. 

In `link_utils.tsx`, I changed '&&' to if statement because I had a type error in `glossary_components.tsx` when I tried to use a function imported from `link_utils.tsx` and using an if statement fixed this error. 

In glossary_components, I replaced ` href={is_a11y_mode ? glossary_href(id) : null}` with `href={glossary_href(id)}` because I kept getting syntax errors when I tried to use ternary operators inside the href. The `is_a11y_mode` boolean variable inside `href={is_a11y_mode ? glossary_href(id) : null}` is within a block that gets executed after `is_a11y_mode` is true so I removed it from the href.


client/src/components/glossary_components.tsx
```diff
@@ -11,27 +11,52 @@ import { is_a11y_mode } from "src/core/injected_build_constants";
...
+  id: string;
+  children?: React.ReactNode;
+  no_bottom_border?: boolean;
+}
+
+interface GlossaryIconProps {
+  id: string;
+  alternate_text?: string;
+  arrow_selector?: any;
+  inner_selector?: any; // i don't know how to do the types for the selectors
```
Stephen-ONeil, line 24: Look like they're totally unused, probably legacy props someone missed cleaning up while refactoring this component? Not used in `GlossaryIcon`, and no consumers of `GlossaryIcon` are trying to pass them in either. Safe to drop them :+1:

client/src/components/glossary_components.tsx
```diff
@@ -11,27 +11,52 @@ import { is_a11y_mode } from "src/core/injected_build_constants";
...
+const GlossaryTooltipWrapper: React.FC<GlossaryTooltipWrapperProps> = ({
+  id,
+  children,
+  no_bottom_border,
+}) =>
   is_a11y_mode ? (
     <a
-      href={is_a11y_mode ? glossary_href(id) : null}
+      href={glossary_href(id)}
       title={is_a11y_mode ? trivial_text_maker("glossary_link_title") : null}
```
Stephen-ONeil, line 43: Huh, strange. Given line 15, on lines 17 and 18 `is_a11y_mode` _must_ be true, don't think either of those ternaries were doing anything. Might as well drop the ternary on 18 (now 43) as well.

client/src/components/glossary_components.tsx
```diff
@@ -11,27 +11,52 @@ import { is_a11y_mode } from "src/core/injected_build_constants";
...
     >
       {children}
     </a>
   ) : (
     <span
       className="nowrap glossary-tippy-link"
-      style={no_bottom_border && { borderBottom: "none" }}
-      tabIndex="0"
+      style={{ borderBottom: no_bottom_border ? "none" : "initial" }}
```
Stephen-ONeil, line 50: Change in behaviour here, is that intentional? Setting the inline style to `borderBottom: "initial"` isn't the same as not setting a `borderBottom` style. 

Also, I didn't know this off the top of my head, but it looks like IE doesn't support the `initial` value https://developer.mozilla.org/en-US/docs/Web/CSS/initial#browser_compatibility

Stephen-ONeil (changes requested):
Mostly small comments from me

client/src/link_utils.js
```diff
@@ -19,7 +19,10 @@ const glossary_href = (subject_or_id, first_character = "#") => {
   const id = _.isString(subject_or_id) ? subject_or_id : subject_or_id.id;
   const is_valid_glossary_item = !_.isUndefined(GlossaryEntry.lookup(id));
 
-  return is_valid_glossary_item && `${first_character}glossary/${id}`;
+  // return is_valid_glossary_item && `${first_character}glossary/${id}`;
```
ktw1016, line 22: yup can remove the comment here

client/src/components/glossary_components.tsx
ktw1016: Supplement explanation: "initial" will be different from not setting it because `glossary-tippy-link` is setting `border-bottom`

IvanFeng1: Should I replace `"initial"` with `""`?

Stephen-ONeil: I believe that'd work, but only because it's an invalid `border-bottom` value and would be ignored in the browser. 

Something to look out for: inline styles will be picked over any CSS-applied styles, unless those CSS styles use `!important`. That's what's happening here, there's some CSS applying a `border-bottom` here but the inline `border-bottom: none` will overrule that when set. Otherwise, we want whatever the CSS says, in which case we can't set any inline `border-bottom` value.

Small side note, `!important` is dangerous and should only be used if absolutely necessary (e.g. to overrule styles on some library component that we can't directly edit). It's dangerous _because_ it throws out the normal rules (like "inline styles beat style sheet styles").

Given that, one of these would be better:

1) `style={{ ...(no_bottom_border && { borderBottom: "none" }) }}`

Kind of ugly, but it's a useful trick with object spread `...` to inline a conditional key in an object. If `no_bottom_border` is false, it short circuits to `{ ...(false) }`, and object spread of anything falsey will add nothing to the resulting object.

2) `style={no_bottom_border ? { borderBottom: "none" } : undefined}`
Probably better here, clearer that we're not setting any inline styles when `no_bottom_border` is false. I showed 1) first just because it's a handy trick in other situations (e.g. if there were multiple styles and only some needed to be set conditionally).

ktw1016 (approved):
👍

Stephen-ONeil (approved):
Seconding Taewan's approval :+1:

Issue #1009, opened 2021-04-13
Add storybook for Nivo charts


GarvinH:
Our `WrappedNivo` charts may contain custom features that we should document (and we might have custom charts):

For example, the line graphs have a special feature of becoming dashed when lines overlap each other

see: http://dev.ea-ad.ca/master/index-eng.html#orgs/dept/186/infograph/financial

![image](https://user-images.githubusercontent.com/25855114/114571755-e47c6980-9c44-11eb-947d-1eb2d80929d2.png)

GarvinH:
`WrappedNivoCircleProportion` might be another chart we want to add(?)

Although I think this one might be a bit harder than the line graph for example

GarvinH:
`GraphOverlay` is another component that is pretty much used only with nivo charts

PR #1120, opened 2021-06-04
Nivo Storybook Components

csong202:
Issue: #1009 

Note: NivoLineBarToggle and WrappedNivoBar have already been reviewed, approved, and merged into master

Completed Components (checked ones have been reviewed and approved)
- [ ] WrappedNivoCircleProportion
- [ ] WrappedNivoLine
- [ ] WrappedNivoPie
- [ ] WrappedNivoTreemap


client/src/charts/wrapped_nivo/WrappedNivoLine/WrappedNivoLine.stories.js
```diff
@@ -0,0 +1,77 @@
...
+      { x: "Year 5", y: 2769 },
+      { x: "Year 6", y: 1111 },
+    ],
+  },
+];
+
+const raw_graph_data = _.chain(graph_data)
+  .map((row) => _.map(row.data, (point) => point.y))
+  .flatten()
+  .value();
```
ktw1016, line 61: It doesn't matter and I'm not requesting for change here, but you could simply use a `_.flatMap`

client/src/charts/wrapped_nivo/WrappedNivoLine/WrappedNivoLine.stories.js
```diff
@@ -0,0 +1,77 @@
...
+
+const colors = scaleOrdinal().range(newIBCategoryColors);
+
+export const Basic = Template.bind({});
```
ktw1016, line 65: I like that you're including a overlapping usecase as well. You should separate that one to its own story though (So group 1 and 2 could be its own story)

client/src/charts/wrapped_nivo/WrappedNivoTreemap/WrappedNivoTreemap.stories.js
```diff
@@ -0,0 +1,108 @@
...
+      total: graph_data_total_amt,
+      value: graph_data_values[7],
+    },
+  ],
+  color: "white",
+  name: "root",
+};
+
+const sorted_children = _.reverse(
+  _.sortBy(graph_data.children, function (o) {
```
ktw1016, line 84: we do prefer using arrow functions in all cases if possible and more descriptive naming (instead of `o`) I'll also show all equivalent statements below here
`_.sortBy(graph_data.children, (child) => { return child.value }`)
`_.sortBy(graph_data.children, (child) => child.value`)
`_.sortBy(graph_data.children, "value"`)

ktw1016: Using lodash chaining here would also be good practice:
`_.chain(graph_data.children).sortBy(...).reverse()`

client/src/charts/wrapped_nivo/WrappedNivoTreemap/WrappedNivoTreemap.stories.js
```diff
@@ -0,0 +1,108 @@
...
+    return o.value;
+  })
+);
+
+const domain = [];
+_.forEach(sorted_children, function (child) {
+  _.concat(domain, child.value);
+});
```
ktw1016, line 92: Gotta warn you before the for loop police show up ([it's in our conventions](https://github.com/TBS-EACPD/infobase/blob/master/docs/getting-started/js-tips-style-guide.md#javascript-coding-conventions))
[First google search on why not to use for loops
](https://medium.com/python-pandemonium/never-write-for-loops-again-91a5a4c84baf)

Before the "actual" fix though, `domain` will be an empty array for 2 different reasons.
1. `_.concat` is **not** changing the contents inside `domain`. (No mutation). So in order for this `concat` logic to work, you'd need to do `domain = _.concat(domain, child.value)`. Which brings me to 2nd reason
2. Since you declared `domain` as **`const`**,  you'll get an error because `domain` is constant! So to make this code work, you need
```
let domain = [];
_.forEach(sorted_children, (child) => {
  domain = _.concat(domain, child.value)
}
```

But as I mentioned, we wanna try to avoid for loops as much as possible. So instead, you could simply do (both equivalent):
`_.map(sorted_children, (child) => child.value)`
`_.map(sorted_children, "value")`

ktw1016 (changes requested):
Few comments, otherwise will be good to go !

ktw1016 (approved):
👍

PR #1121, opened 2021-06-04
Util types for special Omit cases

Stephen-ONeil:
As encountered [here](https://github.com/TBS-EACPD/infobase/pull/1071#pullrequestreview-676613323), using the built in [Omit util](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys) on a something that has any index types on it will give an unexpected result... Although they say it's [the expected behaviour](https://github.com/microsoft/TypeScript/issues/31153#issuecomment-487679270) regardless :shrug: .

Scrounged up a handful of useful util types from the various discussions of this issue. We'll probably want to use `IndexTypeSafeOmit`\* pretty much anywhere we need to omit keys, since it doesn't have the fiddly gotcha of the built-in.

\* names in this branch feel kinda awkward, would welcome any better suggestions!

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,5 +1,8 @@
 node_modules
 
+# TEMPORARY... prettier 2.0 doesn't understand TypeScript's `as` keyword AND had a // prettier-ignore bug, ha
+client/src/types/util_types.d.ts
+
 # generated stuff
 client/app/
 client/build/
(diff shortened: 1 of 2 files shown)


Stephen-ONeil:
Alternatively, and I haven't looked for this yet, there's probably libraries of util types out there already and I bet they'd cover the implementation of something like this for us. Likely preferable to maintaining these utils our self, they're a little tricky to read.

ktw1016 (approved):
Will try it out on my branch👍

PR #1122, opened 2021-06-07
Components to Typescript

IvanFeng1:
* Panel
* LogInteractionEvents
* CountdownCircle


IvanFeng1:
https://github.com/TBS-EACPD/infobase/blob/a2f36cbd3ac2e87b6b2475cb986862875ed2100a/client/src/core/ensure_loaded.ts#L426-L433

I am currently getting an error on Line 433 because `.uniqBy()` requires a parameter but the javascript code had no parameter for this function. My workaround for this locally is making  parameters not required for `.uniqBy()`  in the function definition, which is in the node_modules but I don't know how to fix this for the remote repo.

Stephen-ONeil:
The default `uniqBy` function is `_.identity` https://lodash.com/docs/4.17.15#uniqBy. That'll be the case for most lodash utils that take a predicate function. A little annoying the lodash types doesn't account for that.

client/src/components/LogInteractionEvents.tsx
```diff
@@ -3,7 +3,14 @@ import React from "react";
 
 import { log_standard_event } from "src/core/analytics";
 
-export const LogInteractionEvents = ({
+interface LogInterationEventsProps {
+  event_type: string;
+  event_details: string;
+  style: { display: string };
```
ktw1016, line 9: You can use `React.CSSProperties` here

client/src/components/LogInteractionEvents.tsx
```diff
@@ -3,7 +3,14 @@ import React from "react";
...
+  children: React.ReactElement;
```
ktw1016, line 10: ReactNode over ReactElement I think

client/src/components/Panel.tsx
```diff
@@ -14,7 +14,59 @@ import "./Panel.scss";
...
+  is: (comparator: any) => boolean;
+  legal_title: string;
+  level: string;
+  lookup: Function;
+  name: string;
+  plural: string;
+  singular: string;
+  subject_type: string;
+}
+interface Footnote {
```
ktw1016, line 49: I would rather you drop converting Panel.js for now for the same reason as ensure_loaded.
These footnotes, subject should be built from the ground up

ktw1016 (changes requested):

ktw1016 (approved):
👍👍

PR #1126, opened 2021-06-08
remove "narrowSymbol" from intl.NumberFormat as it was breaking speci…

ktw1016:
…fic safari versions

### DIFF
--- a/client/src/core/format.ts
+++ b/client/src/core/format.ts
@@ -32,7 +32,6 @@ const money_formatter = {
       new Intl.NumberFormat("en-CA", {
         style: "currency",
         currency: "CAD",
-        currencyDisplay: "narrowSymbol",
         minimumFractionDigits: ix,
         maximumFractionDigits: ix,
       })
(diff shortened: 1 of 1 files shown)


Stephen-ONeil:
For additional context, `currencyDisplay: "narrowSymbol",` was added in #1069 due to an unwanted cross-browser addition of currency symbols in French dollar formats.

Safari seems to have generally broken `Intl.NumberFormat` support between 14.0 and 14.1.0 ([here's a different instance of it](https://community.atlassian.com/t5/Trello-questions/Trello-stuck-at-loading-after-Safari-14-1-update-on-macOS-Mojave/qaq-p/1675577#M45687)). Works before 14.0 and seems fixed as of 14.1.1.

(that seems contradicted by this, maybe more research is needed, but it at least didn't cause a hard crash pre-14
image
)

Stephen-ONeil:
Ah, just double checked, this was broken in IE11 too, ha. Totally missed that when I added it.

Stephen-ONeil:
From my PC (was testing IE), haven't checked other chromes yet.
image

ktw1016:
Yeah just checked on IE on my personal desktop, mine is formatted correctly consistent with other browsers. No CA at the end

Stephen-ONeil:
Sorry, meant to clarify that I was on my PC testing IE, but also checked chrome on windows which is where I took that screenshot. I haven't tested chrome on my other devices, but expect it is the same.

Issue #408, opened 2019-12-06
Add FAQ panel to subapps


cookkate:
Things like the estimates comparison could really use a collapsible Q&A panel similar to the one on the infographics. There are several Q&As that explain how estimates work, what the difference between voted and statutory authorities are, etc.

PR #1131, opened 2021-06-10
FAQ Panel for Subapps

csong202:
Issue #408 

- all faq yaml files are in `src/common_text/faq`
- removed `SomeThingsToKeepInMind`
- PinnedFAQ.js: component that is used by curried_render and the subapps
- Consumers pass in an array of faq key pairs and a text maker component
  - faq key pairs are in the form `[<text_key>_q, <text_key>_a]`
  - consumers must provide the text maker component so webpack doesn't need to load all the faq yaml files every time PinnedFAQ is used 
- added background_color props to `AutoAccordion` 

Subapps with FAQ panels:
- Estimates Comparison
- Tag Explorer


client/src/subapps_key_concept_panels/FAQPanel.js
```diff
@@ -0,0 +1,30 @@
...
+import {
+  create_text_maker_component,
+  KeyConceptList,
+} from "src/components/index";
+
+import est_lang from "src/EstimatesComparison/estimates_comparison_questions.yaml";
+
+const { TM } = create_text_maker_component([est_lang]);
+
+export class FAQPanel extends React.Component {
```
ktw1016, line 15: Why not use the key concept panel? It should be to reuse the `curried_render` in key_concept_panel

ktw1016 (changes requested):
Screen Shot 2021-06-11 at 11 12 11 AM

I think there's too much of those bright blue and make it look a bit messy. What about adding color prop and making accordion darker color (primary color)? Or anything else to make it look cleaner
Also I think accordion initially expanded would be better

client/src/subapps_key_concept_panels/subappps_key_concept_panels.js
```diff
@@ -0,0 +1,26 @@
+import { declare_panel } from "src/panels/panel_declarations/common_panel_utils";
+import {
+  curried_render,
+  common_panel_config,
+} from "src/panels/panel_declarations/misc/key_concept_panels/key_concept_panels";
+
+import { primaryColor } from "src/core/color_defs";
+
+export const declare_estimates_comparison_key_concepts_panel = () =>
+  declare_panel({
```
ktw1016, line 10: Hmm this is interesting take, I was thinking more like separate 
```
<SomeThingsToKeepInMind>
  <KeyConceptList>
...
```
from `curried_render`, and define it as its own component. Then use that component inside for example `curried_render` and other places (like estimates explorer). I think defining panels for these is a bit overkill especially since it's not part of infographics, or need any utils from panel component. It'd clean this code up a bit. Let me know if you have different opinion and have any questions !

ktw1016:
Design wise looks good to me personally !

client/src/panels/panel_declarations/misc/key_concept_panels/key_concept_panels.js
```diff
@@ -26,23 +28,23 @@ const { TM } = create_text_maker_component([
...
+export class FAQPanel extends React.Component {
+  render() {
+    const {
+      rendered_q_a_keys,
+      is_initially_expanded,
+      background_color,
+      subject,
+    } = this.props;
+
+    let subj = _.isUndefined(subject) ? Gov : subject;
```
ktw1016, line 47: You can do this for default props:
```
class FAQPanel extends React.Component {
  ...
}
FAQPanel.defaultProps = {
  subject: Gov,
}
```

ktw1016: While I think this is probably fine, adding props like `q_args`, `a_args` would give more flexibility
Alternatively, you can also change the structure of `q_a_keys =[ {key: "text_key", q_args: { Gov }, a_args: { Gov } }, {...} ]`

ktw1016: Upto you what you want of those

ktw1016: Oh, and I think this can move to `components`

csong202: Should I storybook it too?

ktw1016: sure

client/src/common_text/common_questions.yaml
```diff
@@ -0,0 +1,21 @@
+common_faq_q:
```
ktw1016, line 1: So there's another file named the exact same. I think file should be renamed to something like ` common_faq_questions`. Also I think this would fit better along with faq in directory structure

client/src/components/FAQPanel/FAQPanel.yaml
```diff
@@ -0,0 +1,20 @@
+question1_q:
```
ktw1016, line 1: o wait what's the difference from this file to common_questions.yaml?

client/src/components/FAQPanel/FAQPanel.js
```diff
@@ -0,0 +1,72 @@
...
+import { Gov } from "src/models/organizational_entities";
+
+import common_subapp_lang from "src/common_text/common_questions.yaml";
+
+import est_lang from "src/EstimatesComparison/estimates_comparison_questions.yaml";
+import igoc_lang from "src/IgocExplorer/igoc_explorer_questions.yaml";
+import table_picker_lang from "src/rpb/table_picker_questions.yaml";
+import tag_exp_lang from "src/TagExplorer/tag_explorer_questions.yaml";
+import text_diff_lang from "src/TextDiff/text_diff_questions.yaml";
+import treemap_lang from "src/TreeMap/treemap_questions.yaml";
```
ktw1016, line 28: I do wonder if all of these yaml should co-exist with FAQPanel.js. Any thoughts on that?

ktw1016:
I think a lot of it is personal preference but in TextDiff, I think FAQ can come up above the controls. So above the drop down menus. Just so there's kind of a separation between controls and explanation.. And not sure about the FAQ on rpb table picker. I guess that's upto team to decide on, if anyone will have anything useful to put in there!

client/src/components/FAQPanel/FAQPanel.yaml
csong202: This is the one I made for the FAQPanel storybook

client/src/components/FAQPanel/FAQPanel.js
csong202: I think that's a good idea. I guess there isn't much point to keep them in their folders with the subapp

client/src/components/FAQPanel/FAQPanel.yaml
ktw1016: ohh👍

client/src/components/FAQPanel/FAQPanel.js
```diff
@@ -0,0 +1,74 @@
...
+  tag_lang,
+  est_lang,
+  igoc_lang,
+  table_picker_lang,
+  tag_exp_lang,
+  text_diff_lang,
+  treemap_lang,
+  common_subapp_lang,
+  sample_lang,
+]);
```
Stephen-ONeil, line 44: I think either it should be the consumer's responsibility to provide a `text_maker_function` to this component, or this component _and_ all the FAQ content should be lifted up to it's own directory under source (because in that case it's more of a standalone system than a reusable component).

client/src/components/FAQPanel/FAQPanel.js
```diff
@@ -0,0 +1,74 @@
...
+
+export class FAQPanel extends React.Component {
```
Stephen-ONeil, line 46: Nitpick, but "panel" loses meaning outside of the infographic context, I'd suggest renaming this to something more generic like `StandardFAQ` or something.

client/src/components/FAQPanel/FAQPanel.js
Stephen-ONeil: The former's more case-by-case boiler plate code, but it keeps this component more re-usable and is a small benefit to webpack bundling (only load the text as needed, instead of loading it all as soon as you need any FAQ).

The later is a little easier to make use of in general, and could be seen as a step towards moving FAQ text from hardcoded yaml to, potentially, something like an externally managed data source processed through the pipeline. 

I'm leaning towards the first, since moving FAQ items to pipeline data is not something we're settled on ever doing (at least not short term, the pipeline has enough work to handle).

client/src/components/FAQPanel/FAQPanel.js
```diff
@@ -0,0 +1,74 @@
...
+
+    return (
+      <SomeThingsToKeepInMind
+        is_initially_expanded={is_initially_expanded}
+        background_color={background_color}
+      >
+        <KeyConceptList
+          question_answer_pairs={_.map(rendered_q_a_keys, (base_text_key) => [
+            <TM key={"q"} k={base_text_key + "_q"} args={{ subject }} />,
+            <TM key={"a"} k={base_text_key + "_a"} args={{ subject }} />,
```
Stephen-ONeil, line 63: This is an old pattern, and is probably something of an anti-pattern (have an issue about this type of dynamic text maker key, explain more there #1128). You don't have to make cleaning it up part of this PR, but cleaning it up would sort of allow for an alternative to one of the approaches I suggested above. Instead of passing in a text maker function alongside the `rendered_q_a_keys` prop, the consumer could just be expected to provide a prop with type `[React.ReactNode, React.ReactNode][]` containing actual question/answer content.

Stephen-ONeil (changes requested):
See comments below, additionally:
  - [x] clean up the commented out lines from the other files before it's set to merge
  - [x] consider porting the new FAQ component to TS
    - only if other changes remove dependency on subjects
  
  Thanks!

csong202:
> * [ ]  consider porting the new FAQ component to TS

I tried earlier but I can't really do it until `ts-subject-entities` is merged because I need a type for the `subject` props

Stephen-ONeil:
> > * [ ]   consider porting the new FAQ component to TS
> 
> I tried earlier but I can't really do it until `ts-subject-entities` is merged because I need a type for the `subject` props

Ah, right. It's possible you could lose the need for a subject prop if you choose to take one of my optional suggestions (pass in content directly, instead of keys/subjects/etc). If the dependency on subjects remains though, no worries on not porting to TS.

client/src/components/KeyConceptList/KeyConceptList.tsx
```diff
@@ -6,7 +6,7 @@ import "./KeyConceptList.scss";
 
 interface KeyConceptListProps {
   question_answer_pairs: [React.ReactNode, React.ReactNode][];
-  compact: boolean;
+  compact?: boolean;
```
csong202, line 9: before I made `compact` optional, I got an error in `StandardFAQ.tsx` telling me that `KeyConceptList` was missing `compact`

Any ideas for a better approach appreciated

client/src/components/StandardFAQ/StandardFAQ.tsx
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+
+import { SomeThingsToKeepInMind } from "src/panels/panel_declarations/common_panel_components";
```
Stephen-ONeil, line 3: Since this is moving from an infographic-specific thing to a site-wide thing, `SomeThingsToKeepInMind` should be removed from `common_panel_components`. That's essentially what this component should be directly.

In practice, `SomeThingsToKeepInMind` in the infographics always contains a `KeyConceptList` so the existing use cases of it there should be directly replaced with this FAQ component.

client/src/components/StandardFAQ/StandardFAQ.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+import { secondaryColor } from "src/style_constants/colors.interop.scss";
+
+type StandardFAQProps = {
+  faq_content: React.ReactNode[][]; // TODO: change to [React.ReactNode, React.ReactNode][] when faq_utils gets converted to ts
+  is_initially_expanded?: boolean;
+  background_color?: string;
+} & typeof StandardFAQ.defaultProps;
+
+export class StandardFAQ extends React.Component<StandardFAQProps> {
```
Stephen-ONeil, line 15: Hmm, "standard" is a bit vague. Maybe name this `PinnedFAQ` or even `PinnedFAQDrawer`?

client/src/components/StandardFAQ/StandardFAQ.tsx
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+
+import { SomeThingsToKeepInMind } from "src/panels/panel_declarations/common_panel_components";
+
+import { KeyConceptList } from "src/components/KeyConceptList/KeyConceptList";
+
+import { secondaryColor } from "src/style_constants/colors.interop.scss";
+
+type StandardFAQProps = {
+  faq_content: React.ReactNode[][]; // TODO: change to [React.ReactNode, React.ReactNode][] when faq_utils gets converted to ts
```
Stephen-ONeil, line 10: The `faq_content` prop could be called `question_answer_pairs` imo, to keep it clearer that it's a straight pass through to the `KeyConceptList` prop of the same nam.

client/src/TagExplorer/TagExplorerComponent.js
```diff
@@ -2,14 +2,22 @@ import classNames from "classnames";
 import _ from "lodash";
 import React, { Fragment } from "react";
 
+import common_lang from "src/panels/panel_declarations/misc/key_concept_panels/common_questions.yaml";
```
Stephen-ONeil, line 5: Any FAQ yaml files that are actually shared between infographics and other routes should be lifted up to a higher level in the folder tree, maybe put them in somewhere like `src/common_text/faq`.

Small nitpick, but there's a `common_lang.yaml` in `src/common_text` already, importing this as `common_lang` might be confusing at a glance. Maybe generally import the faq yaml as `<prefix>_faq` instead of `<prefix>_lang` or something like that, to differentiate them at a glance.

client/src/components/StandardFAQ/faq_utils.js
```diff
@@ -0,0 +1,19 @@
...
+  text_bundles,
+  q_a_base_keys,
+  subject = Gov
+) => {
+  const { TM } = create_text_maker_component(text_bundles);
+
+  return _.map(q_a_base_keys, (base_text_key) => [
+    <TM key={"q"} k={base_text_key + "_q"} args={{ subject }} />,
+    <TM key={"a"} k={base_text_key + "_a"} args={{ subject }} />,
+  ]);
```
Stephen-ONeil, line 18: I'd kind of prefer consumers of the FAQ component to explicitly declare all their question answer pairs. For one, there's #1128, for another it will make the text args more flexible/clearer. Not all FAQs need a subject arg, and what if a future item needs args other than subjects, or if a given FAQ drawer needs to pass different subjects to different Q/A pairs? Made sense that all items would always be related to a given subject when the FAQs were just an infographic thing, but arbitrary routes outside the infographics don't have the same single-subject scope.

BUT, I'm fine with dropping this pattern being a follow up PR or issue, just to move this PR along (and because I'll admit I gave confusing feedback in my last review when I threw out three different solutions to this problem, haha, sorry) .

Stephen-ONeil (changes requested):
Mostly in the cleanup and reorg phase, hopefully not many more review cycles after this :+1:

client/src/components/PinnedFAQ/PinnedFAQ.js
```diff
@@ -0,0 +1,41 @@
...
+import { SomeThingsToKeepInMind } from "./faq_utils";
+
+export class PinnedFAQ extends React.Component {
+  render() {
+    const {
+      q_a_key_pairs,
+      TM,
+      is_initially_expanded,
+      background_color,
+      subject,
```
csong202, line 19: keeping the subject props for now to move this PR along

client/src/EstimatesComparison/EstimatesExplorerComponent.js
```diff
@@ -11,15 +12,23 @@ import {
   CheckBox,
   Details,
 } from "src/components/index";
+import { create_text_maker_component } from "src/components/misc_util_components";
+
+import common_subapp_faq from "src/components/PinnedFAQ/common_faq_questions.yaml";
+import est_faq from "src/components/PinnedFAQ/estimates_comparison_questions.yaml";
```
csong202, line 18: Keeping the subapp-specific faq yaml files in PinnedFAQ because they shouldn't be needed in the infographics

Stephen-ONeil: Should they move to the specific sub app's dir then?

Stephen-ONeil: Or, since this yaml file's still empty, could just delete it. I'd prefer cleaning up those placeholder files before merging.

client/src/components/PinnedFAQ/faq_utils.js
```diff
@@ -0,0 +1,49 @@
+import classNames from "classnames";
+
+import React from "react";
+
+import { ButtonToolbar } from "react-bootstrap";
+import MediaQuery from "react-responsive";
+
+import { SOME_THINGS_TO_KEEP_IN_MIND_STORAGE_KEY } from "src/panels/panel_declarations/common_panel_components";
```
Stephen-ONeil, line 8: This should stay living alongside the `SomeThingsToKeepInMind` component.

client/src/components/PinnedFAQ/PinnedFAQ.js
```diff
@@ -0,0 +1,41 @@
+import _ from "lodash";
+import React from "react";
+
+import { KeyConceptList } from "src/components/KeyConceptList/KeyConceptList";
+
+import { Gov } from "src/models/organizational_entities";
+
+import { secondaryColor } from "src/style_constants/colors.interop.scss";
+
+import { SomeThingsToKeepInMind } from "./faq_utils";
```
Stephen-ONeil, line 10: I think `SomeThingsToKeepInMind` doesn't need to be a standalone component now, I don't think there's anything using it directly (only via the `PinnedFAQ`). The I think merging it directly with `PinnedFAQ` would be good.

client/src/components/PinnedFAQ/PinnedFAQ.js
Stephen-ONeil: Ok, I'll open the follow up issue or knock it out my self once this merges

Stephen-ONeil (changes requested):
Just a bit more cleanup, and reminder to delete all those placeholder yamls.

client/src/TagExplorer/TagExplorerComponent.js
```diff
@@ -308,13 +336,22 @@ export default class TagExplorerComponent extends React.Component {
...
+                        "MtoM_tag_warning_double_counting_q",
+                        "MtoM_tag_warning_double_counting_a",
+                      ],
                     ],
-                    (key) => [
-                      <TM key={key + "_q"} k={key + "_q"} />,
-                      <TM key={key + "_a"} k={key + "_a"} />,
+                    (pair) => [
+                      <TM key={pair[0]} k={pair[0]} />,
+                      <TM key={pair[1]} k={pair[1]} />,
```
Stephen-ONeil, line 354: Small note on this pattern, which shows up somewhere else in this PR too I think, you can use array destructuring (much like object destructuring) to declare array elements as named variables. Very useful in function signatures, would look like this here `([q_key, a_key]) => `. Makes it easier to write clean code with clear names in cases like this :+1:

Stephen-ONeil (approved):
Approved :tada: 

I'm going to open up two follow up issues for related but not necessary cleanup. Up to you if you want to self-assign them or not but either way they aren't priority :+1:

PR #1134, opened 2021-06-10
Service Inventory refactor to have 1 collection per subject level

### LINKED ISSUE #1100: Refactor service inventory API to use less MongoDB collections
There's a limit of 500 collections across all DBs on the plan used to host our dev links, service inventory currently uses 28 (of the total 45) collections per DB. Service inventory population is currently disabled, but unless that number is reduced then we'll be capped at only ~10 concurrent active dev links once it is enabled.

Upgrading the MongoDB Atlas tier isn't an option, would need to go up two tiers and drastically increase the cost to get past the (arbitrary?) 500 collection limit, and we don't need any of what we'd _actually_ be paying for (disk space, speed) from that.

Only alternative I could think of is to do something hacky like open a bunch of free tier projects and round-robin the dev link DBs between them.

ktw1016:
closes #1100

### DIFF
--- a/server/src/models/services/models.js
+++ b/server/src/models/services/models.js
@@ -148,6 +148,19 @@ export default function (model_singleton) {
     ...bilingual_str("service_name"), // only for dept, program
     website_visits_count: { type: Number },
   });
+  const GovServiceSummarySchema = mongoose.Schema({
+    id: pkey_type(),
+    service_general_stats: ServiceGeneralStatsSchema,
+    service_type_summary: [ServiceTypeSummarySchema],
+    service_digital_status_summary: [ServiceDigitalStatusSummarySchema],
+    service_id_methods_summary: [ServiceIdMethodsSummarySchema],
+    service_standards_summary: [ServiceStandardsSummarySchema],
+    service_fees_summary: [ServiceFeesSummarySchema],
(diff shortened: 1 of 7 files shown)


server/src/models/services/models.js
```diff
@@ -148,120 +148,68 @@ export default function (model_singleton) {
...
+    top_services_application_vol_summary: [
+      TopServicesApplicationVolSummarySchema,
+    ],
+    top_services_website_visits_summary: [
+      TopServicesWebsiteVisitsSummarySchema,
+    ],
+    service_fees_summary: [ServiceFeesSummarySchema],
+    service_standards_summary: [ServiceStandardsSummarySchema],
+  });
+  const ProgramServiceSummarySchema = mongoose.Schema({
```
ktw1016, line 179: Could simply combine org and program as they're equivalent but just leaving it separately as dept and program may diverge in the future

Stephen-ONeil: Seems good to me. Maybe just to make it clearer at a glance what's common between these three schemas, as well as between the org and program schemas in particular, ou could define the common fragments above and just spread them in here?

Stephen-ONeil (changes requested):
Two small comments, but feel free to push back on either if you don't agree.

Stephen-ONeil (approved):
:+1:

Issue #1059, opened 2021-05-03
For April 2022 - evaluate treemap route


Stephen-ONeil:
Close call between  #876 and #1058. If the treemap hasn't received a good amount of polish and/or gets a minimal amount of traffic, by April 2022, then it's getting the axe.

If there's a good enough reason, I think it could even go sooner than that.

PR #1135, opened 2021-06-11
TreeMap javascript files to typescript

IvanFeng1:
- [ ] data
- [ ] TreeMap
- [x] TreeMapControls
- [x] TreeMapInstructions
- [x] TreeMapLegend
- [x] TreeMapSidebar
- [x] TreeMapTopbar
- [ ] TreeMapViz


Stephen-ONeil:
Got a merge commit there, is that because git suggested you do a pull when you tried to push, after rebasing off master?

IvanFeng1:
> Got a merge commit there, is that because git suggested you do a pull when you tried to push, after rebasing off master?

I haven't rebased off master, I am doing the merge because I deleted a file directly on github because git wouldn't detect the difference between two files with the same name but different cases for a letter.

IvanFeng1:
https://github.com/TBS-EACPD/infobase/blob/331927519d33c71ebced9ac5ef68fd4892ba45ee/client/src/TreeMap/TreeMapControls.tsx#L163-L166

I used the props I defined in `TreeMapSideBarProps` again in `TreeMapControls` and I realize now this is probably not good practice. Should I define `TreeMapSideBarProps` in another file and the import it into `TreeMapSidebar` and `TreeMapControls`? 

Another issue with the PR I think is that there is too many commits of me trying to make the naming for `TreeMapSidebar` and `TreeMapTopbar` consistent across the files in the folder so should I squash and merge the PR instead of rebase and merge? I haven't actually done squash and merge before.

Stephen-ONeil:
> > Got a merge commit there, is that because git suggested you do a pull when you tried to push, after rebasing off master?
> 
> I haven't rebased off master, I am doing the merge because I deleted a file directly on github because git wouldn't detect the difference between two files with the same name but different case for a letter.

Ah, ok, few tips:
  1) if your local was the same as remote, minus the deleted file, you do `git fetch origin branch` to update your local `origin/branch` copy, then run `git reset --hard origin/branch` from `branch`, to force your local `branch` to match the remote 
    - but I would hav expected the pull to just fast forward in that case?
  2) if your local and remote are out of sync but both contain new content, you can use `pull --rebase` to reconcile them without a merge commit
  3) as for the file name issue, it's because some file systems (like MacOS' bad file system) are themselves _not_ case sensitive. Most unix-like file systems, where git was born, _are_ case sensitive. On a non-case-sensitive file system, git has to ignore file case changes. To explicitly communicate to git that a file should change name/location, regardless of the file system, you can use the `git mv` utility (same as the standard unix `mv`, just also guarantees git will track the change)

IvanFeng1:
> https://github.com/TBS-EACPD/infobase/blob/331927519d33c71ebced9ac5ef68fd4892ba45ee/client/src/TreeMap/TreeMapControls.tsx#L163-L166
> 
> I used the props I defined in `TreeMapSideBarProps` again in `TreeMapControls` and I realize now this is probably not good practice. Should I define `TreeMapSideBarProps` in another file and import it into `TreeMapSidebar` and `TreeMapControls`? Or should I rename `TreeMapSideBarProps` in `TreeMapSideBar` and still import it from `TreeMapSideBar`?
> 
> Another issue with the PR I think is that there is too many commits of me trying to make the naming for `TreeMapSidebar` and `TreeMapTopbar` consistent across the files in the folder so should I squash and merge the PR instead of rebase and merge? I haven't actually done squash and merge before.

I think I am going to put all the type definitions for TreeMap in one folder

Stephen-ONeil:
> Another issue with the PR I think is that there is too many commits of me trying to make the naming for `TreeMapSidebar` and `TreeMapTopbar` consistent across the files in the folder so should I squash and merge the PR instead of rebase and merge? I haven't actually done squash and merge before.

If you want to drop/squash select commits, you can do that all at once with an interactive rebase. Rebasing at this point will also drop the merge commit and make the commit history linear instead. If you're ready to deal with the prettier changes too, run a `git rebase -i master` (after updating master from the remote).

client/src/TreeMap/TreeMapControls.tsx
```diff
@@ -4,11 +4,39 @@ import React, { Fragment } from "react";
...
+  children: React.ReactElement;
+}
+
+interface TreeMapRadioButtonsProps {
+  options: {
+    display: string;
+    id: string;
+    active: boolean;
+  }[];
+  onChange: Function;
```
Stephen-ONeil, line 37: I think this could be `(id: string) => void`

client/src/TreeMap/TreeMapControls.tsx
```diff
@@ -148,7 +180,7 @@ export class TreeMapControls extends React.Component {
       key,
       value
     );
-    if (location.pathname !== new_path) {
+    if (location && history && location.pathname !== new_path) {
       history.push(new_path);
     }
```
Stephen-ONeil, line 185: I think this would be a good place to use the `!` non-null operator. We're guaranteed that `location` and `history` will exist at this point, but checking them in the conditional might imply to future devs that they might not (when really that's just there to make TS happy).

There have been some cases where I've recommended type narrowing with a conditional to avoid having to spam `!`s all over the place. I _think_ I'd still advise that in some cases, but only if it's a longer code block and the `!`s are really fiddly/annoying.

client/src/TreeMap/TreeMapSidebar.tsx
```diff
@@ -0,0 +1,64 @@
...
+  createHref: Function;
+  go: Function;
+  goBack: Function;
+  goForward: Function;
+  length: number;
+  listen: Function;
+  location: LocationProps;
+  push: (arg: string) => void;
+  replace: Function;
+}
```
Stephen-ONeil, line 24: Should be able to get these types from `@types/react-router`, we've got that in our package.json already.

client/src/TreeMap/TreeMapTopbar.tsx
```diff
@@ -13,14 +13,36 @@ const text_maker = create_text_maker([treemap_text]);
...
+      hash: string;
+      pathname: string;
+      search: string;
+      state: undefined;
+    };
+    push: Function;
+    replace: Function;
+  };
+  org_route: [];
+  setRouteCallback: Function;
```
Stephen-ONeil, line 39: Looks like this is `(route: string, reset: boolean) => void`. A bit awkward to be setting that in a bunch of places though, could export the interface for `setRoute` from `TreeMap`, that should be shared between this and the other `setRouteCallback`s

client/src/TreeMap/TreeMapTopbar.tsx
```diff
@@ -13,14 +13,36 @@ const text_maker = create_text_maker([treemap_text]);
...
+    location: {
+      hash: string;
+      pathname: string;
+      search: string;
+      state: undefined;
+    };
+    push: Function;
+    replace: Function;
+  };
+  org_route: [];
```
Stephen-ONeil, line 38: I think this is an array of strings?

client/src/infographic/BubbleMenu.tsx
```diff
@@ -12,21 +12,33 @@ import { is_a11y_mode } from "src/core/injected_build_constants";
 
 import "./BubbleMenu.scss";
 
-const BubbleMenu = ({ items, active_item_id }) => {
+interface ItemProps {
```
Stephen-ONeil, line 15: Did the bubble menu stuff get mixed in from a different branch? If not, I'd recommend splitting it out from this branch, I feel like `TreeMap` TSing could drag on for a bit.

Stephen-ONeil: I won't review this file for now.

Stephen-ONeil (changes requested):
Some initial misc comments.

If I'd been paying more attention, I'd have warned you about the `TreeMap` sooner. It's some particularly messy legacy-ish code, was nearly dropped from the site recently and only survived on a limited basis because of push back within the team. See issue #1059

The reason I wanted to drop it is that
  1) it gets very little traffic
  2) it's UI/UX never received our standard level of polish
  3) most relevantly, it's low quality code and I'd specifically labeled it as being a big maintenance burden. I expected that it'd be better to refactor the whole thing before porting it to TS/writing tests for it (and, given 1 + 2, I didn't feel it was worth the refactor)
 
At this point though, you'd have a better sense of things than me. I never touched this code. Do you think it'd be easy to clean the code up? Would you be interested in doing that?

IvanFeng1:
> #1059



> Some initial misc comments.
> 
> If I'd been paying more attention, I'd have warned you about the `TreeMap` sooner. It's some particularly messy legacy-ish code, was nearly dropped from the site recently and only survived on a limited basis because of push back within the team. See issue #1059
> 
> The reason I wanted to drop it is that
> 
> 1. it gets very little traffic
> 2. it's UI/UX never received our standard level of polish
> 3. most relevantly, it's low quality code and I'd specifically labeled it as being a big maintenance burden. I expected that it'd be better to refactor the whole thing before porting it to TS/writing tests for it (and, given 1 + 2, I didn't feel it was worth the refactor)
> 
> At this point though, you'd have a better sense of things than me. I never touched this code. Do you think it'd be easy to clean the code up? Would you be interested in doing that?

I am not sure how the code should be refactored to be honest.  If this section of the website is probably going to axed, I'll work on something else then. Should I close this PR and delete this branch?

Stephen-ONeil:
> I am not sure how the code should be refactored to be honest. If this section of the website is probably going to axed, I'll work on something else then. Should I close this PR and delete this branch?

Putting aside whether it will get axed in a year or so, do you feel like you're close to finishing? Did it feel like it was providing value, to add types here? I feel like it might, since it would make the weird APIs on these components clearer, but that's just a guess. I've not read through it all in depth enough to answer that my self.

If the code gets ported to TS then that'd be one more thing in favour of eventually cleaning up the code and UI/UX, rather than letting it die. 

But, yeah, if you're leaning towards this not being worth finishing, then you can close the PR, delete the branch, and update the TS tracking file with a note that this route's not worth porting to TS at this time.

IvanFeng1:
Closing because I don't want to refactor files in the TreeMap folder

PR #1141, opened 2021-06-15
Converting FixedPopover.js to TS

patfire12345:


Stephen-ONeil:
You've got copies of some upstream commits from master here. Rebase again with `-i` in case you need to manually drop those duplicates.

client/src/components/modals_and_popovers/FixedPopover.tsx
```diff
@@ -43,22 +67,27 @@ export class FixedPopover extends React.Component {
...
       footer,
-      dialog_position,
+      dialog_position = "left",
       additional_dialog_class,
-      auto_close_time,
-      close_text,
-      close_button_in_header,
+      auto_close_time = false,
+      close_text = _.upperFirst(trivial_text_maker("close")),
+      close_button_in_header = false,
```
Stephen-ONeil, line 90: Shouldn't `on_close_callback = _.noop` (or equivalent) go here? Why is it up on line 70?

Did TS complain about the type of `_.noop` or something?

Stephen-ONeil (changes requested):
Should have labeled that above comment as a review/change request. Need to sort that out before moving.

Once the commits are untangled, I've only got the one comment on the file itself.

PR #1144, opened 2021-06-15
Drop SpinnerWrapper

### LINKED ISSUE #1140: Clean up SpinnerWrapper
`SpinnerWrapper` supports using both the actual `LeafSpinner` component and the non-react `spin.js` version... but we never use the `spin.js` option (the `use_leaf_spinner` prop is always left to the default of `true`). The API of `SpinnerWrapper` is also a bit messy as a result, it takes a `scale` prop which is actually only used if `use_leaf_spinner` is `false`.

Should just use `LeafSpinner` directly, drop `SpinnerWrapper`. First, check for any instances of `SpinnerWrapper`  trying to use `scale` prop, should at least drop that prop but may also want to double check whether the chosen `LeafSpinner` config should be tweaked.

Reminder: there _is_ one place where we use `spin.js`, for the pre-app bootstrapper spinner. Post-bootstrapper, everything should use react.

Stephen-ONeil:
Use `LeafSpinner` directly. Closes #1140

Bunch of additional leaf spinner related cleanup (both to the react and the pre-bootstrapper version).


client/src/components/LeafSpinner/LeafSpinner.js
```diff
@@ -3,8 +3,8 @@ import React from "react";
 
 import { primaryColor, secondaryColor } from "src/core/color_defs";
 
-import leaf_loading_spinner from "src/svg/leaf-loading-spinner.svg";
-import "src/components/SpinnerWrapper/LeafSpinner.scss";
+import leaf_loading_spinner from "./LeafSpinner.svg";
```
ktw1016, line 6: Are we putting SVGs across the site now?

ktw1016 (approved):

client/src/components/LeafSpinner/LeafSpinner.js
Stephen-ONeil: The svgs in `src/svg` are specifically copied over by build base so that they can be referenced via static URL (for `img` tags, in CSS, etc). This svg never was, it was always imported (and therefore inlined in to webpack bundles), never needed to be in `src/svg`.

The biggest reason for moving it was that the svg has a number of classes from `LeafSpinner.scss` hardcoded in it. So,
  a) it can't be used properly unless those styles are also loaded
  b) they need to be kept in sync
Given that, _not_ collocating them is unsafe.

We have other svgs outside of `src/svg`, just already inside JS or stored as yaml. This just happens to be an svg instead :shrug:

Stephen-ONeil:
Might as well port this to TS while I'm at it, will do that soon.

Stephen-ONeil:
Linter warning for `no-unused-vars`... inside a TypeScript interface. The fix is #1146, but that'll require fixing other linter issues in a LOT of other TS files, so out of scope for this PR.

ktw1016 (approved):
👍

PR #1147, opened 2021-06-16
More TypeScript lint rules

### LINKED ISSUE #1146: Extend with typescript-eslint recommended's in client/.eslintrc.json
Add 
```
    "plugin:@typescript-eslint/eslint-recommended",
    "plugin:@typescript-eslint/recommended"
```
to `client/.eslintrc.json` extends. 

This is going to raise ~2000 warnings and ~200 errors! Need to look at the rules throwing these errors, we'll want to disable some and properly fix others.

The `@typrscript-eslint/ban-types` in particular is really important, and something we should have enabled sooner. Been making a lot of mistakes during out TS ports that this will call out.

Stephen-ONeil:
Closes #1146

This is a big one, lots of our TS practices so far have been anti-patterns.

I'll try and keep track of some of the big ones here:
  -  [non-null assertions aren't recommended](https://github.com/typescript-eslint/typescript-eslint/blob/v4.22.0/packages/eslint-plugin/docs/rules/no-non-null-assertion.md), using them supposedly cancels outs the benefit of the `strictNullChecks` we set to true in the tsconfig?
    - right, duh, the point is that if your type says it could be null, then you really should be handling the null case explicitly (type narrowing the null out of the other branches). Using the non-null assertion freely leads to losing the safety in cases where there could be nulls. That's not the intent of the non-null assertion, it's only supposed to be used if you're really sure null values aren't possible, but it's always better to handle them explicitly (maybe with a comment saying, _at this point_ nulls should be impossible... but assuming they'll never be possible and not handling them is just setting up future run-time bugs)
  - I've left the warning against using explicit `any` types on for now, finding in most cases so far that things can be rewritten with generics rather than disabling the lint rule locally
  - I've disabled the [explicit-module-boundary-types](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md) rule to make this fix PR more manageable, but we might want to enable it later. Will bite us in a lot of places, so try to manually apply the rule when possible to make it easier if we eventually turn it on

To lint:
  - [x] FAQ/faq_data.ts (@ktw1016, #1160)
  - [x] InfoBase/AppState.ts (@ktw1016, #1160)
  - [x] about/about.tsx  (@Stephen-ONeil, #1152)
  - [x] charts/graphRegistry.ts (@ktw1016, #1160)
  - [x] components/Accordion/Accordions.tsx (@ktw1016, #1160)
  - [x] components/AlertBanner/AlertBanner.tsx (@csong202, #1156) 
  - [x] components/AutoHeightVirtualList.tsx (@ktw1016, #1160)
  - [x] components/BackToTop/BackToTop.tsx (@ktw1016, #1160)
  - [x] components/CheckBox/CheckBox.tsx (@ktw1016, #1160)
  - [x] components/DisplayTable/DisplayTable.tsx (@ktw1016, #1160)
  - [x] components/HeightClipper/HeightClipper.tsx (@ktw1016, #1160)
  - [x] components/LabeledBox/LabeledBox.tsx  (@Stephen-ONeil, #1152)
  - [x] components/LabeledTable/LabeledTable.tsx (@IvanFeng1 #1155)
  - [x] components/Select/Select.tsx (@IvanFeng1, #1154)
  - [x] components/TextMaker.tsx (@csong202, #1159) 
  - [x] components/TwoLevelSelect/TwoLevelSelect.tsx (@IvanFeng1 #1158)
  - [x] components/Typeahead/Typeahead.tsx (@csong202, #1163) 
  - [x] components/Typeahead/TypeaheadA11yStatus.tsx (@csong202, #1163)
  - [x] components/glossary_components.tsx (@IvanFeng1, #1153)
  - [x] components/modals_and_popovers/FocusLockedModal.tsx (~@sranc046, PR: TODO~ @Stephen-ONeil, trivial 12817a2)
  - [x] contact/contact.tsx (@Stephen-ONeil, #1152)
  - [x] core/assign_to_dev_helper_namespace.ts  (@Stephen-ONeil, trivial)
  - [x] core/format.ts (@Stephen-ONeil, PR: #1149)
  - [x] general_utils.ts (@ktw1016, #1160)
  - [x] types/global.d.ts  (@Stephen-ONeil, PR: #1164)
  - [x] types/types.d.ts  (@Stephen-ONeil, PR: #1164)
  - [x] types/util_types.d.ts  (@Stephen-ONeil, PR: #1164)

Note: I generated that list based on running eslint directly on `client/src`, _not_ based on the webpack output. Files like `types/global.d.ts` aren't directly part of our webpack'ed tree, so the build-time linter doesn't catch them. TODO: add a separate lint for them in CI or something? Hmm.

### DIFF
--- a/client/src/general_utils.ts
+++ b/client/src/general_utils.ts
@@ -39,7 +39,7 @@ export const text_abbrev = function (text: string, length: number) {
 };
 
 export const make_unique_func = function () {
-  var val = 0;
+  let val = 0;
   return function () {
     return ++val;
   };
@@ -83,14 +83,10 @@ export const escapeRegExp = function (str: string) {
   return str.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&");
 };
 
-type ObjectAny = {
-  [key: string]: any;
-};
-
-export const shallowEqualObjectsOverKeys = (
-  obj1: ObjectAny,
-  obj2: ObjectAny,
-  keys_to_compare: string[]
+export const shallowEqualObjectsOverKeys = <Type, Key extends keyof Type>(
+  obj1: Type,
+  obj2: Type,
+  keys_to_compare: Key[]
 ) =>
   _.reduce(
     keys_to_compare,
@@ -98,10 +94,10 @@ export const shallowEqualObjectsOverKeys = (
     true
   );
 
-export const shallowEqualObjectsExceptKeys = (
-  obj1: ObjectAny,
-  obj2: ObjectAny,
-  keys_to_ignore: string[]
+export const shallowEqualObjectsExceptKeys = <Type, Key extends keyof Type>(
+  obj1: Type,
+  obj2: Type,
+  keys_to_ignore: Key[]
 ) => {
   return _.isEqualWith(
     obj1,
@@ -110,11 +106,11 @@ export const shallowEqualObjectsExceptKeys = (
   );
 };
 
-export const retry_promise = (
-  promise_to_try: Function,
+export const retry_promise = <T>(
+  promise_to_try: () => Promise<T>,
   retries = 2,
   interval = 500
-) => {
+): Promise<T> => {
   return new Promise((resolve, reject) => {
     promise_to_try()
       .then(resolve)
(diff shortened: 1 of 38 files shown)


client/src/general_utils.ts
```diff
@@ -39,7 +39,7 @@ export const text_abbrev = function (text: string, length: number) {
 };
 
 export const make_unique_func = function () {
-  var val = 0;
+  let val = 0;
```
ktw1016, line 42: ha, didn't know eslint prefers let over var

ktw1016 (commented):
So other than the ones you mentioned, mainly, this is turning on the recommended rules (which is including non null assertions)

client/src/general_utils.ts
Stephen-ONeil: If you see a `var` you can't interpret it as a `let`, have to assume it's legacy code (pre-`let`/`const`) and might actually be either a constant _or_ a mutable.

Stephen-ONeil: Oh, but if you were wondering why it hadn't previously been applied, we disabled the rule in JS files because we wanted `var`s to be a sign that you were looking at legacy code that hadn't been refactored in years. I let the no-`var` rule apply to TS modules because we shouldn't be trying to type legacy code anyway, should be refactoring it first/in the process of typing it.

TS can give better type information on `const`s than on `let`s or `var`s too, I believe.

PR #1148, opened 2021-06-16
Stories to ts

IvanFeng1:


IvanFeng1:
https://github.com/TBS-EACPD/infobase/blob/f5d99ea901444ea7f64db1957477390ea5d02e53/client/src/components/PinnedContent/PinnedContent.tsx#L85-L89

I think I should use `!` to assert `local_storage_name` is not null instead of writing a if statement here if there wasn't going to be a Typescript rule in the linter that disallows using `!` for non-null-assertion. https://github.com/typescript-eslint/typescript-eslint/blob/v4.22.0/packages/eslint-plugin/docs/rules/no-non-null-assertion.md
Am I supposed to use `?` instead of `!` to assert non-null?

ktw1016 (approved):
Looks good. For non null assertions, as the link you put suggests, it `cancels the benefits of the strict null-checking mode` and is in the recommended typescript linter rule to disable it. So I think an `if` statement is good here

client/src/components/BackToTop/BackToTop.stories.tsx
```diff
@@ -1,18 +1,27 @@
+import { Story, Meta } from "@storybook/react";
 import React, { Fragment } from "react";
 
 import { BackToTop } from "./BackToTop";
 
-const BackToTopTemplate = ({ focusId }) => (
+interface TemplateProps {
+  focusId: string;
+}
+const BackToTopTemplate: Story<TemplateProps> = (args) => (
```
Stephen-ONeil, line 9: Not seeing the need to go from destructuring out `({focusId})` to just using `(args)` is. It makes the function signature much less useful (well, I guess we have the types now, but still doesn't help when scanning the code).

client/src/components/BackToTop/BackToTop.stories.tsx
```diff
@@ -1,18 +1,27 @@
+import { Story, Meta } from "@storybook/react";
 import React, { Fragment } from "react";
 
 import { BackToTop } from "./BackToTop";
 
-const BackToTopTemplate = ({ focusId }) => (
+interface TemplateProps {
+  focusId: string;
```
Stephen-ONeil, line 7: Based on the conditional on the input element, it looks like `focusId` was an optional arg in the JS version. Was that wrong?

client/src/components/BackToTop/BackToTop.stories.tsx
```diff
@@ -1,18 +1,27 @@
...
+}
+const BackToTopTemplate: Story<TemplateProps> = (args) => (
   <Fragment>
     <div id="ib-site-header-area">Header</div>
-    {focusId && <input id={focusId} placeholder="input for focus..." />}
+    {{ ...args } && (
+      <input id={args.focusId} placeholder="input for focus..." />
+    )}
```
Stephen-ONeil, line 14: Strange conditional, `{...args}` will always be truthy. Even if `args` itself were empty/falsey, destructuring anything in to an object will at least return an empty object. Empty objects are truthy.

Aside from that, by the interface, `focusId` is not optional. Unless it is supposed to be, this check isn't necessary.

client/src/components/BackToTop/BackToTop.stories.tsx
```diff
@@ -1,18 +1,27 @@
...
     <div id="wb-info" style={{ height: "300px", borderTop: "2px black solid" }}>
       Footer
     </div>
     <BackToTop
       focus={() => {
-        document.getElementById("#test").focus();
+        const nullCheck = document.getElementById("#test");
+        if (nullCheck) {
+          nullCheck.focus();
+        }
```
Stephen-ONeil, line 24: Couple of notes on naming:
  - by our conventions, in the docs, this should be `snake_case`
  - the name `nullCheck` is about _why_ you added the variable. Going to make the code a little confusing to future devs, they'll have to work backwards through your thought process for the name to click
    - generally, a name should answer _what_. What it contains if it's data, what it does if it's a method, etc.
    - in this case, the name should be descriptive of whatever the element that's expected to have the id `test` is. The fact that it's only stored in a variable for the sake of checking that it's non-null is apparent from the code itself

client/src/components/KeyConceptList/KeyConceptList.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Fusce quis risus
+    enim. Nulla a libero placerat, commodo purus at, rutrum nibh. Sed urna nunc,
+    eleifend ac hendrerit in, dictum sit amet nulla. Suspendisse quis egestas
+    ligula. Nunc mollis nibh ut ultricies ornare. Donec maximus feugiat eros, et
+    rhoncus risus. Pellentesque dignissim leo velit, vitae hendrerit diam
+    ultrices quis. Vivamus vulputate odio et orci vestibulum egestas.
+    Pellentesque accumsan neque in libero pharetra, et congue urna fermentum.
+    Aenean mattis sem ac erat interdum blandit.
+  </div>
+);
```
Stephen-ONeil, line 26: Noticed this in some other files too, looks like it might be a prop type mistake in the actual component. This _should_ accept strings. Looks like the prop type uses `React.ReactElement` where it should be `React.ReactNode`. Double check that in the other components where you had to make this sort of change. If they work with strings, then please fix the prop types (whether or not the story uses strings or JSX may be unimportant, although it's easier to edit strings in controls than JSX).

client/src/components/PinnedContent/PinnedContent.tsx
```diff
@@ -16,25 +18,49 @@ import text from "./PinnedContent.yaml";
...
-      ? JSON.parse(localStorage.getItem(local_storage_name))
-      : null;
+    if (has_local_storage) {
+      const storedString = localStorage.getItem(local_storage_name);
+      if (storedString !== null) {
+        return JSON.parse(storedString);
+      }
+    } else {
+      return null;
+    }
```
Stephen-ONeil, line 32: Has a case with no explicit return now (which would be `undefined`, instead of the expected `null | boolean`). I'd suggest changing lines 30-32 to
```
}
return null;
```
So that the first conditional branch falls through to returning null if it doesn't otherwise return.

Stephen-ONeil (changes requested):
Just thought to look it up, but I believe this works in general `type ComponentProps = React.ComponentProps<typeof Component>`. Might be preferable to adding prop type exports all over the place.

Didn't fully read through the `PinnedContent.js -> PinnedContent.tsx` changes, would prefer that be in its own PR.

client/src/components/BackToTop/BackToTop.stories.tsx
IvanFeng1: The input element in the JS version also depended on focusId not being null.

Stephen-ONeil: Right, but that might imply that it was intended to be optional, since the component knew to handle the case where it was falsey (by not showing the input). Was that wrong/unnecessary in the JS version?

IvanFeng1: I think it was intended to be optional there was a `BackToTop` story with the component that depended on `focusId` and one that did not. I made a mistake in defining `focusId` to be required and not optional in the interface.

client/src/components/KeyConceptList/KeyConceptList.tsx
```diff
@@ -5,7 +5,7 @@ import { Details } from "src/components/Details/Details";
 import "./KeyConceptList.scss";
 
 interface KeyConceptListProps {
-  question_answer_pairs: [React.ReactElement, React.ReactElement][];
+  question_answer_pairs: [React.ReactElement, React.ReactNode][];
```
Stephen-ONeil, line 8: Both the q and the a can be `ReactNode`s.

client/src/components/ShareButton/ShareButton.tsx
```diff
@@ -29,16 +35,32 @@ const CommonSocialMediaShareButton = ({ complete_url, media_icon }) => (
     {media_icon}
   </a>
 );
-export class ShareButton extends React.Component {
-  constructor(props) {
-    super();
 
+interface ShareButtonProps {
+  button_class_name: string;
+  button_description: string;
```
Stephen-ONeil, line 41: I think any prop with a default value should be marked optional, double check that.

Stephen-ONeil (changes requested):
Two small things, almost all good though!

Stephen-ONeil (approved):
:+1:

client/src/components/ShareButton/ShareButton.tsx
IvanFeng1: I noticed there is a few more components with props that have a default value. 
Should I open another PR for this issue? Sometimes I get this error in the file when I change props from required to optional: `Argument of type 'placeholdertype | undefined' is not assignable to parameter of type 'placeholdertype'.`, is it fine with I fix the error by doing something like `String(variable)` since we can't use `!` to assert non-null? Or should I use if statement instead to make sure the variable is not null?

Stephen-ONeil: Sounds like the React-ways of doing default prop values are somewhat incompatible with TS. Maybe move the default values in to the prop types or something so TS knows? Give it a google, must be a common problem, likely clear advice on how we should be doing this. If you find something, post it in slack (and yeah, either open an issue or PR to clean these cases up).

Either way, something like `String(variable)` is a bad hack, even if the default means it wouldn't happen\*, that'd just be converting `undefined` to `"undefined"`. Remember that TS doesn't exist at run-time, it's just a guide to help us write code that should have safer run-time behaviour. If you're writing something with potentially unwanted run-time behaviour just to make the type system happy, then it's a good sign that you're headed in the wrong direction.

\* wouldn't happen right now, but what if some code changes in the future or the default is dropped?

Stephen-ONeil: https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/default_props/

PR #1152, opened 2021-06-18
TS lint components with {} state

Stephen-ONeil:
Pretty trivial, opening a PR for further discussion though.

The `ban-types` rule doesn't using `{}` as a type (for good reasons it seems)... but `React.Component`'s default state and prop types are `{}` anyway.

Trivial fix for the lint rule is to just not explicitly set `{}` for stateless class components. Silences the lint error without actually changing this code. On the other hand, we could decide to explicitly provide `never` for the component state prop to make it explicit that the component is stateless. I prefer that, but wanted a conversation before we decide on it.

### DIFF
--- a/client/src/about/about.tsx
+++ b/client/src/about/about.tsx
@@ -46,7 +46,7 @@ const tech_icon_list = _.chain([
 interface AboutProps {
   toggleSurvey: () => void;
 }
-export default class About extends React.Component<AboutProps, {}> {
+export default class About extends React.Component<AboutProps, never> {
   render() {
     const { toggleSurvey } = this.props;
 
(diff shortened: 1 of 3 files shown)


ktw1016 (approved):
I don't mind either tbh. Does it actually matter? I tried googling to see how others do it but nobody seems to be talking about this topic

Stephen-ONeil:
With state of type `{}` you can make arbitrary `setState` calls without TypeScript saying anything. With type `never` TS will complain if you try to use `setState`. If `never` is our default, then devs will be reminded to properly add state types when adding state to a component that previously had no state.

PR #1155, opened 2021-06-21
ts lint LabeledTable

IvanFeng1:


Stephen-ONeil (changes requested):
Approved changes from bc1af2ed, but please rebase off `typescript-lint-rules` again and let me know when that's done. I'll double check then and then give final approval :+1:

Stephen-ONeil (approved):
:+1:

PR #1156, opened 2021-06-21
TS Lint Errors: AlertBanner.tsx

csong202:


client/src/components/AlertBanner/AlertBanner.tsx
```diff
@@ -8,7 +8,7 @@ interface AlertBannerProps {
   children: React.ReactNode;
   banner_class: string;
   additional_class_names: string;
-  style: Object;
+  style: Record<string, unknown>;
```
Stephen-ONeil: `React.CSSProperties` is the right type for React `style` objects (it's great, gives real detailed type checking for all styles).

Stephen-ONeil (changes requested):
See comment. In addition, please rebase off `typescript-lint-rules`.

Stephen-ONeil (approved):
:+1:

Issue #1083, opened 2021-05-17
Give consistent order to datasets page items across languages


Stephen-ONeil:
A user submitted a problem report about the metadata route, pointing out that items weren't in the same order between French and English. I'm sorting the items on that route by name, since there's no other clear ordering to pick. Should consider alternatives: do we want to assign a manual order, always order based off the English name, etc. Picking a new approach may not be easy, might need consultation within the team, but implementing any changes to that route should be fairly trivial :+1:

PR #1161, opened 2021-06-21
Consistent Order to data_sources' metadata

patfire12345:
The predetermined order for the metadata routes is:
(text_maker keys)
"IGOC"
"PA"
"ESTIMATES"
"CFMRS"
"RTP"
"DP"
"DRR"
"RPS"
"COVID"
"SERVICES"

Open to suggestions about how to sort the metadata!

### DIFF
--- a/client/src/metadata/data_sources.js
+++ b/client/src/metadata/data_sources.js
@@ -62,7 +62,7 @@ const infobase_open_data_page = {
 
 const sources = _.chain([
   {
-    key: "1 IGOC",
+    key: "IGOC",
     title: text_maker("igoc_source_title"),
     frequency: text_maker("yearly"),
     open_data: infobase_open_data_page,
@@ -76,7 +76,7 @@ const sources = _.chain([
     ],
   },
   {
-    key: "2 PA",
+    key: "PA",
     title: text_maker("pa_title"),
     frequency: text_maker("yearly"),
     open_data: infobase_open_data_page,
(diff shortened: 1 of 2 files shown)


Stephen-ONeil (changes requested):
Did you accidentally bring in the `FixedPopover` commit from another branch? Please drop it.

Some thoughts:
  - I gave you the list of keys just as a convenient way to share the intended order, didn't mean to imply that sorting needed to happen via the keys, if that's what occurred
  - using hardcoded alphanumeric strings for sort indexes is dicey, easy to make a mistake with (especially if you get past 27 items)
  - could add a numeric property for order and sort by that instead, better than encoding order in to the key property. Deriving information from other properties is fine, but specifically storing multiple unrelated information in the same field is a bad pattern. Meaning/rules of the property become unclear, using it might require special parsing (not in this case, but in general), and both of those thing create lots of room for bugs
  - _in either case_, using the `key` string or a numeric `order` property, inserting a new item in to the metadata page would be a little annoying and possible to mess up. Would have to find and update the sorting value for all items intended to appear after the newly inserted item
  
Easy, better, alternative though. The hardcoded `sources` array itself provides an order. Just rearrange the items in it to match the desired order and drop the sorting. Reordering/adding new items in the future will be made easier by this too!

Stephen-ONeil (approved):
:+1:

Stephen-ONeil:
Closes #1083

Reminder: if you write "closes <issue number>" in the PR description, GitHub will auto close the corresponding issue when the branch merges.

PR #1163, opened 2021-06-22
TS Lint Errors: Typeahead.tsx and TypeaheadA11yStatus.tsx

csong202:
- deleted min_length props from TypeaheadA11yStatus because it was unused

### DIFF
--- a/client/src/components/Typeahead/Typeahead.tsx
+++ b/client/src/components/Typeahead/Typeahead.tsx
@@ -72,6 +72,7 @@ export class Typeahead extends React.Component<TypeaheadProps, TypeaheadState> {
 
   static defaultProps = {
     placeholder: text_maker("search"),
+    min_length: 3,
     on_query_debounce_time: 300,
   };
 
--- a/client/src/components/Typeahead/Typeahead.tsx
+++ b/client/src/components/Typeahead/Typeahead.tsx
@@ -123,7 +123,7 @@ export class Typeahead extends React.Component<TypeaheadProps, TypeaheadState> {
       return null;
     }
   }
-  componentDidUpdate(prevProps: TypeaheadProps, prevState: TypeaheadState) {
+  componentDidUpdate(_prevProps: TypeaheadProps, prevState: TypeaheadState) {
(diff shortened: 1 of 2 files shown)


client/src/components/Typeahead/Typeahead.tsx
```diff
@@ -54,19 +64,14 @@ interface TypeaheadState {
...
+export class Typeahead extends React.Component<TypeaheadProps, TypeaheadState> {
   menu_id: string;
 
   private typeahead_ref = React.createRef<HTMLDivElement>();
-  private virtualized_list_ref = React.createRef<any>();
+  private virtualized_list_ref = React.createRef<List>();
 
   static defaultProps = {
     placeholder: text_maker("search"),
-    min_length: 3,
```
Stephen-ONeil, line 69: I think this default's still valid. Let me know if I'm wrong about that!

Stephen-ONeil (changes requested):
Thanks for catching the unused `min_length` prop in `TypeaheadA11yStatus`. I think it was moved from there to the screen reader only  `typeahead_usage` text directly in `Typeahead`.

client/src/components/Typeahead/Typeahead.tsx
csong202: oops didn't mean to delete that one

client/src/components/Typeahead/Typeahead.tsx
```diff
@@ -54,19 +64,14 @@ interface TypeaheadState {
...
-  TypeaheadProps,
-  TypeaheadState,
-  {}
-> {
+export class Typeahead extends React.Component<TypeaheadProps, TypeaheadState> {
   menu_id: string;
 
   private typeahead_ref = React.createRef<HTMLDivElement>();
-  private virtualized_list_ref = React.createRef<any>();
+  private virtualized_list_ref = React.createRef<List>();
```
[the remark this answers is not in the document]
csong202: It hasn't caused a problem here since List has recomputeRowHeights

ktw1016 (approved):
Looks good to me

PR #1164, opened 2021-06-22
TS lint central types

Stephen-ONeil:


client/src/types/types.d.ts
```diff
@@ -1,5 +1,12 @@
 interface TextBundle {
-  [key: string]: Object;
+  [key: string]: {
+    transform?:
+      | ["handlebars"]
+      | ["markdown"]
+      | ["handlebars", "markdown"]
+      | ["handlebars", "embeded-markdown", "markdown"]; // TODO embeded-markdown transform is legacy, hacky. Remove at some point
+  };
+  [(key in "en") | "fr" | "text"]: string;
```
ktw1016, line 9: Really minor but is the brackets supposed to be `key in ("en" | "fr" | "text")`?

ktw1016: Also just a bit confused, shouldn't `transform` and `(key in "en") | "fr" | "text"]` be in the same scope?
So that it should be 
```
[key: string]: {
    transform?: ...
    key in ("en" | "fr" | "text"): string
```

ktw1016 (approved):

client/src/types/types.d.ts
Stephen-ONeil: > Really minor but is the brackets supposed to be `key in ("en" | "fr" | "text")`?

Ugh, prettier changed `[key in "en" | "fr" | "text"]: string;` to `[(key in "en") | "fr" | "text"]: string;`, annoying.

Leaves `[key in ("en" | "fr" | "text")]: string;` alone at least.

Stephen-ONeil: And yup, they were supposed to be in the same scope, must have typoed it.

ktw1016: 👍

PR #1166, opened 2021-06-23
Nivo upgrade

Stephen-ONeil:
Otherwise, as always (god damn it), tons of Nivo "minor sem ver update" breaking changes.
  - [x] oh, upgrade `@nivo/treemap`, missed it
  - [x] `Bubble` charts renamed to `CirclePacking`
  - [x] all line tootltips not appearing
    - https://github.com/plouc/nivo/issues/1557
      - the issue poster didn't specify if this was about slice tooltips or not. Either way, I edited the sandbox the maintainer linked to use slices and it _does_ work there
  - [x] Government of Canada Spending Areas tooltips appearing with incorrect styles
  - [x] pie chart tooltips not appearing
  - [x] pie chart labels reverted to unwanted defaults
  - [x] `CirclePacking` throws an error on AGR finance route
  - [x] our patch package fix for the treemap may not be necessary any more, or at least will require require tweaks
    - ... whoever wrote this didn't leave a comment in the patch file for what was being fixed, ouch. Dig up the source PR
    - looks like it's managing to apply the patch, just with a warning, for now. Try removing the patch and reinstalling, tweak props, to see if the patch isn't needed. Otherwise, generate a new one
  - [x] TODO keep checking the site, sure to be more


ktw1016:
I think there might be a bug with circle packing in Nivo? Or its intended behaviours changed...

So before update, we have 1 parent, 1 child values which shows 2 circles
But after update, having 1 child seem like it always fill up the entire circle (2 circles of same size, effectively showing only 1 circle. You can inspect and change around with `r` value to see this)

As a test, I tried with value like:
```
    const graph_data = {
      id: "parent",
      name: "parent",
      color: color_scale("parent"),
      children: [
        {
          id: "child",
          name: "child",
          value: 50,
          color: color_scale("child"),
        },
        {
          id: "child2",
          name: "child2",
          value: 1,
          color: color_scale("child2"),
        },
      ],
    };
```
which will show 3 circles.. but couldn't get it to work as the same way

ktw1016:
Hmm other than the circle packing issue, everything else seem to be working. Anyone is welcome to stab at fixing circle packing..

Stephen-ONeil:
Looks like the previous nivo bubble component had a prop for `nodeComponent`, which we were relying on to heavily change how the graph rendered. Expected this to totally break some day, going by this comment "Hacky abuse of ResponsiveCirclePacking... very fragile against nivo changes", ha.

The comment also says "Would be trivial to make this graph ourselves, only reason for doing it like this is to get nivo native tooltips" which I think is true and probably what we need to do now. I'll do that soon.

Stephen-ONeil:
Related #1175

Issue #939, opened 2021-03-08
Revisit storybook for EmailFrontend


GarvinH:
This is seriously a pain of a component to try to add into the storybook haha...

Annoying issue with postcss loader. Took a while to get it to work -> will link the PR to see how it was done.

Mocking documentation on storybook doesn't seem to work? I tried what they said to do + other resources but none of them managed to get mocking to work...

GarvinH:
related #921 

If merged, then the postcss loader has been added. Refer to #940 if there are still any issues on postcss loader (as `postcss.config.js` was not used in #921

GarvinH:
related #921 

Paginated displaytable managed to not have any issues using `SpinnerWrapper` so it might be possible the svg issue no longer is present

PR #1168, opened 2021-06-25
Storybooking EmailFrontend.js

patfire12345:
https://github.com/TBS-EACPD/infobase/issues/939

Currently, the submit button is disabled and won't work as a result since only mock calls are being made.

Not sure what controls should be implemented if any. 

Awaiting review just for general feedback about further implementation.


client/src/components/EmailFrontend/EmailFrontend.stories.js
```diff
@@ -0,0 +1,85 @@
...
+      form_label: { en: "Details (required)", fr: "Détails (obligatoire)" },
+    },
+    sha: { required: true, value_type: "string", form_type: false },
+    route: { required: true, value_type: "string", form_type: false },
+    lang: { required: true, value_type: "string", form_type: false },
+    app_version: { required: true, value_type: "string", form_type: false },
+    client_id: { required: true, value_type: "string", form_type: false },
+    additional: { required: false, value_type: "json", form_type: false },
+  };
+  fetchMock.mock("*", template);
```
Stephen-ONeil, line 73: Look out, `fetchMock` is global/persistent!

Should use a more specific than the wild card. Skimming the `fetchMock` docs, might also want to either use `.getOnce`, clean up all mocks during unmount, or both.

Stephen-ONeil (changes requested):
If we're not going to refactor the component to split behaviour from presentation\*, then fetch-mock seems like the right tool for the job :+1: 

\* makes testing/storybooking either but the point of this component _is_ to encapsulate the email backend interface

Generally look through the `fetch-mock` docs carefully before continuing. I think it can do some other handy stuff that'd improve this story (e.g. adding a delay, to simulate a realistic use case).

Suggestions/TODOs
  - [ ] mocking the specific URLs/methods instead of a wildcard just for the initial GET will let you mock the submission POST as well
    - actually, relying on specific URLs would be brittle to changes on the server side and in `email_backend_utils`. Using `getOnce` and `postOnce` should be a pretty safe bet, as long as they're cleared on unmount
  - [x] should probably hand craft a specific template that covers all the cases `get_form_for_user_field` can construct. The report a problem form doesn't use enums, and doesn't cover all combinations of required forms/"connected" text areas
    - this stuff is mostly undocumented. If you can't identify them all from reading the code, let me know and I can help
  - [ ] a separate story where the mocked POST always returns an error response would be useful to demonstrate how those are handled/displayed by the component
  - [x] controls:
    - [x] they aren't working right now?
    - [ ] maybe make `template` an editable control?

client/src/components/EmailFrontend/EmailFrontend.js
```diff
@@ -162,20 +164,22 @@ class EmailFrontend extends React.Component {
 
     const { include_privacy } = this.props;
 
-    const user_fields = _.omitBy(
-      template,
-      ({ form_type }, key) => key === "meta" || !form_type
-    );
+    const user_fields = _.omitBy(template, ({ form_type }, key) => {
+      key === "meta" || !form_type;
+    });
```
Stephen-ONeil, line 169: Some how this got mangled. Look closely, the applied function went from a one-liner to having a body (but no return function). Nothing was being omitted as a result, and the future logic relying on `user_fields` was broken. That's why the submit button wasn't working as expected.

client/src/components/EmailFrontend/EmailFrontend.js
```diff
@@ -162,20 +164,22 @@ class EmailFrontend extends React.Component {
...
-      .keys()
-      .every(
-        (required_field_key) =>
-          !_.isUndefined(completed_template[required_field_key]) &&
-          !_.isEmpty(completed_template[required_field_key])
-      )
-      .value();
+    const all_required_user_fields_are_filled =
+      // specifically for storybook, will not return true
+      this.props.all_required_user_fields_are_filled ||
```
Stephen-ONeil, line 173: Delete, not necessary if above is fixed. Besides which, if a component really needs to be modified for the sake of storybook testing, then the component should probably be refactored in general but, I think, it should never be left containing specific artifacts related to the story. The stories need to make the components easier to maintain, but if they add weird special props to the interface then they're probably only making things worse (absolutely worst case, using this prop in the actual client would need to some how throw an error, based of environment variables or something),

client/src/components/EmailFrontend/EmailFrontend.stories.js
```diff
@@ -0,0 +1,686 @@
...
+    loading: {
+      defaultValue: true,
+    },
+    include_privacy: {
+      defaultValue: true,
+    },
+    // for some reason, this is needed to activate the submit button, will always be disabled otherwise
+    // all_required_user_fields_are_filled: {
+    //   defaultValue: true,
+    // },
```
Stephen-ONeil, line 21: Drop this.

Stephen-ONeil (changes requested):
Bunch of scattered things. Believe I spotted the hidden bug that was messing with the submit button behaviour. At least one of these is a repeat suggestion from the last review, so please double check that you got all of the previous comments too.

client/src/components/EmailFrontend/EmailFrontend.js
patfire12345: Ohh I see now, thanks for pointing that out!

client/src/components/EmailFrontend/EmailFrontend.stories.js
patfire12345: Done!

client/src/components/EmailFrontend/EmailFrontend.js
patfire12345: Done! Thanks for the advice.

client/src/components/EmailFrontend/EmailFrontend.js
```diff
@@ -350,6 +354,12 @@ class EmailFrontend extends React.Component {
                   disabled={!ready_to_send}
                   onClick={(event) => {
                     event.preventDefault();
+                    console.log("All Required User Fields Are Filled");
+                    console.log(all_required_user_fields_are_filled);
+                    console.log("All Connected User Fields Are Filled");
+                    console.log(all_connected_user_fields_are_filled);
+                    console.log("Privacy Acknowledged");
+                    console.log(privacy_acknowledged);
```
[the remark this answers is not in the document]
patfire12345: Yup! Got it.

package.json
```diff
@@ -25,5 +25,8 @@
     "eslint-plugin-node": "^11.0.0",
     "husky": "^6.0.0",
     "prettier": "^2.3.0"
+  },
+  "dependencies": {
+    "fetch-mock": "^9.11.0"
```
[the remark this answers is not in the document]
patfire12345: Done!

Stephen-ONeil:
Unassigning myself before vacation, up for grabs. 

TODO:
  - [ ] general problem in storybooks (and will be a similar problem in other test code), calls to analytic's `log_standard_event` need to be mocked out
    - any calls throw an error about analytics not being initialized, but calling `initialize_analytics` from `.storybook/preview.js` isn't the easy-fix it sounds like, because _that_ relies on a snippet of JS that's hardcoded in to the index html files
  - [ ] for this specific story, once that's done, the only thing left is confirming the submit behaviour, since I could test that fully with analytics throwing errors

PR #1169, opened 2021-06-28
TS subject entities

ktw1016:


client/src/models/populate_stores.js
```diff
@@ -260,7 +260,6 @@ function populate_programs(rows) {
       crso,
       activity_code: row[activity_code],
       dept: Dept.lookup(row[dept_code]),
-      data: {},
```
ktw1016, line 263: I couldn't find this data actually being populated / used but it might've been just that I couldn't find it. If anyone knows we actually need this, let me know

Stephen-ONeil (commented):
Lots to think about here, might be easier to refactor certain patterns, like the mixins, than to type this properly as-is. Let's get on a call and talk about it some time tomorrow.

client/src/models/storeMixins.ts
```diff
@@ -0,0 +1,196 @@
...
+export const getStaticStore: () => typeof StaticStore = () => {
+  const _storeMap = new Map<string, StaticStore>();
+  StaticStore.register = (id: string, instance: StaticStore) => {
+    _storeMap.set(id, instance);
+  };
+  StaticStore.lookup = (id: string) => _storeMap.get(id);
+  StaticStore.get_all = () => _.uniq(Array.from(_storeMap.values()));
+  StaticStore.__store__ = _storeMap;
+  return StaticStore;
+};
```
ktw1016, line 23: up until here is the only actual change. Below hasn't changed.

There is one major problem tho. `_storeMap` is storing all types of these static stores. So by the end, `_storeMap` has tables, glossary entries, footnotes, etc in that one single instance of variable.

Theoretically, I don't think this should be happening because `_storeMap` is being initialized as a new Map object every time this function is called. @Stephen-ONeil do you know why?

Trying different things on mixins to force it to work made me wonder the future of store mixins (and basically all of our models, subjects, etc) when we fully drop Table API and fully transition to server side. Will we keep the client data structures as is, but only swap out how it's coming in? Or refactor it to more appropriately suit the new data flow?

Stephen-ONeil: Initializing a new map, but not a new `StaticStore`. Each call to `getStaticStore` mutates the static properties on the `StaticStore` class itself, clobbering over the map referenced by `__store__` each time. Everything extending from it is sharing this one static `__store__` (so there's a second bug here, which is that timing on calls to `getStaticStore` might lead to _wiping_ any stores that have been populated already).

I'm adopting the branch, so I'll sort that out :+1:

Stephen-ONeil:
Reminder to my self, for when I get back to this, new TS 4.4 [static blocks](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#static-blocks) could be useful here!

Stephen-ONeil:
Stopping here. More than enough changes on this branch.

Future steps:
  - [ ] refactor and port`Result` and `Indicator` "subjects"
  - [ ] after that I think some of the modules/directories under `models/` could do with a re-organization

Stephen-ONeil:
- [ ] IGOC "Show organizations without data" feature not working as expected

Issue #1070, opened 2021-05-07
bug with Details on Statutory Items in Estimates and Details on Voted Items in Estimates 


AlexCLeduc:
http://dev.ea-ad.ca/master/index-eng.html#orgs/gov/gov/infograph/financial/.-.-(panel_key.-.-'in_year_stat_breakdown)

![Screen Shot 2021-05-07 at 11 20 55 AM](https://user-images.githubusercontent.com/7769215/117471875-4c615f80-af26-11eb-87df-7cbab04d7b26.png)

Stephen-ONeil:
Load the finance infographic, switch to another infographic (e.g. people), then switch back to finances. Do that a couple times, it starts appending additional copies of the dept name on each re-load of the panel. Must be some junk in there mutating the underlying data, ugh.

PR #1170, opened 2021-06-28
Fixed bug with Details on Statutory Items in Estimates and Details on Voted Items in Estimates

csong202:
Issue #1070

- adjusted text_func so department name only gets appended to desc once
- added _.clone to packing_data.children so _.extend won't change data
- deleted an unused WrappedNivoTreemap props

### DIFF
--- a/client/src/panels/panel_declarations/finances/vote_stat/in_year_vote_stat_breakdown.js
+++ b/client/src/panels/panel_declarations/finances/vote_stat/in_year_vote_stat_breakdown.js
@@ -30,11 +30,15 @@ const { Dept } = Subject;
 const text_func = (vs, d, break_str) => {
   if (vs == "voted") {
     return d.dept
-      ? `${Dept.lookup(d.dept).name} ${break_str}  ${d.desc}`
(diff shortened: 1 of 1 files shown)


client/src/panels/panel_declarations/finances/vote_stat/in_year_vote_stat_breakdown.js
```diff
@@ -106,7 +110,7 @@ const planned_vote_or_stat_render = (vs) =>
       name: "root",
       color: "white",
       children: _.map(data, (d, i) =>
-        _.extend(d, {
+        _.extend(_.clone(d), {
           id: i,
```
Stephen-ONeil, line 114: More directly, could be 
```
_.map(data, (d, i) => ({
  ...d,
  id: i,
  // etc.
```

Intent's a bit clearer and avoids all the lodash util nesting (otherwise, would suggest chaining instead, at least within the `map`).

Stephen-ONeil (approved):
Approved with one pretty minor comment, good to merge once addressed!

Issue #1151, opened 2021-06-18
User feedback: RPB column toggle behaviour


Stephen-ONeil:
When using the `DisplayTable` "Select columns" feature to toggle off an index column, the rows are not re-grouped by the remaining columns. For example, the [FTE by program table](http://dev.ea-ad.ca/master/index-eng.html#rpb/.-.-(subject.-.-'gov_gov.-.-columns.-.-false)) does _not_ become an FTE by org table just by [toggling off the program](http://dev.ea-ad.ca/master/index-eng.html#rpb/.-.-(table.-.-'programFtes.-.-subject.-.-'gov_gov.-.-columns.-.-(.-.-'*7b*7bpa_last_year_5*7d*7d.-.-'*7b*7bpa_last_year_4*7d*7d.-.-'*7b*7bpa_last_year_3*7d*7d.-.-'*7b*7bpa_last_year_2*7d*7d.-.-'*7b*7bpa_last_year*7d*7d.-.-'*7b*7bplanning_year_1*7d*7d.-.-'*7b*7bplanning_year_2*7d*7d.-.-'*7b*7bplanning_year_3*7d*7d))) column.

Regrouping wasn't ever the intended behaviour of the "Select columns" menu. At least one user assumed it would be though, and it does sound like a useful feature to have ... I don't think it's feasible to add to the default "Select columns" though, can't regroup arbitrary table column contents (some columns will be explicitly non-summable). Might just be a special, separate, control for tables (with only summable columns) to opt in to?

Lots of open questions here, please share your thoughts. I didn't refresh my self on the `DisplayTable` API/features before opening this, and I think there's a lot of open room to design any number of "solutions" to this.

csong202:
I think this feature would only make sense for tables with a "Program" column. It wouldn't really make sense on Employee Type or Geographic Region

Stephen-ONeil:
Hm, it might! Using the employee tables for instance, rolling up other rows when the dimension is disabled would give a table of per-dept employee totals while disabling the dept column who give government totals broken down by the remaining dimension (e.g. Employee Type).

Can't say it would always make sense for every dataset though, and the value of the feature would vary a lot. For example, all people tables sum to the same total if you remove the dimension, so lots of redundancy there. So it might not be a good default behaviour for all tables, even just within the RPB. I think the question is whether it's worth the effort/added ongoing complexity for the sake of the cases where it seems to make sense.

Issue #889, opened 2021-02-11
RPB Group By dropdown missuses table dimensions


Stephen-ONeil:
It functions something like a filter menu right now (a clunky one, because it's a radio menu), where you pick what index column value to show. That's not the intended use of the table `dimensions` though, they're supposed to regroup all rows, and potentially rename the index values.

E.g. we have these options, as a radio, in the Ex level report
![image](https://user-images.githubusercontent.com/11301919/107687601-ad093500-6c74-11eb-858e-7dabbb305b94.png)

Should be something like this instead
![image](https://user-images.githubusercontent.com/11301919/107688138-4c2e2c80-6c75-11eb-8b43-a74af61a31b0.png)

These are fairly legacy though, and not likely to survive the inevitable switch from Tables to GraphQL in the RPB. Could choose to either fix this or drop the Group By menu altogether.

Stephen-ONeil:
Try to address the idea behind #1151 in the process.

PR #1173, opened 2021-06-30
RPB Column Toggle Row Grouping Idea

csong202:
Issue #1151

### DIFF
--- a/client/src/components/DisplayTable/DisplayTable.tsx
+++ b/client/src/components/DisplayTable/DisplayTable.tsx
@@ -389,13 +389,14 @@ export class _DisplayTable extends React.Component<
                 active: _.includes(visible_ordered_col_keys, key),
               }))}
               onClick={(clicked_key: string) => {
+                const toggled_columns = toggle_list(
+                  visible_col_keys,
+                  clicked_key
+                );
                 col_configs_with_defaults[clicked_key].visibility_toggleable &&
                   this.setState({
-                    visible_col_keys: toggle_list(
-                      visible_col_keys,
-                      clicked_key
-                    ),
-                    row_grouping: _.includes(visible_col_keys, "prgm"),
+                    visible_col_keys: toggled_columns,
+                    row_grouping: !_.includes(toggled_columns, "prgm"),
                   });
               }}
             />
@@ -431,17 +432,18 @@ export class _DisplayTable extends React.Component<
     const row_grouped_data = _.chain(data)
       .groupBy("dept")
       .map((dept_data) => {
+        const summable_columns = _.keys(
+          _.omit(dept_data[0], ["dept", "prgm", "legal_title"])
+        );
+        const column_sums = summable_columns.map((col) => {
+          return _.sumBy(dept_data, col);
+        });
+        console.log(column_sums);
+        const columns = _.zipObject(summable_columns, column_sums);
         return {
           dept: dept_data[0].dept,
           legal_title: dept_data[0].legal_title,
-          "{{pa_last_year_2}}": _.sumBy(dept_data, "{{pa_last_year_2}}"),
-          "{{pa_last_year_3}}": _.sumBy(dept_data, "{{pa_last_year_3}}"),
-          "{{pa_last_year_4}}": _.sumBy(dept_data, "{{pa_last_year_4}}"),
-          "{{pa_last_year_5}}": _.sumBy(dept_data, "{{pa_last_year_5}}"),
-          "{{pa_last_year}}": _.sumBy(dept_data, "{{pa_last_year}}"),
-          "{{planning_year_1}}": _.sumBy(dept_data, "{{planning_year_1}}"),
-          "{{planning_year_2}}": _.sumBy(dept_data, "{{planning_year_2}}"),
-          "{{planning_year_3}}": _.sumBy(dept_data, "{{planning_year_3}}"),
+          ...columns,
         };
       })
       .value();
--- a/client/src/components/DisplayTable/DisplayTable.tsx
+++ b/client/src/components/DisplayTable/DisplayTable.tsx
@@ -396,7 +396,7 @@ export class _DisplayTable extends React.Component<
                 col_configs_with_defaults[clicked_key].visibility_toggleable &&
                   this.setState({
                     visible_col_keys: toggled_columns,
-                    row_grouping: !_.includes(toggled_columns, "prgm"),
+                    row_grouping: !_.includes(toggled_columns, trigger_column),
                   });
               }}
             />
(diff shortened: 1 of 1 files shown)


csong202:
Not really sure what to do for tables that have multiple unsummable columns other than the basic columns. For example, Program Expenditures by Authority Type. If program is toggled off should the table rows be grouped based on organization and voted/stat type?

client/src/components/DisplayTable/DisplayTable.tsx
```diff
@@ -421,6 +429,43 @@ export class _DisplayTable extends React.Component<
...
+      .map((dept_data) => {
+        const basic_columns = ["dept", "prgm", "legal_title", trigger_column];
+        const summable_columns = _.chain(dept_data[0])
+          .omit(basic_columns)
+          .keys()
+          .value();
+        const column_sums = summable_columns.map((col) => {
+          return _.sumBy(dept_data, col);
+        });
+        const columns = _.zipObject(summable_columns, column_sums);
```
ktw1016, line 456: Above two could be written as
```
_.chain(summable_columns)
          .map((col) => [col, _.sumBy(dept_data, col)])
          .fromPairs()
          .value();
```
Or you could even put `summable_columns` into the above chain

client/src/components/DisplayTable/DisplayTable.tsx
```diff
@@ -274,51 +277,53 @@ export class _DisplayTable extends React.Component<
...
-            ? _.map(unsorted_array).sort(
-                (a: DisplayTableData, b: DisplayTableData) =>
-                  sorting_config.sort_func(a[sort_by], b[sort_by], descending)
-              )
-            : _.sortBy(unsorted_array, (row: DisplayTableData) =>
-                is_number_string_date(
-                  sorting_config.raw_formatter(row[sort_by])
-                ) /*Please Leave On: sorting data containing BOTH string and date is not consistent
+
+    const sorted_filtered_data = (data: DisplayTableData[]) =>
```
ktw1016, line 281: - this should be renamed as a _function_ now like `sort_and_filter_data`
- But before doing that, I don't think this needs to be a function. Can you just reorder and group the data before getting `total_row`, `csv_string`, etc, so then group the data first, sort and filter the data, then do everything else? Haven't looked at it thorough enough to see if that's possible so lemme know

ktw1016 (changes requested):
So this whole thing only applies to report builder right? so I wonder if you could take this logic out of DisplayTable and put it separately within report builder? Also rolling it up at dept level works, but not the other way around. For example, toggling off dept and only having `so`, or toggling off `prgm` doesn't roll data up at those level..

Tbh, I think this whole approach needs to be discussed and @Stephen-ONeil should take a look also. I don't have high confidence in this one

client/src/components/DisplayTable/DisplayTable.tsx
csong202: Yep, it can be reordered

Stephen-ONeil (commented):
Had a discussion about this with Taewan just now, ended up agreeing that what the user wanted could be provided as part of #889 and wouldn't be useful enough as a general `DisplayTable` feature. Sorry about the indecisiveness here!

Taewan went back and read the original user report again. It seems like the user only went to the column filter option, hoping it would roll up, _after_ they realized that the group by button wasn't doing what they wanted. Seems addressing #889 in the first place would most likely have covered this off in practice.

Otherwise, my main point in favour of this feature existing within `DisplayTable` was that toggling off dimensions is currently useless in almost all cases. Rolling up, _when possible_, would provide more utility. I had some vague thoughts on how that could be implemented relatively cleanly, but hadn't explored them too deeply (otherwise I'd have opened a PR for the issue myself, haha).

The cons are mostly from that "when possible" clause:
  - it doesn't apply if there's only one dimension column
    - often the case outside the RPB where tables are generally simpler and smaller
    - on that note, if the table's only a handful of lines, the practical utility is also not very much
  - requires special behaviour in tables with non-summing data
    - do nothing if any visible columns can't sum? Replace non-summing cell values with N/As?
  - requires the assumption that summing by any one dimension is correct
    - counter example, which unfortunately only just occurred to me: programs across departments can share the same name, but it would generally not make sense to roll them together
    - ... further related gotcha, the raw data value for a "program" column could be the program id in some cases, with the formatter being what maps it to the plain name, could be the name in others. Just some extra "hidden logic" to this that can create inconsistencies for users
    - _or_ instead we only allow dimensions to roll up from right to left, assuming that dimensions columns are in order of granularity ?
    
Regardless, it looks like a) it wouldn't apply to most tables outside of the RPB anyway, and b) there's more special cases to think through than originally anticipated. Ouch! Double whammy, lots of new code for marginal additional utility (to top it off `DisplayTable`'s already rather bloated).

I'll leave some general advice on the direction your implementation was going in as follow-up comment. That way some good can come out of us waffling on this one :+1:

Stephen-ONeil:
The main change in direction I would have requested, if this branch was continuing, would be to not make the special behaviour rely on a hard coded list of column keys, i.e. `cols_with_row_grouping`. I can think of at least three big ways this could go wrong in the future. I'll provide one example, try to think of at least one other on your own.
  - False negatives. What if a new `Table` instance is created and has columns which _should_ share this behaviour, but happen to use different column keys?
    - added level of gotcha, how long before anyone notices this has happened? Is it even obviously a bug at first glance?
    
It's taken me a couple tries to write this feedback, sorry about the delay. I found I could classify this under a couple different common software design anti-pattern labels. I keep having to fight the impulse to try and turn this in to a very broad discussion about software design/architecture itself, and some of the "big ideas" I have about it, haha. Had to cut my self off from that or this would never get finished. I'll try and give a condensed dev-talk for everyone on the topic soonish.

The basic design issue is about information leaks, violating the principles of [information hiding](https://en.wikipedia.org/wiki/Information_hiding). These keys come from `Table` class definitions where they have no inherent meaning, they're just internal identifiers for the table's columns. Using them outside the context of any specific table (and adding new special meaning to them in the process) is sort of a two-way leak. Internals from a handful of table instances have leaked in to `DisplayTable`, and special case features from `DisplayTable` have leaked in to the inherent meaning of a table's column keys... _if we're lucky_.

The second-order issue is the _implicit_ [coupling](https://en.wikipedia.org/wiki/Coupling_(computer_programming)) this creates between `DisplayTable` and the table definitions. The linked wiki article does a good job at providing motivation for why low coupling is preferred in general. My focus here is on the bonus risks caused by that coupling being implicit. It's why I said we'd be _lucky_ if the two modules manage to leak meaning in to each other. If the leak is happening then that means developers are aware of and maintaining the coupling, and that this feature probably still works as intended. Being implicit, this relies entirely on the future developers (including me, I'd forget eventually) being the ones to know this coupling exists, since the code itself doesn't "know" anything about it!

Generally, a lot of these sorts of software design issues are fixed at the API level. The `DisplayTable` already has a (fairly heavy) declarative API. The agreement of this API is that the consumer describes "what" its data is and the `DisplayTable` decides "how" that data is displayed. Pretty flexible, and allows for a certain amount of "magic" in what `DisplayTable` actually does... but the agreement is broken by this implementation! Via `cols_with_row_grouping`, `DisplayTable` now makes decisions based off of its own assumptions, rather than what was actually declared.

My concrete advice would have been to extend `DisplayTable`'s column config API with something like `is_dimension`. In the case of the RPB/the table classes, this could be set based off where the table's column definition has `key === true` (... don't get me started on the `Table` class, it's got poor naming; "key" means "dimension column" essentially). Otherwise, it'd be an opt-in declaration for tables that want to automatically roll-up. I believe this would address every gotcha and design issue I've pointed out. For instance, instead of an (implicit) two way coupling between the table definitions and the `DisplayTable` itself, there's just a one-way flow of table configuration data through the RPB and in to the `DisplayTable`.

Of course, earlier comment stands, there's other flaws with this behaviour and reasons we shouldn't implement it :( 

Let me know if there's anything here you want me to clarify!

PR #1174, opened 2021-06-30
Drop react FC type

Stephen-ONeil:
Following from https://github.com/facebook/create-react-app/pull/8177

Lots of good reasons to not use the FC/FunctionComponent type turns out, seems better to just use normal functions.

Semi-related, cleaned up a couple props in Typeahead.tsx that were using `React.Component` where `React.ReactNode` was the appropriate type.

Possible TODO:
  - [x] maybe add `React.FC` etc to the linter's banned types, or look for an existing plugin with that rule


ktw1016 (approved):
Seems pretty trivial. Nice catch!👍

PR #1178, opened 2021-07-07
TS: HeaderNotification and HeaderNotification.stories

csong202:
- converted HeaderNotification.js and HeaderNotification.stories.js to ts


ktw1016 (approved):
👍

PR #1180, opened 2021-07-07
Drop graph registry

ktw1016:
Not a refactor, just copied over what it was doing

I can also work to convert canadad3component and canada.js to Typescript


Stephen-ONeil (approved):
Good incremental step to me, can merge as is and leave cleaning up/refactoring for when it's ported to TS. If you don't plan to do that next, maybe leave a comment in the file reminding that it should be refactored before it's ported, due to being mostly legacy code.

PR #1183, opened 2021-07-08
Need feedback on creating mongoose schema from csv files

IvanFeng1:


server/src/models/people/populate.js
```diff
@@ -30,17 +30,37 @@ const format_csv_records_then_register_to_headcount_models = (
...
+    _.each(years, (year) => {
+      age_groups_totals_by_year[group][year] = 0;
+    });
+  });
+  // convert the string that should be int to int
+  _.each(rows, (row) => {
+    _.each(years, (year) => {
+      age_groups_totals_by_year[row.dimension][year] += +row[year] || 0;
+    });
+  });
```
ktw1016, line 50:
```_.chain(rows)
    .groupBy("dimension")
    .map((dimension_arr, dimension_group) => [
      dimension_group,
      _.chain(years)
        .map((year) => [
          year,
          _.sumBy(dimension_arr, (row) => _.toNumber(row[year])),
        ])
        .fromPairs()
        .value(),
    ])
    .fromPairs()
    .value();
```

IvanFeng1:
https://github.com/TBS-EACPD/infobase/blob/87e704440f721c0555508462a8deaadbe4ad3314/server/src/models/people/populate.js#L63-L91

A lot of the variables that hold the data from csv files have the same fields. Should I change the `dimension` field for each variable to something specific to that csv file? Ex. `dimension` to `gender` as a field.

Stephen-ONeil:
> A lot of the variables that hold the data from csv files have the same fields. Should I change the `dimension` field for each variable to something specific to that csv file? Ex. `dimension` to `gender` as a field.

Nope, no need to worry about that. Think of them as "name spaced" to the collection, e.g. `employee_ex_lvl_rows`. Inside the context of that collection, they're unique. Outside the context of that collection they'd have to be referenced as something like `employee_ex_lvl_rows[index].avg_share`, which still retains the context of what file the dimensions are from (well, you'd never write that code, more likely to be processing all the rows through a chain of maps and filters, etc, which are once again thought of as entering back in to the specific context).

Aside from that, since the structure and use of these datasets _is_ very similar (aside from perhaps avg age), the consistency in dimension names is actually a positive. For instance, it will make it easier to write utility functions that can process any of these datasets, or easily access data from the same year across all of them.

That aside, the names in these processed csv `..._row` collections isn't too important. This file's just a script for processing the data and loading it in to MongoDB. The mongoose schemas are the interface you'll be accessing data via in the actual run-time code. Let me know if you want to get on a call to talk about those some time (but feel free to take a stab at them yourself first).

server/src/models/people/populate.js
```diff
@@ -30,17 +30,120 @@ const format_csv_records_then_register_to_headcount_models = (
...
+    get_standard_csv_file_rows("org_employee_age_group.csv")
+  )
+    .reject(["dept_code", "ZGOC"])
+    .map((row) => ({
+      ...row,
+      2016: +row[2016],
+      2017: +row[2017],
+      2018: +row[2018],
+      2019: +row[2019],
+      2020: +row[2020],
```
Stephen-ONeil, line 46: Lots of repetition of these years. All these would need to be maintained on each data update.

I'd prefer to have the csv itself switch from "wide" (years as column headers) to "long" (column for year dimension, row per year) but that can't change until we're actually ready to drop the old `Table` class way of loading this data in the client.

If you look at the current csv's though, you'll see that there's a common pattern to the columns. All of these datasets (small exception for avg age) could be processed by a utility like
```
const process_population_data = (csv_name) => _.chain(
    get_standard_csv_file_rows(csv_name)
  )
    .reject(["dept_code", "ZGOC"])
    .map(({dept_code, dimension, ...data_columns}) => ({
      dept_code,
      dimension,
      ..._.mapValues(data_columns, _.toNumber)
    }))
    .value();
```

Less repetition ([DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)), and no more hard-coded years (also sort of DRY, but more importantly "[single source of truth](https://en.wikipedia.org/wiki/Single_source_of_truth)" for the years from the csv column headers themselves).

Stephen-ONeil (changes requested):

server/src/models/people/models.js
```diff
@@ -1,69 +1,58 @@
...
+  parent_fkey_type,
+  fyear_type,
+  number_type,
+  str_type,
+  bilingual,
+} from "../model_utils.js";
+
+export default function (model_singleton) {
+  const EmployeeWholeDataSchema = mongoose.Schema({
+    dept_code: str_type,
```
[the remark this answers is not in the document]
Stephen-ONeil: Follow up, more of a next-steps thing once these by-department models are worked out, note that by cases above didn't cover the government level infographic. Generally, we do not want to make the gov level infographic load all the granular department data! What we tend to do is pre-calculate the government level totals separately and store them in their own mongo collection(s). Preferably all in one, since we _do_ have a limit on total number of collections, and burning them for a single document each would be a waste.

In some cases, the gov level documents may look exactly the same as the dept level documents, just without a `dept_code` field. In other cases, the gov level data may need additional pre-calculated stats. It depends on what data the current corresponding infographics uses. In the people data case, the gov panels do _not_ (currently) require any additional stats. Many of the financial panels, on the other hand, do (as they contain references to things like which department is the largest spender in addition to plain government totals; don't want to load all of the data just to calculate that client side, might as well re-calculate it server side). This is a big part of what I referred to when I said there's going to be feedback loops between the server side stuff and the client side stuff.

Note: we're free to save a lot of pre-calculated (derived) data in our databases because all our data is re-populated on every deploy and read-only in between them :tada:. Lots of optimizations we can make because of that. The schema design would be VERY different if this data was read-write, since storing duplicate/derived values in that case would just be asking for data inconsistencies.

Stephen-ONeil (changes requested):
Left a couple big comments with advice on mongoose schemas. Gave an example of how you could format it to leverage mongo's document structure but keep in mind that it's off the top of my head. You might end up tweaking it as you write the corresponding gql schemas and generally spend more time on this.

This may be dense and I'm not sure if I made it all clear enough. Chew it over then let me know if you have any questions!

server/src/models/people/populate.js
```diff
@@ -1,46 +1,87 @@
...
+    employee_fol_rows = process_employee_csv("org_employee_fol.csv");
+
+    employee_gender_rows = process_employee_csv("org_employee_gender.csv");
+
+    employee_region_rows = process_employee_csv("org_employee_region.csv");
+
+    employee_type_rows = process_employee_csv("org_employee_type.csv");
+  } catch (err) {
+    console.log(err);
+  }
```
Stephen-ONeil, line 76: Don't want to be try-catching here. Logs the error but then keeps on running. The DB population should fail early and loud on an error so that we don't accidentally let bad data get out in to prod.

server/src/models/people/populate.js
```diff
@@ -1,46 +1,87 @@
...
+    EmployeeType,
+  } = models;
+
+  let employee_age_rows,
+    employee_avg_age_rows,
+    employee_ex_lvl_rows,
+    employee_fol_rows,
+    employee_gender_rows,
+    employee_region_rows,
+    employee_type_rows;
```
Stephen-ONeil, line 22: Hmm, `let`s are usually a "code smell" to me since they normally indicate mutation. In this case they don't though, right? It's because the variables want to be used in the scope above the `try` block where they're actually declared.

I suggest below to not use the try/catch anyway, but in general there's a few ways you could get around this, e.g. put the try/catch in a function that returns the row variables as keys in an object, and then use a destructuring declaration like `const {a, b, c} = function()` to declare the variables back in to the scope you want them in.

Of course, that's not necessary without the try/catch.

server/src/models/people/populate.js
```diff
@@ -1,46 +1,87 @@
...
+      .map(({ dept_code, dimension, ...data_columns }) => ({
+        dept_code,
+        dimension,
+        ..._.mapValues(data_columns, _.toNumber),
+      }))
+      .groupBy("dept_code")
+      .map((dept_rows, dept_code) => ({
+        dept_code,
+        data: _.chain(dept_rows)
+          .map(({ dept_code, dimension, avg_share, ...data_columns }) => ({
```
Stephen-ONeil, line 35: Small nitpick, on line 26 `avg_share` is included in the thing called `data_columns`, but on line 35 it is excluded. Best to try not reusing names for things that are different in subtle ways like this. Maybe `numeric_columns` the first time and `values_by_year` the second time?

server/src/models/people/populate.js
```diff
@@ -1,46 +1,87 @@
...
+            dimension,
+            by_year: Object.keys(data_columns).map((key) => ({
+              year: key,
+              value: data_columns[key],
+            })),
```
Stephen-ONeil, line 40: More lodash-y
```
_.map(
  data_columns,
  (value, year) => ({year, value})
)
```

The general lodash pattern for the function args in iterators like `map` etc. is to take the value as the first arg, the key or index as the second arg, and the whole original array/object as the third arg (that third args hardly ever used though).

server/src/models/people/models.js
```diff
@@ -1,69 +1,66 @@
...
+            value: number_type,
+          },
+        ],
+        avg_share: number_type,
+      },
+    ],
+  });
 
-      const inst = new HeadcountModel(obj);
+  const EmployeeAveragesSchema = mongoose.Schema({
```
Stephen-ONeil, line 34: Small name nitpick, since this is specifically just average age right now (and we don't specifically plan for future average datasets), I'd just specifically name it `EmployeeAverageAgeSchema`.

server/src/models/people/models.js
```diff
@@ -1,69 +1,66 @@
...
-      try {
-        return this.model_name();
-      } catch (e) {
-        console.log(e);
-      }
-    }
-    static register(obj) {
-      const { dept_code } = obj;
+export default function (model_singleton) {
+  const EmployeeWholeDataSchema = mongoose.Schema({
```
Stephen-ONeil, line 18: Small name nitpick, because "whole data" doesn't mean too much to me, I'd call these `EmployeePopulationDataSchema` or something. The general family of datasets it applies to is the "population" data (different ways of slicing up the same total population numbers, all those datasets sum to the same total).

Stephen-ONeil (changes requested):
Some relatively minor notes. Reminder of my early comment about "next-steps thing[s]", should start working out the GraphQL schema and at least thing about how the current client setup would use it in practice, just in case that ends up inspiring DB schema changes (plus the likely separate government summary collection).

IvanFeng1:
> Some relatively minor notes. Reminder of my early comment about "next-steps thing[s]", should start working out the GraphQL schema and at least thing about how the current client setup would use it in practice, just in case that ends up inspiring DB schema changes (plus the likely separate government summary collection).


Should I calculate the government totals client side or server side?

Stephen-ONeil:
> > Some relatively minor notes. Reminder of my early comment about "next-steps thing[s]", should start working out the GraphQL schema and at least thing about how the current client setup would use it in practice, just in case that ends up inspiring DB schema changes (plus the likely separate government summary collection).
> 
> Should I calculate the government totals client side or server side?

Preferably server side. If you look at the gov level people infographics, there's nothing in there that needs dept level data so it'd be much more efficient to only send over the totals. On the server side, might as well calculate them at populate-time to further speed things up.

ktw1016:
You need id's for schemas
Also, instead of dept_code, I'd put org_id

ktw1016:
Code we wrote for `schema.js` on call (there could be errors):
```
const schema = `
extend type Org{
  employee_avg_age: [EmployeeAvgAge]
}
type EmployeeAvgAge {
  dept_code: String
  data: ByYearDataPoint
}
type ByYearDataPoint {
  by_year: [DataPoint]
}
type DataPoint {
  year: Float,
  value: Float
}
`;

export default function ({ models, loaders }) {
  const { EmployeeAvgAge } = models;
  const resolvers = {
    Org: {
      employee_avg_age: ({ dept_code }) => {
        return EmployeeAvgAge.find({ dept_code: dept_code });
      },
    },
  };
  return {
    schema,
    resolvers,
  };
}
```

IvanFeng1:
> You need id's for schemas
> Also, instead of dept_code, I'd put org_id

The people csv files use `dept_code` instead of `org_id` like in the covid csv files so it was easier to just use `dept_code`. I can add the `org_id` to the schema but I am not sure how to map the `org_id`s to the corresponding `dept_code`s. I checked on the graphql playground to see how they match but it seems the order of `dept_code`s according to `org_id`s isn't in alphabetical order like in the csv files.

Stephen-ONeil:
> The people csv files use `dept_code` instead of `org_id` like in the covid csv files so it was easier to just use `dept_code`. I can add the `org_id` to the schema but I am not sure how to map the `org_id`s to the corresponding `dept_code`s. I checked on the graphql playground to see how they match but it seems the order of `dept_code`s according to `org_id`s isn't in alphabetical order like in the csv files.

You could load the data from the igoc csv and use it it to map `dept_code`s to `org_id`s within the populate step.

ktw1016 (commented):
Do we wanna group all of these together, to have graphql structure like:
```
Gov{
 employee {
    employee_age...
    ....
  }
}
```
Just so that Gov, Org, etc don't get too huge and is more organized?

server/src/models/people/models.js
```diff
@@ -1,69 +1,123 @@
...
+          year: fyear_type(),
+          value: number_type,
+        },
+      ],
+    },
+  });
 
-      const model_name = this.get_model_name();
+  const EmployeeDataGovTotalsSchema = mongoose.Schema({
+    dimension: pkey_type(),
```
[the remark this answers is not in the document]
IvanFeng1: Do I add `id` only for mongoose schemas that do not already have `org_id`?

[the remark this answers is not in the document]
ktw1016: I meant for all of them but if org_id works for primary, we can just use org_id

server/src/models/people/schema.js
```diff
@@ -1,534 +1,79 @@
...
+    id: String
+    employee_age_totals: [EmployeeGovData]
+    employee_ex_lvl_totals: [EmployeeGovData]
+    employee_gender_totals: [EmployeeGovData]
+    employee_fol_totals: [EmployeeGovData]
+    employee_region_totals: [EmployeeGovData]
+    employee_type_totals: [EmployeeGovData]
+    employee_gov_avgs: [EmployeeGovAvgs]
+  }
+  type DataObjectWithDim {
```
ktw1016, line 31: Let's be a bit more descriptive like `EmployeeDataWithDimension` and `EmployeeDataWithoutDimension`

server/src/models/people/populate.js
```diff
@@ -1,46 +1,199 @@
...
+              value,
+            })),
+            avg_share,
+          }))
+          .value(),
+      }))
+      .value();
+
+  const process_employee_data_sums = (csv_name) => {
+    let raw_data = get_standard_csv_file_rows(csv_name);
```
ktw1016, line 37: Please change all the `let`s to `const` unless they need to be `let`

server/src/models/people/populate.js
```diff
@@ -1,46 +1,199 @@
...
+    let years = _.chain(raw_data[0])
+      .keys()
+      .toArray()
+      .map((element) => {
+        if (!isNaN(element)) {
+          return element;
+        }
+      })
+      .compact()
+      .value();
```
ktw1016, line 47: `keys()` should be returning an array already. You can also use `.filter` or `.reject` to filter out non year values

server/src/models/people/populate.js
```diff
@@ -1,46 +1,199 @@
...
+      .map((value, key) => ({
+        by_year: {
+          year: [key][0],
+          value: value,
+        },
+      }))
+      .value(),
+  };
+
+  let org_id_list = _.map(employee_age_rows, "org_id"); // all the csv files should have the same list of org_ids
```
ktw1016, line 137: You should use `igoc_rows` defined at the top here. It should probably be more accurate list of org ids

server/src/models/people/populate.js
```diff
@@ -1,46 +1,199 @@
...
+
+  let org_summary = _.map(org_id_list, (org_id) => ({
+    org_id: org_id,
+    employee_age_group: {
+      org_id: org_id,
+      data: _.find(employee_age_rows, (z) => {
+        return z.org_id == org_id;
+      }).data,
```
ktw1016, line 145: we can define utility function to use for all the repeats
`const find_by_org_id = (rows, org_id) => _.find(employee_age_rows, row => row.org_id === org_id).data`

server/src/models/people/populate.js
IvanFeng1: The list of `org_ids` for `igoc_rows` and the people csv files seem to be different with `igoc_rows` having more `org_ids` so can I keep using one of the people csv files for the list of `org_ids`

ktw1016: Having more `org_ids` is good. Just filter them out if it's empty

server/src/models/people/populate.js
ktw1016: Still haven't changed

server/src/models/people/populate.js
```diff
@@ -1,46 +1,184 @@
...
+          year: [key][0],
+          value: value,
+        },
+      }))
+      .value(),
+  };
+
+  const org_id_list = _.map(employee_ex_lvl_rows, "org_id"); // all the csv files should have the same list of org_ids
+  const find_by_org_id = (rows, org_id) =>
+    _.find(rows, (row) => row.org_id === org_id).data;
```
ktw1016, line 137: Now that I look, this could even return the summary itself like `const find_by_org_id = (rows, org_id) => ({
  org_id,
  data: _.find(...).data,
})`

server/src/models/people/populate.js
```diff
@@ -1,46 +1,184 @@
...
+          .value(),
+      }))
+      .value();
+
+  const process_employee_data_sums = (csv_name) => {
+    let raw_data = get_standard_csv_file_rows(csv_name);
+    let years = _.chain(raw_data[0])
+      .keys()
+      .filter((key) => {
+        return !isNaN(key);
```
ktw1016, line 41: Non issue and minor, but `reject(_.isNaN)` seem more intuitive

server/src/models/people/populate.js
```diff
@@ -1,46 +1,184 @@
...
+    employee_ex_lvl_totals: employee_ex_lvl_totals,
+    employee_gender_totals: employee_gender_totals,
+    employee_fol_totals: employee_fol_totals,
+    employee_region_totals: employee_region_totals,
+    employee_type_totals: employee_type_totals,
+    employee_gov_avgs: employee_gov_avgs,
+  };
+  return await Promise.all([
+    OrgEmployeeSummary.insertMany(org_summary),
+    GovEmployeeSummary.insertMany(gov_summary),
```
ktw1016, line 182: What about program summary?

Stephen-ONeil: People data doesn't exist below the org level

server/src/models/people/populate.js
IvanFeng1: Should I have a `org_id_list` with every org_id from `igoc.csv` and then filter out org_id that is not in the people csv files or should I keep the `org_id_list` but when I iterate over the list, only use the org_ids in the people files?

IvanFeng1:
https://github.com/TBS-EACPD/infobase/blob/67c5cb1d41b24ce500925218ee0a6c538164df09/server/src/models/people/populate.js#L37-L44

I don't get the same results when I use `isNaN` compared to `_.isNaN` for checking if a string can be converted into a number so I am sticking with `isNaN`

Stephen-ONeil:
According to the lodash docs, `_.isNaN` is a wrapper for `Number.isNaN`. The difference between the (legacyish) global `isNaN` and `Number.isNaN` is that `isNaN` will internally coerce the value to a number before checking and `Number.isNaN` will not.

In other words:
  - `Number.isNaN` only returns true if the argument is both of type number AND value `NaN`
  - `isNaN` will return true if, when coerced to type number, the argument has the value `NaN`

So, here, `isNaN` makes sense... ish. A string that can't be coerced to a number WILL return `true` with `isNaN` (false with `_.isNaN`). But some other values we likely don't want could still get through (for example, `null` casts to 0, `isNaN(null) === false`). We should probably explicitly cast to numbers, and then filter for numbers greater than 0 instead. Will filter out `NaN`s and falsey values that would cast to 0.

I feel like `_.isNaN(+value)` (manually casting to int, so kind of recreating the global `isNaN` behaviour) is something of a common pattern in the server populate code. Might not always be doing what we want it to, would have to think about the different cases. It may mostly be done because Mongoose is going to try to coerce (and throw on NaNs) when populating the DB anyway, but it might not actually be the most useful check to be making.

server/src/models/people/populate.js
ktw1016: yeah that sounds good to me

server/src/models/people/populate.js
```diff
@@ -1,46 +1,181 @@
...
+            .map((year) => [
+              year,
+              _.sumBy(dimension_arr, (row) => _.toNumber(row[year])),
+            ])
+            .fromPairs()
+            .map((value, key) => ({
+              year: [key][0],
+              value: value,
+            }))
+            .value(),
```
Stephen-ONeil, line 62: Could use a reduce, something like 
```
_.reduce(
  dimension_arr, // dimension_rows?
  (row, memo) => _.map(memo, ({year, value}) => ({year, value: value + row[year]}),
  _.map(years, (year) => ({year, value: 0}))
)
```

Map, filter, and reduce are something like the three primitives of functional data manipulation, can do most things in terms of them. The most simple use of a reduce is probably a sum. You don't _have_ to write everything in terms of them (lodash provides all its other utils to save us the time of re-implementing common patterns with them), but using them directly can help simplify things.

server/src/models/people/populate.js
```diff
@@ -1,46 +1,181 @@
...
+        ..._.mapValues(data_columns, _.toNumber),
+      }))
+      .groupBy("dept_code")
+      .map((dept_rows, dept_code) => ({
+        org_id: _.find(
+          igoc_rows,
+          (igoc_row) => igoc_row.dept_code === dept_code
+        ).org_id,
+        data: _.chain(dept_rows)
+          .map(({ dept_code, dimension, avg_share, ...values_by_year }) => ({
```
Stephen-ONeil, line 25: Hmm, since we rely on all the csvs having this structure, probably worth writing a function that checks the csv headers and throws if there's anything unexpected. It'd be good to run it as a first step in each of these common csv processing utils.

server/src/models/people/populate.js
```diff
@@ -1,46 +1,181 @@
...
+      employee_ex_lvl: find_by_org_id(employee_ex_lvl_rows, org_id),
+      employee_fol: find_by_org_id(employee_fol_rows, org_id),
+      employee_gender: find_by_org_id(employee_gender_rows, org_id),
+      employee_region: find_by_org_id(employee_region_rows, org_id),
+      employee_type: find_by_org_id(employee_type_rows, org_id),
+      employee_avg_age: find_by_org_id(employee_avg_age_rows, org_id),
+    }))
+    .reject((row) => {
+      return row.employee_age_group === null;
+    })
```
Stephen-ONeil, line 164: Any reason for specifically checking age group? If not, I'd suggest leaving a comment saying it's arbitrary (and pointing out that the assumption is that all the population tables should contain the same sets of orgs, seeing as they're just different ways of slicing the same data). Otherwise future devs might assume there's something special about this choice.

Stephen-ONeil: Documenting assumptions and clarifying special meaning vs arbitrary choices are always good times for comments (unless it's possibly to document that in the code directly, via naming).

server/src/models/index.js
```diff
@@ -27,7 +27,7 @@ const sub_module_defs = _.compact([
   // pses,
   // sobjs,
   // search,
-  // people,
+  people,
```
Stephen-ONeil, line 30: Right before merging, it'll be worth commenting this back out just so it's clear it's not in use on master

server/src/models/people/populate.js
```diff
@@ -1,46 +1,181 @@
...
+      }))
+      .value(),
+  };
+
+  const full_org_id_list = _.map(igoc_rows, "org_id");
+
+  const find_by_org_id = (rows, org_id) => {
+    const entry_by_org_id = _.find(rows, (row) => row.org_id === org_id);
+    if (entry_by_org_id !== undefined) {
+      // make sure the org_id is in the rows
```
Stephen-ONeil, line 142: Comment seems unnecessary? Or am I missing some nuance?

Stephen-ONeil (changes requested):
Some small suggestions

I expect the mongoose and GraphQL schemas will change a little bit once it's being used in practice (e.g. once the RPB can use GraphQL data, there will need to be a way to request all data per population dataset from the schema root), but I won't try to anticipate all that now. Better to be flexible about changing it as we go.

Even though it's likely to change eventually, it's probably worth adding snapshot test coverage for the current schema before merging. Short term there's coverage for the current design, long term it'll remind us to test this stuff. Let me know if you need help with that.

IvanFeng1:
https://github.com/TBS-EACPD/infobase/blob/575131e696abea10f05821edc204da49e1409689/server/src/models/people/populate.js#L12-L23

I added this to check whether the csv files have the required headers but I am not sure what kind of error I should throw.

Issue #1022, opened 2021-04-15
Clean up `extended_bootstrap_css` and `common_css`


Stephen-ONeil:
They serve different purposes, but they have similar major problems. They're both organized poorly and absolutely stuffed full of styles (and style overrides) that are likely either very legacy or totally unused.

`extended_bootstrap_css` should only be:
    - bootstrap overrides (and only the ones that we really still want; these were all from bootstrap 3 in the first place)
    - css specific to the index html files (UX optimization for initial load, just something we have to live with until we can change our index html file management)
    
`common_css` should probably be a collection of very generic styles/css utils, organized in a bunch of small modules. Rather than importing them all as a bundle in `bootstrapper.js` (... the verb, not the noun), individual components/modules should import them as needed. A lot more book keeping to do, but it should make it way easier to keep track of what CSS is actually being used and where. 

Not totally inflexible on this, might be proposing an extreme solution. Maybe there's still room for some non-bootstrap (the noun) CSS globally required during the app bootstrapping (verb). Open for debate.

PR #1184, opened 2021-07-12
Tippy Styles in extended_bootstrap_css

patfire12345:


Stephen-ONeil (changes requested):
At least some of these are still in use. They're overwrites/extensions of the default styles for `tippy.js`, our tooltip library. Check some glossary tooltips etc on the dev link, they've lost their styling.

On the other hand, it'd be good to move these styles out of `bootstrap-wet-fixes-extensions.scss`. We use `tippy.js` via out custom `TooltipActivator` component. These styles could be in a `scss` module for that component to make it clear where they're relevant.

PR #1186, opened 2021-07-12
TS:  Multiple stories to typescript

csong202:
This PR is for stories whose corresponding component has already been converted to ts. Separate PRs will be made for components that have not been converted to ts. The stories in modals_and_popovers will be in a separate PR

Stories converted:
- CardList
- DisplayTable
- LabeledBox
- LabeledTable
- LeafSpinner
- RadioButtons
- Select
- SortDirection
- Tombstones
- Typeahead

### DIFF
--- a/client/src/components/CardList/CardList.stories.tsx
+++ b/client/src/components/CardList/CardList.stories.tsx
@@ -1,17 +1,21 @@
+import { Story, Meta } from "@storybook/react";
 import React from "react";
 
-import { CardList } from "./CardList";
+import { CardList, CardListElementProps } from "./CardList";
 
 export default {
   title: "CardList",
   component: CardList,
-};
+} as Meta;
 
-const Template = (args) => <CardList {...args} />;
+type CardListProps = React.ComponentProps<typeof CardList>;
+
+const Template: Story<CardListProps> = (args) => <CardList {...args} />;
 
 const elements = [
   {
     display: "Non-link Title",
+    href: "#",
     children: [
       { display: "Non-link Title 1" },
       { display: "Non-link Title 2" },
@@ -27,7 +31,7 @@ const elements = [
       { display: "Href 3", href: "#" },
     ],
   },
-];
+] as unknown as CardListElementProps[];
 
 export const Basic = Template.bind({});
 Basic.args = {
--- a/client/src/components/CardList/CardList.stories.tsx
+++ b/client/src/components/CardList/CardList.stories.tsx
@@ -12,14 +12,14 @@ type CardListProps = React.ComponentProps<typeof CardList>;
 
 const Template: Story<CardListProps> = (args) => <CardList {...args} />;
 
-const elements = [
+const elements: CardListElementProps[] = [
   {
     display: "Non-link Title",
     href: "#",
     children: [
-      { display: "Non-link Title 1" },
-      { display: "Non-link Title 2" },
-      { display: "Non-link Title 3" },
+      { display: "Non-link Title 1", href: "" },
+      { display: "Non-link Title 2", href: "" },
+      { display: "Non-link Title 3", href: "" },
     ],
   },
   {
@@ -31,7 +31,7 @@ const elements = [
       { display: "Href 3", href: "#" },
     ],
   },
-] as unknown as CardListElementProps[];
+];
 
 export const Basic = Template.bind({});
 Basic.args = {
--- a/client/src/components/CardList/CardList.stories.tsx
+++ b/client/src/components/CardList/CardList.stories.tsx
@@ -1,7 +1,7 @@
 import { Story, Meta } from "@storybook/react";
 import React from "react";
 
-import { CardList, CardListElementProps } from "./CardList";
+import { CardList } from "./CardList";
 
 export default {
   title: "CardList",
@@ -12,7 +12,7 @@ type CardListProps = React.ComponentProps<typeof CardList>;
 
 const Template: Story<CardListProps> = (args) => <CardList {...args} />;
 
-const elements: CardListElementProps[] = [
+const elements = [
   {
     display: "Non-link Title",
     href: "#",
--- a/client/src/components/CardList/CardList.stories.tsx
+++ b/client/src/components/CardList/CardList.stories.tsx
@@ -1,6 +1,8 @@
 import { Story, Meta } from "@storybook/react";
 import React from "react";
 
+import { ComponentProps } from "src/types/util_types.d";
+
 import { CardList } from "./CardList";
 
 export default {
@@ -8,7 +10,7 @@ export default {
   component: CardList,
 } as Meta;
 
-type CardListProps = React.ComponentProps<typeof CardList>;
+type CardListProps = ComponentProps<typeof CardList>;
 
 const Template: Story<CardListProps> = (args) => <CardList {...args} />;
 
--- a/client/src/components/CardList/CardList.stories.tsx
+++ b/client/src/components/CardList/CardList.stories.tsx
@@ -17,9 +17,9 @@ const elements: CardListElementProps[] = [
     display: "Non-link Title",
     href: "#",
     children: [
-      { display: "Non-link Title 1", href: "" },
-      { display: "Non-link Title 2", href: "" },
-      { display: "Non-link Title 3", href: "" },
+      { display: "Non-link Title 1" },
+      { display: "Non-link Title 2" },
+      { display: "Non-link Title 3" },
     ],
   },
   {
--- a/client/src/components/CardList/CardList.tsx
+++ b/client/src/components/CardList/CardList.tsx
@@ -6,7 +6,7 @@ interface CardListElementChildProps {
   href: string;
 }
 
-interface CardListElementProps {
+export interface CardListElementProps {
   display: string | React.ReactNode;
   href: string;
   children: CardListElementChildProps[];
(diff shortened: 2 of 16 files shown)


Stephen-ONeil (changes requested):
Checkout out the branch locally to poke at some types, realized that we had poor type checking across all the story books (missed this in prior story TS PRs because I didn't check them out locally). It's the `Template.bind({})` pattern, was swallowing up type information.

Googled up [this issue](https://github.com/storybookjs/storybook/issues/11916) and caught what we were missing, we should add `"strictBindCallApply": true` to `client/tsconfig.json`.

With `"strictBindCallApply": true`, the type system will actually know that `template_instance.args` has type `Partial<WhateverCorrespondingStoryProps>`. Right now it has type `any`, which is why we've got that pattern of manually declaring the type of the defaults for the story args.

I'll hotifix `"strictBindCallApply": true` in to the config on master since it's something we should already have had. We may or may not need to clean up the existing TS stories a little after that. Rebase the branch off of master after that and try tightening up the typing as needed.

client/src/components/CardList/CardList.stories.tsx
```diff
@@ -27,7 +31,7 @@ const elements = [
       { display: "Href 3", href: "#" },
     ],
   },
-];
+] as unknown as CardListElementProps[];
```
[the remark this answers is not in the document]
Stephen-ONeil: To be clear, with `"strictBindCallApply": true` none of this will be necessary. `Basic.args` will know its own types, so `elements` will infer that it should be `CardListElementProps[]`.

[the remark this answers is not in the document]
csong202: I forgot that having href with empty string would have the same effect as having no href for the Non-link Title elements

client/src/components/CardList/CardList.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+export default {
+  title: "CardList",
+  component: CardList,
+} as Meta;
+
+type CardListProps = React.ComponentProps<typeof CardList>;
+
+const Template: Story<CardListProps> = (args) => <CardList {...args} />;
+
+const elements: CardListElementProps[] = [
```
Stephen-ONeil, line 15: Exporting/importing this type isn't necessary anymore. The type checking will be applied when element is used inside `Basic.args` now.

You might need to reset VS Code/you build scripts to refresh against the changed tsconfig rule.

client/src/components/CardList/CardList.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+  {
+    display: "Non-link Title",
+    href: "#",
+    children: [
+      { display: "Non-link Title 1", href: "" },
+      { display: "Non-link Title 2", href: "" },
+      { display: "Non-link Title 3", href: "" },
```
Stephen-ONeil, line 22: Seems like a typing mistake in the actual `CardList` props, shouldn't `href` just be optional?

client/src/components/DisplayTable/DisplayTable.stories.tsx
```diff
@@ -27,16 +31,23 @@ const common_column_configs = {
   price: {
     index: 1,
     header: "Price",
-    is_summable: true,
+    is_sortable: true,
   },
-};
+} as unknown as ColumnConfigProps;
```
Stephen-ONeil, line 36: Should drop this. Same as before, this is actually forcing it to cast to this type even though it doesn't actually satisfy it. The proper type checking should happen inside `Basic.args` below.

client/src/components/DisplayTable/DisplayTable.stories.tsx
```diff
@@ -48,13 +59,13 @@ const table_with_URL_column_configs = {
...
-    formatter: (id) => (
+    formatter: (id: CellValue) => (
       <a
         href={`https://en.wikipedia.org/wiki/List_of_pizza_varieties_by_country#${id}`}
         target="_blank"
         rel="noopener noreferrer"
       >
-        {country_lookup[id]}
+        {country_lookup[id as Country]}
       </a>
```
Stephen-ONeil, line 69: I think, other type issues asside, this could be `(id: Country) => (`. `Country` is just a union of strings, and `Strings` is a valid `CellValue` type.

client/src/components/DisplayTable/DisplayTable.stories.tsx
Stephen-ONeil: Going to have the type system complaining about a bunch of missing fileds in `common_column_configs` once you do that and the proper type checking applies. Either the typing on `DisplayTable` is bad (missing optional flags again) or this config is actually incomplete. 

Edit: oh, yup, the `DisplayTable` typing is junk, my bad for letting that get past review. A bunch of column config props that have defaults are still marked as required. Part of a history of us doing a bad job of typing React components with default props, refer to this if you're going to try to fix it https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/default_props/

Stephen-ONeil (changes requested):

client/src/components/CardList/CardList.tsx
```diff
@@ -2,12 +2,12 @@ import _ from "lodash";
...
-  display: string | React.ReactNode;
-  href: string;
+  display: React.ReactNode;
+  href?: string;
 }
 
 interface CardListElementProps {
-  display: string | React.ReactNode;
+  display: React.ReactNode;
   href: string;
```
Stephen-ONeil, line 11: Haven't looked in the render, but I think this top-level `href` prop should also be optional? Double check, might have to update the story too depending.

client/src/components/DisplayTable/DisplayTable.stories.tsx
```diff
@@ -66,7 +76,7 @@ TableWithURL.args = {
 };
 
 const custom_utils = {
-  downloadCsvUtil: null,
+  downloadCsvUtil: <></>,
```
Stephen-ONeil, line 79: I think you can just leave this property out of the object.

client/src/components/DisplayTable/DisplayTable.tsx
```diff
@@ -84,7 +84,7 @@ const get_column_config_defaults = (index: number) => ({
 type ColumnConfig = Partial<ReturnType<typeof get_column_config_defaults>> & {
   index: number;
   header: string;
-  formatter: FormatKey | ((val: CellValue) => string | React.ReactNode);
+  formatter: FormatKey | ((val: CellValue) => React.ReactNode);
```
Stephen-ONeil, line 87: :+1: good catch, thanks

client/src/components/Typeahead/Typeahead.stories.tsx
```diff
@@ -43,19 +44,38 @@ const data = [
...
-const filter = (query, datum) =>
-  _.includes(_.lowerCase(datum.title), _.lowerCase(query));
+const filter = (
+  query: string,
+  datum: Record<
+    string,
+    | string
+    | JSX.Element
+    | ((...args: (React.ReactNode | string | (() => void))[]) => void)
+  >
```
Stephen-ONeil, line 54: Could this be
```
{
  title: string;
  [string]: string | JSX.Element | ((...args: (React.ReactNode | string | (() => void))[]) => void); // maybe reasonable for this to be type any
}
```
?

Wouldn't need to cast the `title` property to string below then.

Stephen-ONeil (changes requested):

client/src/components/DisplayTable/DisplayTable.stories.tsx
```diff
@@ -37,24 +46,25 @@ Basic.args = {
...
     ...common_column_configs.label,
-    formatter: (id) => (
+    formatter: (id: CellValue) => (
       <a
         href={`https://en.wikipedia.org/wiki/List_of_pizza_varieties_by_country#${id}`}
         target="_blank"
         rel="noopener noreferrer"
       >
-        {country_lookup[id]}
+        {typeof id === "string" ? country_lookup[id] : <></>}
```
csong202, line 67: Could I just cast id to string here?

Stephen-ONeil: TS is supposed to help us write safer run time code. Painful type system situations could be a good hint that more run time checks should be added (ideally ones that type narrow). Best I could come up with that's both TS-correct and runtime safe is something like this:
```
const country_lookup = {
  Italy: "Italy",
  Korea: "Korea",
  Canada: "Canada",
  United_States: "United States",
  United_Kingdom: "United Kingdom",
};

const table_with_URL_column_configs = {
  ...common_column_configs,
  label: {
    ...common_column_configs.label,
    formatter: (label: CellValue) => {
      const country_name = _.get(country_lookup, _.toString(label));

      if (typeof country_name !== "undefined") {
        return (
          <a
            href={`https://en.wikipedia.org/wiki/List_of_pizza_varieties_by_country#${label}`}
            target="_blank"
            rel="noopener noreferrer"
          >
            {country_name}
          </a>
        );
      } else {
        throw new Error("Something to explain what this situation might be");
      }
    },
  },
};
```

In theory we could "know" `label` will always be a string (and use an `as` to cast its type) but that'll still leave TS complaining that `string` is too broad for `country_lookup` (note I removed the `Record<string,string>` type) which is not wrong. So either way this could use an extra layer of runtime checking for whether the `label` value was actually in `country_lookup`.

Stephen-ONeil: ... it's still pretty painful though, don't want to be wrestling with the over-broad `CellValue` type on every (TS) use of `DisplayTable`. I'll think about if it can be made easier to use with generics or something.

Stephen-ONeil (approved):
:+1:

PR #1187, opened 2021-07-12
Updating h1 font within extended_bootstrap_css

patfire12345:
Files where h1 tags are used:

- [x] EstimatesComparison
- [x] glossary
- [x] a11y_home (doesn't look affected by h1 change)
- [x] home (doesn't look affected by h1 change)
- [x] IgocExplorer
- [x] Infographic
- [x] metadata
- [x] IsolatedPanel
- [x] PanelInventory
- [x] SingleServiceRoute
- [x] index (not sure if this is affected by the h1 change)
- [x] TablePicker
- [x] Survey
- [x] TagExplorer
- [x] TreeMap
- [x] index.hbs.html (not sure if this is affected by the h1 change)

### DIFF
--- a/client/src/EstimatesComparison/EstimatesComparison.scss
+++ b/client/src/EstimatesComparison/EstimatesComparison.scss
@@ -30,5 +30,5 @@
 h1 {
   margin-top: 20px;
   margin-bottom: 20px;
-  border-bottom: 3px solid $highlightColor;
+  border-bottom: 3px solid #da3a38;
 }
--- a/client/src/EstimatesComparison/EstimatesComparison.scss
+++ b/client/src/EstimatesComparison/EstimatesComparison.scss
@@ -26,3 +26,9 @@
   }
   justify-content: space-evenly;
 }
+
+h1 {
+  margin-top: 20px;
+  margin-bottom: 20px;
+  border-bottom: 3px solid $highlightColor;
+}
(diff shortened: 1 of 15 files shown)


Stephen-ONeil (changes requested):
Lookout, the browser applies all CSS globally! Local scss modules for a given component, etc, are for source code organization and bundle optimization (i.e. don't need to load the styles until you actually need the component).

One loaded in to the actual page, the styles are no longer related to the original module in any way. The [specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) of all selectors are used to determine which styles are applied to any given element on the page. If there's a bunch of selectors just on `h1`, the specificity will be determined by position (e.g. whatever the latest one added is). Doesn't matter here because all the styles are the same, this is just creating unnecessary duplication (and room for future bugs if any particular instance drifts).

The global scope of style sheets is how we can have the central stylesheets in the first place. The problem with the central stylesheets (what #1022 was about) is that they're not well organized and haven't been well maintained. 

Moving styles outside of the central style sheets _will_ probably be part of that clean up, but most likely only when those styles are only used in one specific place (and therefore don't belong in the central style sheets in the first place). Overrides or styles that _should_ apply globally (these `h1` styles are both) still belong in the central sheets.

Let's talk on slack again about what approaches to #1022 make the most sense. Message me when you're free.

PR #1190, opened 2021-07-13
Service Inventory 2021

ktw1016:
TODO:

- [ ] Make individual services searchable (#1196 but will be opening a new one)
- [ ] Ask pipeline to add/remove glossary (High volume services, CHANNEL)
- [x] **Specify year for all panels where necessary like “In 2018-19 …"**
- [x] Delivery channels pie chart year selector (can put in both panels)


Stephen-ONeil:
TODO:
  - [ ] check displays of "other" type service standards
  - [ ] write further TODOs

ktw1016:
To Mike: 
For service id 1003 (`Employment Insurance (EI) Benefits`) and standard id 1456, (`EI payments are accurate`):
This standard performance is displayed accurately on InfoBase. All of the fields were empty (hence `N/A`), and `is_target_met=1`. Unfortunately, all of the detailed data was left as comment.

This standard is shown as `Met` on InfoBase because target type is `Other type of target`

As discussed long ago with you and Ambreen, logic for determining the standard is as follows:
```
if (`target_type` is NOT `Other type of target` AND
          `lower` is empty AND
          `count` is empty AND
          `met_count` is empty), then status is no data available
Otherwise, use whatever is in the data.
```

Jmikelittle:
This is a complicated one, because want to retain the ability for departments to indicate pass/fail on standards with target type: 'other'. It just looks silly.

Issue #1189, opened 2021-07-12
Review all tsx files for React components with default props


Stephen-ONeil:
Review and apply guides like this one https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/default_props/

A number of TS ports got merged where I didn't know to look out for this yet, oops. E.g. `DisplayTable.tsx` `ColumnKeyProps` type has a bunch of required keys, with comments beside them documenting that they have defaults... This is a more complicated case and might need a fair bit of clean up in the process, just the one I noticed that finally made me open an issue over it.

PR #1191, opened 2021-07-13
Cleaned up tsx components that use 'defaultProps' 

IvanFeng1:
Regarding issue #1189

### DIFF
--- a/client/docs/typescript/react-default-props.md
+++ b/client/docs/typescript/react-default-props.md
@@ -62,12 +62,14 @@ const SomeComponentDefaultProps = {
 ```
 
 2.  - Problem: as mentioned, there's magic inside the class component. The type `SomeComponentProps` itself still thinks the props with defaults are required, so outside of the class component declaration it isn't that useful. For an additional gotcha, the standard react utility type pattern `React.ComponentProps<typeof SomeComponentProps>` doesn't recover the desired type either!
-    - Fix: there's a pattern that does work, we have a utility type that implements it in `src/types/util_types.d.ts`, exported as `ComponentProps`. If you use that, you'll recover the type information with the correct mix of required/optional props.
+    - Fix: there's a pattern that does work, we have a utility type that implements it in `src/types/util_types.d.ts`, exported as `ComponentProps`. If you use that, you'll recover the type information with the correct mix of required/optional props
 
 ## Function Components
 
 The straightforward case. Just type it as you would any function with default values.
 
-_Do not_ use `React.FunctionalComponent` (https://github.com/facebook/create-react-app/pull/8177) or `SomeComponent.defaultProps` (set to be deprecated anyway).
+_Do not_ use `React.FunctionalComponent` (https://github.com/facebook/create-react-app/pull/8177)
 
-_Do_ use our util type `ComponentProps<typeof SomeComponent>` instead of `React.ComponentProps` if you need to reuse the prop types in an external module.
+_Do not_ use `SomeFunctionalComponent.defaultProps` (set to be deprecated, not going to get good typing)
+
+_Do_ use our util type `ComponentProps<typeof SomeComponent>` instead of `React.ComponentProps` if you need to recover the prop types in an external module
--- a/client/docs/typescript/react-default-props.md
+++ b/client/docs/typescript/react-default-props.md
@@ -12,14 +12,62 @@ I'm going to jump straight to documenting the solutions. The reasoning and secon
 
 ## Class Components
 
-The more complicated case.
+There's a little bit of React/TypeScript magic behind the scenes, so the normal rules aren't exactly followed. I'll point out where that happens.
 
-TODO: finalize pattern in practice and document it here.
+Steps for a simple case (additional patterns for tricky cases below):
+
+1. declare the prop defaults, e.g.
+
+```
+const SomeComponentDefaultProps = {
+ title: "bleh",
+ disabled: false,
+};
+```
+
+(Reminder: if a prop like`title` can be a string _or_ react component, use the `React.ReactNode` type)
+
+2. declare the prop types/interface, unioned or extending from the type of the defaults. Do not re-declare types for any props with defaults!
+
+```
+const SomeComponentDefaultProps = ...
+type SomeComponentProps = typeof SomeComponentDefaultProps & {
+  style?: React.CSSProperties;
+};
+```
+
+3. declare the class with the combined prop type and the default props
+
+```
+const SomeComponentDefaultProps = ...
+type SomeComponentProps = ...
+class SomeComponent extends React.Component<SomeComponentProps> {
+  static defaultProps = SomeComponentDefaultProps;
+  ...
+```
+
+Done! Looks a little weird though, right? If you look at `SomeComponentProps`, the props with defaults do not appear to be optional! This is the bit of magic I mentioned, the type system magically knows that they _are_ optional when actually using the component because of `static defaultProps` (guess React is big enough that TS devs worked this out for them). The reason this is preferable to declaring the props with defaults as optional explicitly, the standard way, is that if you do that then the type system will include `undefiend` as a possible value even though we know the component's `defaultProps` would prevent that.
+
+Note: totally still room for optional props that do not take defaults. I included one in step 2 of the example above to point that out. Be careful no to confuse optional props and props with defaults!
+
+Additional gotchas:
+
+1.  - Problem: the type infered from the default might not be complete
+    - Fix: just use an `as MoreComplexType` statement within the default props declaration, e.g.
+
+```
+const SomeComponentDefaultProps = {
+ height: 100 as string | number,
+};
+```
+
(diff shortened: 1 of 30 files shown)


client/src/components/CheckBox/CheckBox.tsx
```diff
@@ -41,12 +36,12 @@ export class CheckBox extends React.Component<CheckBoxProps> {
...
-      isSolidBox,
+      color = primaryColor,
+      isSolidBox = false,
       container_style,
       checkbox_style,
       label_style,
-      checkmark_vertical_align,
+      checkmark_vertical_align = 0.1,
       disabled,
     } = this.props;
```
Stephen-ONeil, line 46: For a class component, `static defaultProps` is still the correct solution I think. The class could be, now or in the future, accessing `this.props` in multiple places (e.g. constructor, life cycle methods, etc). Would have to manually repeat these defaults in all `props` destructuring/direct accessing of `this.props`. Oof, also in things like `preProps` args to certain life cycle methods.

I think the advice about setting defaults inside prop destructuring is for functional components. Totally makes sense in that case.

Stephen-ONeil (changes requested):
I was thinking of this approach my self
```
type GreetProps = typeof Greet.defaultProps & {
  age: number;
};

class Greet extends React.Component<GreetProps> {
  static defaultProps = {
    age: 21,
  };
  /*...*/
}

// Type-checks! No type assertions needed!
let el = <Greet age={3} />;
```
but, as the cheat sheet that recommends it notes, while this works as desired for the component's props (knows age is not required)  something like `const args: GreetProps = {}` 
 or `const args: React.ComponentProps<typeof Greet> = {}` will have type errors about `age` being missing, which means it isn't actually a good solution for the case that brought this up (using prop types in storybooks).

If we used the above pattern, we'd have to use this following utility in the place of the `React.ComponentProps` utility
```
type ComponentProps<T> = T extends
  | React.ComponentType<infer P>
  | React.Component<infer P>
  ? JSX.LibraryManagedAttributes<T, P>
  : never;
``` 

Which could be a small gotcha, since any React-TS tutorials will keep telling devs to use the React util type.. A linter rule could enforce that though.

Did you have any alternate reason why you took the approach you did? I'm mostly following that one cheat sheet's advice, so if you dug in to the problem deeper before you tackled this please let me know!

client/src/components/CheckBox/CheckBox.tsx
```diff
@@ -16,19 +16,14 @@ export interface CheckBoxProps {
   label: string; // required text for checkbox
   active?: boolean; // required, equivalent of 'checked'
   color?: string; // defaults to primaryColor
-  isSolidBox: boolean; // defaults to false
+  isSolidBox?: boolean; // defaults to false
   container_style?: React.CSSProperties; // style for div, containing checkbox and label
   checkbox_style?: React.CSSProperties; // style for checkbox
   label_style?: React.CSSProperties; // style for label
-  checkmark_vertical_align: number; // defaults to 0.1
+  checkmark_vertical_align?: number; // defaults to 0.1
```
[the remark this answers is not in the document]
IvanFeng1: I chose my method because it was on the cheatsheet and it seems easy to implement. Should I remove the comments that say the default value for the props?

IvanFeng1:
> I was thinking of this approach my self
> 
> ```
> type GreetProps = typeof Greet.defaultProps & {
>   age: number;
> };
> 
> class Greet extends React.Component<GreetProps> {
>   static defaultProps = {
>     age: 21,
>   };
>   /*...*/
> }
> 
> // Type-checks! No type assertions needed!
> let el = <Greet age={3} />;
> ```
> 
> but, as the cheat sheet that recommends it notes, while this works as desired for the component's props (knows age is not required) something like `const args: GreetProps = {}`
> or `const args: React.ComponentProps<typeof Greet> = {}` will have type errors about `age` being missing, which means it isn't actually a good solution for the case that brought this up (using prop types in storybooks).
> 
> If we used the above pattern, we'd have to use this following utility in the place of the `React.ComponentProps` utility
> 
> ```
> type ComponentProps<T> = T extends
>   | React.ComponentType<infer P>
>   | React.Component<infer P>
>   ? JSX.LibraryManagedAttributes<T, P>
>   : never;
> ```
> 
> Which could be a small gotcha, since any React-TS tutorials will keep telling devs to use the React util type.. A linter rule could enforce that though.
> 
> Did you have any alternate reason why you took the approach you did? I'm mostly following that one cheat sheet's advice, so if you dug in to the problem deeper before you tackled this please let me know!

 If I am only going to clean up the default props for the class components and not the stories, I don't have to worry about the type error right?

client/src/components/CheckBox/CheckBox.tsx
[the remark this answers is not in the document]
Stephen-ONeil: Yeah, none of those comments should be necessary. Maybe give more self-descriptive prop names in some cases too, if it's too vague without the comment.

Stephen-ONeil:
Should still be concerned about the stories. Any that have been ported to TS may have inherited the bad typing/done hacky things to work around it. With respect to what I was talking about, where the alternate (possibly better) approach requires using a custom `ComponentProps` util type, the storybooks are in theory just an example of somewhere that currently uses `React.ComponentProps`. Would have to consider that anywhere else might use it in the future.

It's a little bit more rearranging to do, plus the need to add a custom `ComponentProps` util (and maybe write a lint rule to enforce using it over `React.ComponentProps`) but I think it's worth doing over your current approach. It's somewhat simpler to mark optional fields in the props and separately declare default values, but the fact that it decouples those things means it's more likely to wind up broken in the future. Better to explicitly couple the default values and the prop types even if it's more "boilerplate" code.

TS is all boilerplate anyway, haha.

client/src/components/CheckBox/CheckBox.stories.tsx
```diff
@@ -8,7 +8,13 @@ export default {
   component: CheckBox,
 } as Meta;
 
-type CheckBoxProps = React.ComponentProps<typeof CheckBox>;
+type ComponentProps<T> = T extends
+  | React.ComponentType<infer P>
+  | React.Component<infer P>
+  ? JSX.LibraryManagedAttributes<T, P>
+  : never;
```
Stephen-ONeil, line 15: Since we'll want to be reusing this a bunch, and may or may not have to update the pattern in the future, please move it to `src/types/utility_types.d.ts` and import from there as needed :+1:

client/src/components/CheckBox/CheckBox.tsx
```diff
@@ -9,19 +9,19 @@ import { IconCheckmark } from "src/icons/icons";
 
 import "./CheckBox.scss";
 
-export interface CheckBoxProps {
-  disabled?: boolean; // greys out checkbox and make it non-clickable if true
-  onClick?: (id: string) => void; // (id) => {}
+export type CheckBoxProps = typeof CheckBox.defaultProps & {
```
Stephen-ONeil, line 12: Would be good if you could catch any react components that are exporting their props and clean that up too. Should use the `ComponentProps` utility type where ever the props are needed.

Stephen-ONeil (changes requested):
Gotcha to look out for, for some reason if there's overlap between the keys in the explicit prop types and the types inferred from the defaults (specifically when accessed via the class) then the resulting type is `any`.

image

Remove any of the props with defaults from the explicit prop types and it resolves as expected.

image

**OR** declare the defaults as a const outside of the class declaration

image

I'm going to try and figure out why that first case resolves to `any`, not totally sure why right now. Personally, I prefer the first fix as there's less duplication and the prop types seen inside the component are clearer.

If the correct type for one of the props with defaults is more complex than what ever's inferred from the default value, you can us `<prop>: <default> as <actual type>` inside `defaultProps` to clarify.

client/docs/typescript/react-default-props.md
```diff
@@ -0,0 +1,76 @@
...
+```
+const SomeComponentDefaultProps = {
+ height: 100 as string | number,
+};
+```
+
+2.  - Problem: as mentioned, there's magic inside the class component. The type `SomeComponentProps` itself still thinks the props with defaults are required, so outside of the class component declaration it isn't that useful. For an additional gotcha, the standard react utility type pattern `React.ComponentProps<typeof SomeComponentProps>` doesn't recover the desired type either!
+    - Fix: there's a pattern that does work, we have a utility type that implements it in `src/types/util_types.d.ts`, exported as `ComponentProps`. If you use that, you'll recover the type information with the correct mix of required/optional props
+
+## Function Components
```
IvanFeng1, line 68: Just making sure, is the straightforward case for writing default values for functional components supposed to be like this: 
```
const Component = ({ prop1 = false, prop2 = 'My Prop' }) => (
  <div></div>
)
```

Stephen-ONeil: Yup, plus whatever function typing syntax you want to use, `({ prop1 = false, prop2 }: {prop1: boolean, prop2: string}) => ...` etc.

Note that with that example, you still have to be explicit about the `prop1` type, it won't infer boolean for you. The point is that you just use the standard TS function syntax, no special/magic stuff like in the class component case.

PR #1193, opened 2021-07-14
Deleted tables.scss

patfire12345:
Seems like none of the table styles are currently being applied.


Stephen-ONeil (changes requested):
Took a side by side look at this dev link and master's, caught one thing. The `center-text` class from `tables.scss` was being used in a few places, both in `DisplayTable` and in a few other places around the site.

Looks like Bootstrap has its own utility class for text centering, `text-center`. I think we should delete `center-text` here and just update all the uses of it to `text-center`.

I think a good general tip for this CSS cleanup is to take extra care with any simple selectors. Anything that's just a single generic sounding class name (and not [BEM](http://getbem.com/naming/) style) is likely to be a "utility class" that could be used just about anywhere. Those will have to each be searched for. On the other hand, all the complex selectors in this file that use `.table-dark-blue` as their root are really easy to rule out if `table-dark-blue` never shows up (which it doesn't) :+1:

client/src/components/DisplayTable/DisplayTable.scss
```diff
@@ -63,10 +63,6 @@ table.display-table > thead > tr {
 
     color: $textLightColor;
 
-    &.center-text {
-      vertical-align: middle;
-    }
```
Stephen-ONeil, line 68: Oh, sorry, one more weird thing here, didn't see this my self. This specific\* selector against `center-text` seems to be extending the `center-text` style to also center vertically. A little weird, but find if the `&` expands to something specific enough that this only applies here. Should probably leave this, just with `center-text` updated to `text-center`.

\* `&` is scss shorthand for whatever the selector one level up in the nesting is, so this is probably only applies to some specific `DisplayTable` html

Stephen-ONeil (commented):
One thing, I think.

Stephen-ONeil (approved):
:+1: :+1:

PR #1196, opened 2021-07-15
Search services

ktw1016:
So I see 3 alternatives for an individual service route and its searching.
1. Make it as a regular "infograph". It doesn't really fit in structurally. (Service isn't a `level` like dept, program, cr, etc) So naturally it's not gonna share any bubbles except like intro bubble. I guess we could just use the intro bubble and fit everything in there tho.. For searching, add it to config
2. Keep `/service` route, adding services to search configs
3. Keep `/service` route. Make a sub app "Service Inventory", which is basically the `List of provided services` but for all services, flat. Would be easier to add filter functionalities.

I think 2 is the best option which is what's on this branch currently. Search "Service name" on typeahead


client/src/search/search_configs.js
```diff
@@ -352,6 +352,17 @@ const crsos = {
     memoized_re_matchers(query, ["name", "activity_code"], "crsos")(datum),
 };
 
+const services = {
+  config_name: "services",
+  header_function: () => trivial_text_maker("services"),
+  name_function: (service) => service.name,
+  get_data: () => [
```
ktw1016, line 359: Something to think about here..
I removed Service models already.

Should I be bringing that back, and load services on load? Or GraphQL query in here to load services?

Stephen-ONeil: Oh, ouch, forgot about that when I agreed that it'd be easy to let users search by it. Everything else that is searchable from the home page is included in the big blast of additional loading. services.csv is about 2M too (will be roughly 1M if we split by lang), decently heavy addition. For comparison, orgs (igoc_en.csv) are around 2.3K and programs (program_en.csv) are around 1.5M.

Reminder, all those necessary initial CSVs are bundled in to lang specific `lookup_en.json` etc files by the base build script. After compression, the current en lookup in prod is 543K of the home page's (maximum) 1.7M download.  Don't want to make that any heavier, but we also are a waaaays off from moving the whole search to the back end and cutting it down. Try experimenting with how much pre-loading all the services would be (just need to add it to the right part of `build_code/copy_static_assets.json`, maybe with or without making at least an en specific version for testing, can get an idea for how bad the loading is before we go about re-adding a static store subject for it).

There's nowhere that needs to have every single service loaded otherwise right now, yeah? Part of why it's not a big priority to move searching to the backend, to take orgs etc out of the initial load, is because many place will still want all that data loaded at once (igoc is the big example of that, otherwise it's mostly just places that want at least the name of every org and or program loaded). More of a waste to load all the data for every service up front if it's just to search on them. Might get a bit complicated, but maybe we could only put service names and ids in the lookup and lazy-load the rest of the service's fields as needed? Might be worth considering that for the orgs, programs, etc too.

Stephen-ONeil (changes requested):
Tricky, needs more thought/experimenting.

PR #1197, opened 2021-07-19
TS: modals_and_popovers

csong202:
Converting files to typescript

- FixedPopover.stories
- FocusLockedModal
- FocusLockedModal.stories
- ModalButton
- ModalButton.stories
- StatelessModal
- StatelessModal.stories

### DIFF
--- a/client/src/components/modals_and_popovers/FixedPopover.stories.tsx
+++ b/client/src/components/modals_and_popovers/FixedPopover.stories.tsx
@@ -1,4 +1,5 @@
 import { useArgs } from "@storybook/client-api";
+import { Story, Meta } from "@storybook/react";
 import React, { Fragment } from "react";
 
 import { FixedPopover } from "./FixedPopover";
@@ -9,9 +10,11 @@ export default {
 
   // Need decorators to use useArgs()
   decorators: [(Story) => <div>{Story()}</div>],
-};
+} as Meta;
 
-const Template = (args) => {
+type FixedPopoverProps = React.ComponentProps<typeof FixedPopover>;
+
+const Template: Story<FixedPopoverProps> = (args) => {
   const [_, updateArgs] = useArgs();
 
   function on_close_callback() {
--- a/client/src/components/modals_and_popovers/FixedPopover.stories.tsx
+++ b/client/src/components/modals_and_popovers/FixedPopover.stories.tsx
@@ -2,6 +2,8 @@ import { useArgs } from "@storybook/client-api";
 import { Story, Meta } from "@storybook/react";
 import React, { Fragment } from "react";
 
+import { ComponentProps } from "src/types/util_types.d";
+
 import { FixedPopover } from "./FixedPopover";
 
 export default {
@@ -11,12 +13,6 @@ export default {
   // Need decorators to use useArgs()
   decorators: [(Story) => <div>{Story()}</div>],
 } as Meta;
-
-type ComponentProps<T> = T extends
-  | React.ComponentType<infer P>
-  | React.Component<infer P>
-  ? JSX.LibraryManagedAttributes<T, P>
-  : never;
 
 type FixedPopoverProps = ComponentProps<typeof FixedPopover>;
 
--- a/client/src/components/modals_and_popovers/FixedPopover.stories.tsx
+++ b/client/src/components/modals_and_popovers/FixedPopover.stories.tsx
@@ -12,7 +12,13 @@ export default {
   decorators: [(Story) => <div>{Story()}</div>],
 } as Meta;
 
-type FixedPopoverProps = React.ComponentProps<typeof FixedPopover>;
+type ComponentProps<T> = T extends
+  | React.ComponentType<infer P>
+  | React.Component<infer P>
+  ? JSX.LibraryManagedAttributes<T, P>
+  : never;
+
+type FixedPopoverProps = ComponentProps<typeof FixedPopover>;
 
 const Template: Story<FixedPopoverProps> = (args) => {
   const [_, updateArgs] = useArgs();
--- a/client/src/components/modals_and_popovers/FixedPopover.stories.tsx
+++ b/client/src/components/modals_and_popovers/FixedPopover.stories.tsx
@@ -60,7 +60,7 @@ Basic.args = {
 
   // css
   dialog_position: "",
-  additional_dialog_class: "",
+  additional_dialog_class: undefined,
 };
 
 export const Timer = Template.bind({});
@@ -83,7 +83,7 @@ Timer.args = {
 
   // css
   dialog_position: "",
-  additional_dialog_class: "",
+  additional_dialog_class: undefined,
 
   // clock
   auto_close_time: 1000,
--- a/client/src/components/modals_and_popovers/FixedPopover.stories.tsx
+++ b/client/src/components/modals_and_popovers/FixedPopover.stories.tsx
@@ -58,7 +58,7 @@ Basic.args = {
 
   // css
   dialog_position: "",
-  additional_dialog_position: "",
+  additional_dialog_class: "",
 };
 
 export const Timer = Template.bind({});
--- a/client/src/components/modals_and_popovers/ModalButton.tsx
+++ b/client/src/components/modals_and_popovers/ModalButton.tsx
@@ -1,19 +1,19 @@
 import React, { Fragment } from "react";
 
-import { withRouter } from "react-router-dom";
+import { withRouter, RouteComponentProps } from "react-router-dom";
 
 import { SafeJSURL } from "src/general_utils";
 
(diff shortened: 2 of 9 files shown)


client/src/components/modals_and_popovers/FixedPopover.stories.tsx
```diff
@@ -9,9 +10,17 @@ export default {
...
   // Need decorators to use useArgs()
   decorators: [(Story) => <div>{Story()}</div>],
-};
+} as Meta;
+
+type ComponentProps<T> = T extends
+  | React.ComponentType<infer P>
+  | React.Component<infer P>
+  ? JSX.LibraryManagedAttributes<T, P>
+  : never;
```
Stephen-ONeil, line 19: Just a note, #1191 is probably going to declare this utility type centrally in `src/types/utility_Types.d.ts`. Depending which branch merges first, please coordinate with Ivan to switch to using the central definition here.

client/src/components/modals_and_popovers/StatelessModal.tsx
```diff
@@ -6,7 +6,7 @@ import { Modal } from "react-bootstrap";
 import { trivial_text_maker } from "src/models/text";
 import "./bootstrap_modal_exstension.scss";
 
-interface StatelessModalProps {
+type StatelessModalProps = typeof StatelessModal.defaultProps & {
```
Stephen-ONeil, line 9: Looks to have the same `any` issue from [this](https://github.com/TBS-EACPD/infobase/pull/1191#pullrequestreview-710647649) comment.
image

Stephen-ONeil: And oops, this and other files are being updated on both this and Ivan's branch. Guess I could have asked you to pause this branch until his was done.

client/src/components/modals_and_popovers/StatelessModal.tsx
```diff
@@ -15,14 +15,18 @@ interface StatelessModalProps {
...
   close_text?: string;
-  include_close_button_in_header: boolean;
+  include_close_button_in_header?: boolean;
   additional_dialog_class_name?: string;
-}
+};
 
 export class StatelessModal extends React.Component<StatelessModalProps> {
   static defaultProps = {
+    subtitle: "",
```
Stephen-ONeil, line 24: I think because this has a default of `""`, even though it's declared optional in the explicit types, it's going to internally say that it's type is _just_ `string`. That means in practice we're giving the empty string the special meaning of "do not render a subtitle". I think it probably makes sense to not render any subtitle related DOM when the string is empty, but for the sake of the component's interface it's clearer if it explicitly takes `undefined` to mean "render no subtitle". Same on the other optional props here.

So, to be clear, I think they should just be optional string props, and should not be in default props.

client/src/components/modals_and_popovers/ModalButton.tsx
```diff
@@ -1,13 +1,26 @@
...
 
 import { SafeJSURL } from "src/general_utils";
 
 import { StatelessModal } from "./StatelessModal";
 
-class ModalButton_ extends React.Component {
-  constructor(props) {
+type ModalButtonProps = RouteComponentProps & {
+  show_condition: { name: string; value: string };
+  match: { params: { options: string } };
```
Stephen-ONeil, line 11: Would expect `match` typing to come from `RouteComponentProps`?

client/src/components/modals_and_popovers/ModalButton.tsx
```diff
@@ -1,13 +1,26 @@
...
+  button_text: string;
+  title: string;
+  children: string | React.ReactNode;
```
Stephen-ONeil, line 14: `ReactNode` covers strings.

Here's a little pattern I use often to double check my assumptions about types:
image

Stephen-ONeil: That can be read as "can something of type `string` be stored in type `React.ReactNode`? `any` for true, `never` for false".

Stephen-ONeil (changes requested):
Some comments, unfortunately a number of things are tangled up with #1191. If you haven't been, please keep an eye on that PR while these are both WIP.

client/src/components/modals_and_popovers/FixedPopover.stories.tsx
csong202: I'm going to wait for Ivan's branch to merge first then I'll rebase off of master and adjust mine since he added the utility type centrally.

client/src/components/modals_and_popovers/StatelessModal.tsx
csong202: I'll drop the commits on this file

Stephen-ONeil (changes requested):
Some inconsistency between `additional_dialog_class` and `additional_dialog_class_name`. Didn't check if that's a typo or inconsistency in the original components, but it'd be good to standardize on one or the other (the first  is shorter and clear enough imo).

Second minor thought, I would think `undefined` would be the better value for the "no class name provided" case, since that'd be the actual use-case version of not providing that prop (assuming it was typed as optional in the component). Would have to double check how that works with the storybook controls section, since I'm assuming that's why it's there with empty strings in the first case.

csong202:
> Second minor thought, I would think `undefined` would be the better value for the "no class name provided" case, since that'd be the actual use-case version of not providing that prop (assuming it was typed as optional in the component). Would have to double check how that works with the storybook controls section, since I'm assuming that's why it's there with empty strings in the first case. They're on the storybook controls with empty strings because they're in the args. Should I get rid of them?

`additional_dialog_class` is optional on StatelessModal and FixedPopover but required on FocusLockedModal (has type `string | string[]`).

Stephen-ONeil:
> `additional_dialog_class` is optional on StatelessModal and FixedPopover but required on FocusLockedModal (has type `string | string[]`). They're on the storybook controls because the args have `additional_dialog_class: ""`. Should I get rid of them?

I'm guessing it should be optional on `FocusLockedModal` too. Didn't look, but it's called "additional" which implies optional imo. If it looks like it really isn't optional, maybe rename the prop to just `dialog_class` on `FocusLockedModal`. Will cause less confusion for why the same prop is optional sometimes and required others (but again, I'm guessing it should always be optional).

So it doesn't show up in the controls with `additional_dialog_class: undefined`? If not, it's fine to leave them as empty strings, I was just curious if `undefined` worked (since it's the "natural" value for an optional prop).

csong202:
> So it doesn't show up in the controls with `additional_dialog_class: undefined`? If not, it's fine to leave them as empty strings, I was just curious if `undefined` worked (since it's the "natural" value for an optional prop).

Nvm it still shows up with undefined and I get a `Set string` button
Screen Shot 2021-07-27 at 9 56 35 AM

client/src/components/modals_and_popovers/ModalButton.tsx
Stephen-ONeil: > Would expect `match` typing to come from `RouteComponentProps`?

Still curious about this. Is `match: { params: { options: string } };` redundant to `RouteComponentProps`?

Stephen-ONeil (changes requested):
Sorry, still think the `additional_dialog_class` prop on `FocusLockedModal` should have just been made optional. Took a look, and the component is currently only used in the report builder, where that prop isn't being used.

In the component, `classNames(`modal-dialog`, dialog_class)` is still an optional use of the prop. The behaviour of the `classNames` util is to concat string and string[] args together _and_ to drop any falsey values (it's mainly a short hand for filtering out conditional/optional classes). So, a value of `undefined` in the default case is valid there.

csong202:
> Still curious about this. Is match: { params: { options: string } }; redundant to RouteComponentProps?

When I leave out `RouteComponentProps & {...` I get this error:
```
Argument of type 'typeof ModalButton_' is not assignable to parameter of type 'ComponentType<RouteComponentProps<any, StaticContext, unknown>>'.
  Type 'typeof ModalButton_' is not assignable to type 'ComponentClass<RouteComponentProps<any, StaticContext, unknown>, any>'.
    Types of parameters 'props' and 'props' are incompatible.
      Type 'RouteComponentProps<any, StaticContext, unknown>' is missing the following properties from type 'ModalButtonProps': show_condition, button_text, title, children, aria_label
```

Stephen-ONeil:
Sorry, to clarify I meant the `match: { params: { options: string } };` typing seemed redundant. `match.params` is part of the router API added by the `withRouter` wrapper, so I assumed the `RouteComponentProps` handled the typing for it.

Took a look, I believe the proper usage is `RouteComponentProps<{ options: string }> & ...`. `RouteComponentProps` adds the typing for the props added by `withRouter`, and is generic on the type for the expected route params.

Stephen-ONeil (approved):
:+1:

PR #1198, opened 2021-07-20
Span-Hover to Checkbox.scss

patfire12345:


client/src/components/CheckBox/CheckBox.scss
```diff
@@ -9,3 +9,12 @@
   flex-shrink: 0;
   flex-grow: 0;
 }
+
+// in CheckBox.tsx (1)
+.span-hover {
```
Stephen-ONeil, line 14: To match the existing class names used in `CheckBox`, and use BEM, I'd suggest renaming this class as `checkbox-span--interactive`. Should avoid generic names outside of the global style sheet modules (to avoid accidentally reusing a name in multiple places).

Stephen-ONeil (changes requested):
Are the comments just to help you track things while cleaning up? I'd prefer not to merge them in master, will want to delete them when this process is done, since there'll be nothing to keep them up to date in the future. "Lying" comments are way worse than no comments.

client/src/components/CheckBox/CheckBox.scss
patfire12345: Done! Also I apologize for the comments, those were just for tracking purposes, I'll delete them!

Stephen-ONeil (approved):
:+1:

client/src/components/CheckBox/CheckBox.scss
Stephen-ONeil: No worries!

PR #1199, opened 2021-07-21
Moved one-off scss into Panel.scss

patfire12345:


client/src/components/Panel/Panel.scss
```diff
@@ -108,3 +108,12 @@ button.panel-heading-utils {
   margin: 20px 60px 20px 60px;
   border-top: 1px solid rgba($separatorColor, 0.5);
 }
+
+.panel-source-link {
+  font-weight: bold;
```
Stephen-ONeil, line 113: Should use the centrally defined variables when setting font weights, `$heavyFontWeight` etc.

Stephen-ONeil (changes requested):
Since both classes only set bold, I'd recommend deleting them and just adding the `bold` class to both.

client/src/components/Panel/Panel.scss
patfire12345: Done!

Stephen-ONeil (changes requested):
To clarify, there's a bootstrap utility class named `bold` (we override it with our scss variable for heavy font weight somewhere). I was suggesting to just use that.

client/src/components/Panel/Panel.scss
```diff
@@ -1,4 +1,5 @@
 @import "src/common_css/common-variables";
+@import "src/extended_bootstrap_css/bootstrap-wet-fixes-extensions";
```
Stephen-ONeil, line 2: Not necessary. The `extended_bootstrap_css` is loaded globally early in the sites startup.

Not recommended either. If a JS module needed to guarantee multiple specific style sheets were loaded we'd have the imports inside the JS. That way webpack can set up common chunks and avoid downloading the styles multiple times. If you `@import` the styles directly in the stylesheet, the output CSS will have those styles inlined (and webpack won't know, so duplication will likely result). 

The main use of `@import` in our code base would be importing SCSS utilities like the variables declared in `common-variables` or SCSS utilities from the `Math` name space, etc. Those sorts of things can be thought of as being transformed away in the output CSS, just being replaced with the equivalent value.

Stephen-ONeil (changes requested):

client/src/components/Panel/Panel.scss
patfire12345: Thanks! Removed the import now.

Stephen-ONeil (approved):
:+1:

PR #1202, opened 2021-07-22
Axe warning fixes

sranc046:
Cleaning up some of the warnings axe is finding during the tests.

Lots of interesting information to be learned here: https://dequeuniversity.com/.


client/src/explorer_common/explorer_components.js
```diff
@@ -49,7 +49,6 @@ export const ExplorerHeader = ({
             }
             onClick={() => col_click(id)}
             tabIndex={0}
-            role="button"
```
sranc046, line 52: Basically not great practice to have buttons nested within buttons as it can cause problems with screen readers (https://dequeuniversity.com/rules/axe/4.2/nested-interactive). By removing the `role="button"` the header of this table is still clickable and will toggle the sorting but will no longer have the "hand click" mouse icon that you see when you hover other clickable elements. I feel like this comes back to the original issue with these toggling arrows that don't actually have the functionality within them?

sranc046:
There are 2 other issues that affect most of the tests:

1. Our $tertiaryColor is not suitable for smaller fonts like in this picture:
image
With an okay I can add a slightly darker tone to be used in these specific situations.

2. Our infographic routes use `role=main` nested deeper within the html `<main>` element. There should only be one use of main (being that the `<main>` has `role=main`).  Didn't touch this one yet because there might have been some reason it was implemented this way? `<main>` currently contains everything which is not the header (like the breadcrumbs) and the footer whereas the div with `role=main` contains everything below the search and infographic bubbles.

client/src/explorer_common/explorer_components.js
Stephen-ONeil: Original issue?

For the cursor: https://developer.mozilla.org/en-US/docs/Web/CSS/cursor

client/src/components/Typeahead/Typeahead.tsx
```diff
@@ -165,13 +165,16 @@ export class Typeahead extends React.Component<TypeaheadProps, TypeaheadState> {
...
             autoComplete="off"
             value={input_value}
+            type="search"
             onFocus={this.handle_input_focus}
             onChange={this.handle_input_change}
             onKeyDown={this.handle_key_down}
             role="combobox"
             aria-autocomplete="none"
             aria-owns={this.menu_id}
+            aria-label={"search"}
```
Stephen-ONeil, line 175: Needs to be bilingual.

client/src/components/Typeahead/Typeahead.tsx
```diff
@@ -165,13 +165,16 @@ export class Typeahead extends React.Component<TypeaheadProps, TypeaheadState> {
             placeholder={placeholder}
             autoComplete="off"
             value={input_value}
+            type="search"
```
Stephen-ONeil, line 168: Not 100% on this, but I think it should be `type="text"` since it has `role="combobox"`. Have you tried these changes with Mac Voice Over? Problem's always that screen readers can be very inconsistent so it's better to stick as close to the spec/best practices as possible, rather than "whatever works best with Voice Over", but it might help clarify something like this.

Stephen-ONeil (changes requested):
Coupling this to the cypress branch will hold up progress. Maybe you could use the output from the axe tests on the cypress branch to find problem areas, but fix them on a clean branch off master. The browser plugin will help you confirm the fixes as you make them.

Stephen-ONeil:
> 1....

Hmm, oops. How much darker does it need to be to pass in all cases? If it isn't too drastic, changing the colour in all cases would be easier for future maintenance. 

> 2. ...

Could be wrong, but the general infographic page html has got to be pretty ancient. Most likely we didn't have axe/know that much about a11y when we wrote it, so I doubt there's a _good_ reason, if any, to leave that `role="main"` in. Could try to git blame it if you want to dig through the history, might be a commit message about it or something. Likely the file's moved/been renamed a few times in a way that loses it's history though, depending how far back you'd have to go to find this.

client/src/explorer_common/explorer_components.js
sranc046: Original issue that had been brought up when Christina was working on something in the report builder (maybe storybooking those arrows?). Rather than the individual arrows being a "proper" button, they act as a toggle for each other. For example: if you click on the left upwards arrow, it won't activate that specific filter but it will toggle the filter (up or down). So you could keep clicking on the left upwards arrow and keep switching back and forth between filtering upwards or downwards. Especially confusing when starting off with no filter and then clicking on the left upwards arrow but the right downwards one activates.

sranc046:
> How much darker does it need to be to pass in all cases? If it isn't too drastic, changing the colour in all cases would be easier for future maintenance.

https://dequeuniversity.com/color-contrast you can check out the difference in contrast here. Our foreground colour is #8c949e and background colour is #fff. The foreground (tertiary colour) would need to be shifted over to around the 49% mark.

client/src/explorer_common/explorer_components.js
Stephen-ONeil: Ah, makes sense. Don't see any open issues for that, make one if you want.

Stephen-ONeil:
Significantly darker, yeah. Would have to double check there's nowhere using `tertiaryColor` _as_ a background colour for dark text before changing it that much, haha.

Worth noting, `tertiaryColor` is explicitly noted in `_common-variables.scss` to be combined with scss utilities to vary it's alpha and darkness values depending on use case. Could double check how often this colour is used for text, if it's just one off than you could darken it there. Alternatively (probably preferable, bit more work), add a new central `textGrey` variable with a sufficiently contrasting value, switch grey text to that.

client/src/components/HeightClipper/HeightClipper.tsx
```diff
@@ -37,7 +40,26 @@ export class HeightClipper extends React.Component<
     };
     this.main = React.createRef();
     this.content = React.createRef();
+
+    this.debounced_mutation_callback = _.debounce(
+      (mutationList: MutationRecord[], observer: MutationObserver) => {
+        this.componentDidUpdate();
```
sranc046, line 46: Will look into trimming this down rather than calling the whole componentDidUpdate()

client/src/components/HeightClipper/HeightClipper.tsx
```diff
@@ -37,7 +40,26 @@ export class HeightClipper extends React.Component<
...
+      }
+    );
+
+    this.observer = new MutationObserver(this.debounced_mutation_callback);
+
+    this.observer.observe(document, {
```
sranc046, line 52: Using `document` as target node for now since calling a query selector returns a possible null type (which typescript won't accept)

Stephen-ONeil (changes requested):
To make this easier to review, and avoid a bunch of conflicts with the actual CI branch post-merge, please use an interactive rebase off master to drop the cypress commits directly.

client/src/components/modals_and_popovers/ModalButton.tsx
```diff
@@ -52,6 +52,7 @@ class ModalButton_ extends React.Component<ModalButtonProps, ModalButtonState> {
           className="btn btn-link"
           onClick={this.toggle_modal}
           aria-label={aria_label}
+          style={{ color: "#0275d8" }}
```
ktw1016, line 55: what's this color? Can it use a color from one of our defined colors?

client/src/components/HeightClipper/HeightClipper.tsx
```diff
@@ -37,7 +40,24 @@ export class HeightClipper extends React.Component<
     };
     this.main = React.createRef();
     this.content = React.createRef();
+
+    this.debounced_mutation_callback = _.debounce(() => {
```
ktw1016, line 44: Why did you need to setup mutation observer for debouncing componentDidUpdate?

And also, is this an equivalent of `_.debounce(this.componentDidUpdate)`

sranc046: Sorry forgot to answer this... The mutation observer is needed since tagging "untabbable" elements wasn't always happening because of timing issues. Now with the mutation observer, componentDidUpdate will look through the elements again to tag the missing hidden elements.

ktw1016:
Were you running Axe manually or is it cypress running axe? If it's cypress is there CI step running axe? I just couldn't find it

.circleci/config.yml
```diff
@@ -124,7 +124,7 @@ jobs:
 
       - restore_cache:
           keys:
-            - client-dependencies-{{ checksum "client/package-lock.json" }}-{{ checksum "client-patch-checksums.txt" }}-v2
+            - client-dependencies-{{ checksum "client/package-lock.json" }}-{{ checksum "client-patch-checksums.txt" }}-v3
```
Stephen-ONeil, line 127: Can you remind me why this needed to roll all the keys forward? That still necessary?

sranc046: Leftovers when trying to debug failing CI tests

client/src/components/Tombstones/Tombstones.tsx
```diff
@@ -24,17 +24,19 @@ interface LabeledTombstoneProps {
...
+          <dt className="col-12 col-lg-2">{label}</dt>
+          <dd className="col-12 col-lg-10">{item}</dd>
+        </dl>
         {ix !== _.size(labels_and_items) - 1 && (
           <hr style={{ width: "100%" }} />
         )}
       </Fragment>
     ))}
-  </dl>
+  </div>
```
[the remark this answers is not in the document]
Stephen-ONeil: No screen reader should be announcing an `hr` tag, as far as I know. I expect this was a false-positive. Does leaving it as-was and marking the `hr` as `aria-hidden` silence the alert?

I think the screen reader experience for individual `dl`s per `dt`/`dd` pair will be much different/worse than a single `dl`. Double check that please. If so, and if slapping `aria-hidden` on the `hr` isn't enough, you might need to recreate the horizontal line as pure CSS (it'll be a bother, take a couple media queries and potentially create a brittle sync between the column classes on these elements and the component's style sheet).

client/src/charts/canada/canada.js
```diff
@@ -221,6 +221,7 @@ export class Canada extends React.Component {
                 onSelect={this.year_select_callbback}
                 title={text_maker("select_year")}
                 className={"bold"}
+                ariaLabel={text_maker("selector")}
```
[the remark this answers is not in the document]
Stephen-ONeil: Any answers to my questions from the first paragraph?

client/src/components/Tombstones/Tombstones.tsx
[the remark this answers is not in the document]
sranc046: You're right about the screen reader not announcing any `hr` tag. When I had tested the difference between the two implementations, the only difference was that having the paired up elements allowed for selecting and reading the full row rather than just either the `dt` or `dd` tag.

client/src/components/HeightClipper/HeightClipper.tsx
```diff
@@ -37,10 +42,39 @@ export class HeightClipper extends React.Component<
     };
     this.main = React.createRef();
     this.content = React.createRef();
+
+    this.debounced_mutation_callback = _.debounce(
+      (mutationList: MutationRecord[], observer: MutationObserver) => {
+        const height_clipper_node = ReactDOM.findDOMNode(this) as HTMLElement;
+        const untabbable_children_node = height_clipper_node.querySelector(
+          ".untabbable_children"
+        );
```
Stephen-ONeil, line 51: Could just refs for this (thought that's what the existing ones were for, but haven't looked at the rest of this component too closely).

client/src/components/Select/Select.tsx
```diff
@@ -16,6 +16,7 @@ interface SelectProps {
   disabled?: boolean;
   style?: React.CSSProperties;
   title?: string;
+  ariaLabel?: string;
```
Stephen-ONeil, line 19: No longer used.

Stephen-ONeil (changes requested):

client/src/components/HeightClipper/HeightClipper.tsx
sranc046: I might've misunderstood your comment, apologies if the new push doesn't address this comment.

client/src/components/HeightClipper/HeightClipper.tsx
```diff
@@ -37,10 +40,37 @@ export class HeightClipper extends React.Component<
...
+
+    this.debounced_mutation_callback = _.debounce(() => {
+      const height_clipper_node = ReactDOM.findDOMNode(this) as HTMLElement;
+      const untabbable_children_node = height_clipper_node.querySelector(
+        ".untabbable_children"
+      );
+
+      if (untabbable_children_node) {
+        this.setFocusableFalse(untabbable_children_node);
+      }
```
Stephen-ONeil, line 52: Close! Not technically wrong, but I believe `findDOMNode` is considered shaky (looks like it's deprecated in React's strict mode, they'd probably fully deprecate it if it wouldn't be a huge breaking change). Refs are the preferred approach.

Set up a `this.untabbable_children = React.createRef();` in the constructor, set `ref={this.untabbable_children}` on the `.untabbable_children` node, and then you can check for/reference the corresponding DOM node directly with `this.untabbable_children.current`. This a few other places in this file that look to use some outdated practices (guessing they're what influenced this), please clean them up while you're in there.

Stephen-ONeil (changes requested):

client/src/components/HeightClipper/HeightClipper.tsx
Stephen-ONeil: No worries. Feel free to ask for clarification any time (or take a stab at it and put things back up for review, up to you case-by-case).

Stephen-ONeil (changes requested):
One last minor clean up tip, and I've still got questions about [the dl stuff](https://github.com/TBS-EACPD/infobase/pull/1202#discussion_r713292831). Almost done though!

client/src/components/Tombstones/Tombstones.tsx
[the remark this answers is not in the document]
sranc046: Oh sorry, hadn't seen this... two implementations being the one that was already in the code and then my solution. My solution allowed for selecting the full rows.

client/src/components/HeightClipper/HeightClipper.tsx
```diff
@@ -37,44 +40,49 @@ export class HeightClipper extends React.Component<
     };
     this.main = React.createRef();
     this.content = React.createRef();
+    this.untabbable_children = React.createRef();
```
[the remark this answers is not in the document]
sranc046: Not sure what you mean, where would I declare them?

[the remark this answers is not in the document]
Stephen-ONeil: Anywhere in the class body technically, although I'm specifically suggesting replacing the current type declarations. E.g. can replace `main: React.RefObject<HTMLDivElement>;` with `main = React.createRef<HTMLDivElement>();`, up around lines 29-33.

client/src/components/Tombstones/Tombstones.tsx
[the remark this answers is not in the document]
Stephen-ONeil: Hm, unfortunately every screen reader probably handles that differently. As much as possible, we need to respect the semantics of a tag as described in the spec, since it'd be impractical to test with (and satisfy all) different screen reader implementations. Doing something iffy because it works better for Mac voice over probably makes things worse for other, unknown, screen readers :shrug:.

This is iffy because the `dl` is explicitly a (description) list. X number of sibling, single element, lists is semantically different from one list with X elements.

[Here's something helpful](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl#wrapping_name-value_groups_in_div_elements): per the spec, the `dt`/`dd` pairs can be wrapped in `div`s. That'll make replacing the `hr` with a bottom border style much more trivial.

Side note, that mozilla doc page [also mentions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl#accessibility_concerns) that every screen reader is going to express them in their own special ways, haha.

Stephen-ONeil (approved):
Know it's not got the awaiting review label but it's good to me so I'm approving now (reminder, I'm off Friday)!

PR #1203, opened 2021-07-27
Remove Date as valid CellValue type on DisplayTable

Stephen-ONeil:
Address a bunch of other TODO and cleanup comments in DisplayTable while I'm at it.

Was always a messy hack that the `DisplayTable` had to handle `Date` objects for the sake of one use case, never wanted that. Adding TS was only going to make this worse as all custom sort functions etc would have to explicitly do something about potentially receiving a `Date` type. Still have to do something about a given column config function prop not knowing whether it's corresponding data is either numbers or strings, but at least those can be easily type-narrowed on. 

TODO:
  - [x] remove `Date` from `CellValue` type, related cleanup that allows in the `DisplayTable` code
  - [x] big cleanup and standardization of the `DisplayTable` sort funcs
    - [x] first pass
    - [x] important to double check any collumns that supplied a raw formatter but no sort func. The default sort func used to sort of "raw" formatted values (non-default sort funcs did not). Removed that in cleanup, so may have broken some assumptions in some `DisplayTable` use cases
    - [x] part of my cleanup was standardizing on passing the `descending` boolean in to all sort funcs and removing the data reversing step from the `DisplayTable` itself. Remembered afterwards _why_ that boolean existed (one special case that no longer exists I believe). Reconsider whether the sort funcs should have to/be able to implement the reverse sort behaviour, or whether that should be `DisplayTable`'s job (leaning towards the former, might always have another special case some day anyway) 
  - [x] remove use of `Date` type values in `results_flat_table.js`
  - [x] feel like there's room for a bit more cleanup still (e.g. sort utils might be better as a top level util, rather than a DisplayTable util exported out through the components index)


client/src/components/DisplayTable/DisplayTable.tsx
```diff
@@ -76,17 +77,23 @@ const get_column_config_defaults = (index: number) => ({
...
   sum_func: (sum: number, value: number) => sum + value,
-  // "raw" value for some complex datasets actually requires processing, as we "raw" values are used for sorting, searching, AND the output csv string
-  raw_formatter: _.identity as (val: CellValue) => string,
   sum_initial_value: 0,
   visibility_toggleable: index !== 0,
+  // "plain" value for some complex datasets actually requires processing, as we use "plain" formatted values in the output csv AND for searching and (by default) sorting
+  plain_formatter: _.identity as (value: CellValue) => string | number,
+  sort_func: (
+    plain_a: string | number,
+    plain_b: string | number,
```
Stephen-ONeil, line 86: Specifically not using `CellValue` type for these, on the off chance that `CellValue` gets more complicated again in the future. Want to be clear that "plain" should be simple string/number values.

ktw1016 (approved):
👍👍

PR #1206, opened 2021-07-28
RPB Group By

csong202:
Issue: #889

- deleted dimensions.js
- simplified table dimensions and deleted filters
- created a group by util that replaces make_horizontal_func
  - has 3 arguments:
    - col: String[] or String: column(s) to sum the grouped data by
    - dimension: String: dimension (or column) to group the data by
    - subject (optional, default = Gov): which subject entities data that you want included in the grouped and summed data 
- added instance variables and functions for handling dimensions with non-default behaviour:
  - all_special_dimensions: dimensions from all tables that do not use the default grouping
  - special_dims: special dimensions of current table
  - get_special_dim_value(row, dimension): gets value for row[dimension]
  - set_special_group_by(dimension): sets the table's special grouping function for dimension
- adjusted `DisplayTable` sorting
  - table rows are initially sorted (descending) by the first column

Infographic levels to fix that were affected by deleting make_horizontal_func():
- [x] About
- [x] Finance
  - [x] gov
  - [x] dept  
- [x] People
  - [x] gov


client/src/tables/orgEmployeeAgeGroup.js
```diff
@@ -100,18 +100,7 @@ export default {
...
-      include_in_report_builder: true,
-
-      filter_func: function (options) {
-        return function (row) {
-          return row.age;
-        };
-      },
-    },
-  ],
+  dimensions: ["dept", "age"],
```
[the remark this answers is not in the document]
Stephen-ONeil: Could also move the special case vote/stat roll up code out of the RPB by giving the columns an optional `group_by_function` property defined here. The RPB would only need to "know" that it's _possible_ for columns to declare that function and how to apply it if it exists. Much better for maintainability; shouldn't need to double check and possibly change RPB logic every time a table is changed/added.

ktw1016 (commented):
Works well, I didn't find much red flags. Good work!!

Stephen-ONeil (changes requested):
> Could also move the special case vote/stat roll up code out of the RPB by giving the columns an optional `group_by_function` property defined here. The RPB would only need to "know" that it's _possible_ for columns to declare that function and how to apply it if it exists. Much better for maintainability; shouldn't need to double check and possibly change RPB logic every time a table is changed/created.

Sticking by this. Neither the `Table` class nor RPB should know about the special vote stat column cases. The RPB should just know how to check if a given column can be grouped by, in which case it calls the standard grouping function on it. The `Table` class should own the default group by logic, for columns with `can_group_by: true` (throw an error if called with a column that can't explicitly be grouped by), but should also also check the individual column definition for an optional `group_by_function` to be used when non-default logic is necessary.

The `Table` class is an abstraction around the individual datasets. A good abstraction presents a consistent interface and hides special cases from the consumers (RPB, panels, etc). A _great_ abstraction avoids even explicitly knowing about special cases itself, providing configuration options that instances can use to smooth over their own quirks.

Consider what would happen if a new special vote stat category was added to one of the tables with a vote stat column, or if they ended up diverging for some reason. In the current implementation, a developer would have to know about (or be lucky to find) code scattered over at least three files, and would have to likely add even more code in each of those locations. By pushing all the special case handling down in to the configuration level of the tables themselves, the rest of the code base is isolated from any future changes.

Let me know if you have any questions, or if you think I'm missing something that makes my suggested approach impractical!

client/src/components/DisplayTable/DisplayTable.tsx
```diff
@@ -164,10 +164,9 @@ export class _DisplayTable extends React.Component<
     if (nextProps !== prevState.initial_props) {
       const new_default_state = get_default_state_from_props(nextProps);
 
-      const prev_sort_still_valid = _.includes(
-        new_default_state.visible_col_keys,
-        prevState.sort_by
-      );
+      const prev_sort_still_valid =
+        !nextProps.unsorted_initial &&
+        _.includes(new_default_state.visible_col_keys, prevState.sort_by);
```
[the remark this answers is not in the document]
csong202: The `!nextProps.unsorted_initial` check is to make `sort_by` null when you switch from grouping by one of the dimensions to grouping by `all`. Previously if I switched from grouping by Executive Level to grouping by `all` the Executive Level column would still be sorted. But now that I think about it adding the `!nextProps.unsorted_initial` check always makes`prev_sort_still_valid` false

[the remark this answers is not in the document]
Stephen-ONeil: Should changing grouping effect sorting? I think sorting controls and grouping controls being disjoint would be cleaner for the code and have less potentially unexpected behaviour for the users. They might even want to sort by one of the value columns the grouping will be summing up, not the grouped by column itself, so can't say for sure that switching sorting on grouping is actually saving them any clicks, I think.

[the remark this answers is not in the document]
csong202: Thing is, if grouping doesn't affect sorting then some of the tables will have weird default orders. In Executive Level the Non-Ex row ends up be between the EX 04 and EX 05 rows. The Voted and Statutory Item rows will be in different orders on different tables. Geographic Region and Transfer Payments by Recipient Region will order the rows differently when grouped by region.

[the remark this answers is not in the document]
ktw1016: okay how about setting `unsorted_initial=false`, then even after grouping, it'll still be sorted by whatever it's grouped by (e.g. grouping by EX level), not sure what you meant in voted and stat and grouping by region but seems fine to me

If a user sorted by a specific column, then how it's sorted is on user

I'm just trying to remember if there was a specific reason why we had it `false`, but seems fine enough

ktw1016 (commented):
Pretty minor comments overall !

client/src/components/DisplayTable/DisplayTable.tsx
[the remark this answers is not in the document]
csong202: This is with`unsorted_initial=true`
II don't know if this would be considered a problem, but Geographic Region and Transfer Payments by Recipient Region list the regions in a different order when they're both grouped by region. I think the Report Builder is just more cohesive when rows are listed in the same order between different tables

Geographic Region:
Screen Shot 2021-09-25 at 4 14 56 PM

Transfer Payments by Recipient Region:
Screen Shot 2021-09-25 at 4 15 20 PM

ktw1016 (approved):
LGTM🙂

Stephen-ONeil:
In some meetings, but doing a quick review myself. I've got at least one comment (just finishing writing it up), please wait till I've got that in before going ahead!

client/src/core/TableClass.js
```diff
@@ -218,6 +225,10 @@ export class Table extends mix().with(staticStoreMixin) {
         return h.nick || h.wcag;
       })
       .value();
+
+    this.get_dimensions();
+    this.get_group_by_func();
+    this.get_dimension_col_values_func();
```
Stephen-ONeil, line 231: Not going to blame you or necessarily ask for this to be cleaned up right now, because you're just matching how the super legacy `TableClass` was written, just want to comment that this is not a good/modern way to deal with JS class construction. I believe this was likely mostly written pre-ES6, before the JS spec actually had real class support.

A lot of the properties set by these calls could be defined as keyword `get` getters. If they were actually expensive to compute, they could use memoization too. Otherwise, no point to do all this constructor-time calculation and indirect extension of the class. Just makes the whole file harder to read through.

Stephen-ONeil (changes requested):
All this special case dimension stuff in `TableClass.js` is an excellent chance for some good old [inversion of control](https://en.wikipedia.org/wiki/Inversion_of_control)! 

Everything else in this feedback is motivation, here's the concrete suggestion: 
  - remove all the special dimension stuff from `TableClass`
  - instead, specifically in `group_by_func`, check if the relevant column definition includes some function-valued property like `custom_group_by`. Otherwise use the default logic

Want to keep code modular, as a result you never want hard coded knowledge about *specific* individual instances to exist in the class definition\*. Avoiding that keeps things more flexible, easier to maintain, and leaves less room for future bugs. 

\* it's fine for the class to know about *categories* of instances, as long as there's a clearly defined agreement in the class' API for how an instance belongs to a given category

Think of it as the `TableClass` knowing:
> a special case **does** exists for some column called "vote_vs_stat" and **this** is how to handle it
v.s.
> special cases **can** exist, and when they do they will **tell me** how to handle them

The first phrasing is a much more rigid statement, and can be made false in multiple ways (vote_vs_stat stops being a special case, the rules for how to handle it change, other special cases are added). In each of those cases the `TableClass` code itself needs to be changed, which has the potential to impact all the instances of it. 

The second phrasing pushes all of the responsibility down to the instances. If something goes wrong, it will be an isolated bug for a single column of a single instance. Easier to identify and easier to fix (in particular, don't need to check all other instances to confirm no new bugs were introduced).

Stephen-ONeil:
Possible I didn't look too closely, but I assumed that the special case group by effectively over-wrote the group by option/"dimension" for the column. If it's something more like a certain column wanting to provide an _additional_ dimension along side the default, then that'd be a slight tweak to my suggested API.

csong202:
> Possible I didn't look too closely, but I assumed that the special case group by effectively over-wrote the group by option/"dimension" for the column. If it's something more like a certain column wanting to provide an _additional_ dimension along side the default, then that'd be a slight tweak to my suggested API.

The special dimensions provide an additional dimension along the default. The only place where this happens right now is the `type` column in `orgTransferPayments`.  With the implementation I have at the moment it hasn't caused any problems

client/src/core/TableClass.js
csong202: Used a getter for `dimensions` but leaving the `this.get_group_by_func()` and `this.get_dimension_col_values_func()` for now.

Stephen-ONeil:
> > ...
> 
> The special dimensions provide an additional dimension along the default. The only place where this happens right now is the `type` column in `orgTransferPayments`. With the implementation I have at the moment it hasn't caused any problems

The design decisions I'm suggesting are about preventing/limiting future problems. Please check the motivation portion of my suggestion again, let me know if you want me to elaborate after that and we can get on a call some time!

Ok though, that complicates my suggestion slightly, but it's fine. How about allowing for index columns in table definitions to have an optional `groupings` property (leaning towards renaming "dimensions" to something like "groupings", for more descriptive/plainer language), working like this:
  - `undefined` (default): the table class provides a dimension with the column's name, using the default of directly grouping by actual values in the corresponding column
  - Empty array: no groupings for this column
  - something like `{name: string, group_by_func: (table_data: Record<string, any>[]) => Record<string, any>[]}[]`: the table class provides dimensions with the provided name that apply the provided `group_by_func`. If a column wants a custom grouping option alongside the default, it just has to also redeclare the default group by
  
That's the most direct route to the overall idea I was suggesting, of pushing down the responsibility for handling special cases in to the table definitions themselves. There's definitely ways that suggestion could be improved/made cleaner, I just wanted to demonstrate _some_ possible implementation to get the ball rolling.

client/src/rpb/granular_view.js
```diff
@@ -100,70 +111,70 @@ class GranularView extends React.Component {
...
+                htmlFor={dim}
+                className={"normal-radio-btn-label"}
+                key={`${dim}-radio-btn-label`}
+              >
+                {dim === "all" || !_.find(sorted_key_columns, ["nick", dim])
+                  ? text_maker(dim)
+                  : dim === "dept"
+                  ? text_maker("org")
+                  : _.find(sorted_key_columns, ["nick", dim]).header[lang] ||
+                    _.find(sorted_key_columns, ["nick", dim]).header["en"]}
```
ktw1016, line 175: is the "en" supposed to be here?

client/src/rpb/index.js
```diff
@@ -189,9 +173,8 @@ class RPB extends React.Component {
 
     const columns =
       !_.isEmpty(all_data_columns) &&
-      _.filter(all_data_columns, ({ nick }) =>
-        _.includes(this.state.columns, nick)
-      );
+      _.filter(all_data_columns, ({ nick }) => _.includes(data_columns, nick));
+    // const columns = all_data_columns;
```
ktw1016, line 177: comment hanging

ktw1016 (approved):
Didn't look too deep into it but looks pretty good!

client/src/rpb/granular_view.js
sranc046: Doubling down on this, what is the use of the last `||` condition?

sranc046:
I think the group by menu could use a CSS touch up, would be nice to match the style of the new filter dropdown Taewan made (rounded corners, drop shadow, etc...)

Screen Shot 2021-11-12 at 11 57 15 PM

sranc046:
> I think the group by menu could use a CSS touch up, would be nice to match the style of the new filter dropdown Taewan made (rounded corners, drop shadow, etc...)
>
Actually I think Taewan's PR might already address the styling for all these types of dropdowns, my mistake, hopefully no conflicts

client/src/core/TableClass.js
```diff
@@ -68,6 +65,20 @@ function add_child(x) {
   return this;
 }
 
+export function filter_row_by_subj(row, subject) {
+  switch (subject.subject_type) {
```
Stephen-ONeil, line 69: Should have a default case, even if it's to throw an error about an unexpected subject. Will be more useful to future devs than this throwing undefined and that causing a generic error further down the line (or worse, not throwing and becoming a run-time bug :scream:)

client/src/core/TableClass.js
```diff
@@ -241,6 +251,112 @@ export class Table {
...
+  }
+
+  is_custom_dim(dimension) {
+    return (
+      !_.chain(this._cols)
+        .flatMap((col) => (_.has(col, "children") ? col.children : col))
+        .map("nick")
+        .includes(dimension)
+        .value() && dimension != "all"
+    );
```
Stephen-ONeil, line 282: Should be `!==` not `!=`.

Also, not massively important, but a minor optimization point. Putting the cheaper checks first is a free win, i.e. should short circuit and not perform the lodash chain here if `dimension !== "all"`.

client/src/core/TableClass.js
```diff
@@ -241,6 +251,112 @@ export class Table {
...
+
+      if (!this.is_custom_dim(dimension)) {
+        return [dimension, row[dimension]];
+      } else {
+        return dimension_col.custom_groupings[dimension].dim_col_value(row);
+      }
+    };
+  }
+
+  // TODO: come up with a shorter, better name for this sum_cols_by_grouped_data
```
Stephen-ONeil, line 327: Seems like a fine name to me. Short only matters if you have to type it _a lot_. Otherwise it's good as long as it's descriptive enough.

client/src/core/TableClass.js
```diff
@@ -432,7 +548,7 @@ export class Table {
         const [key, val] = pair;
         const type = this.col_from_nick(key).type;
         // in case we have numbers represented as string, we'll convert them to integers
-        // need to handle special cases of '.', '', and '-'
+        // need to handle custom cases of '.', '', and '-'
```
Stephen-ONeil, line 551: Guessing this was from a find and replace, haha

Stephen-ONeil (changes requested):
I've got some comments and questions. I'll also want to go through the RPB behaviour and spot check some panels before the final approval, but I'll do that a bit later.

client/src/rpb/granular_view.js
csong202: Some of the `sorted_key_columns` only have a "en" header and some have "en" and "fr" headers

For example on Expenditures by Standard Object:
Screen Shot 2021-11-15 at 1 51 44 PM

client/src/components/DisplayTable/DisplayTable.tsx
```diff
@@ -128,7 +128,7 @@ const get_default_state_from_props = (props: _DisplayTableProps) => {
   return {
     visible_col_keys,
     sort_by,
-    descending: false,
+    descending: true,
```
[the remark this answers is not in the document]
csong202: After I made `unsorted_initial` false I thought the ReportBuilder looked nicer with the tables sorted descending

client/src/components/DisplayTable/DisplayTable.tsx
```diff
@@ -164,10 +164,9 @@ export class _DisplayTable extends React.Component<
     if (nextProps !== prevState.initial_props) {
       const new_default_state = get_default_state_from_props(nextProps);
 
-      const prev_sort_still_valid = _.includes(
-        new_default_state.visible_col_keys,
-        prevState.sort_by
-      );
+      const prev_sort_still_valid =
+        _.includes(new_default_state.visible_col_keys, prevState.sort_by) &&
+        prevState.sort_by === new_default_state.visible_col_keys[0];
```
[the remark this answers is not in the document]
csong202: It was for grouping by a column that wasn't initially the first column in the table.

For example on Expenditures by Standard Object, if i group by "Standard Object" then group by "all" `sort_by` is still set to "so"

client/src/tables/orgTransferPayments.js
```diff
@@ -58,6 +55,19 @@ export default {
           en: "Type",
           fr: "Type",
         },
+        custom_groupings: {
```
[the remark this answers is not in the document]
csong202: it was from this:


> > > ...
> > 
> > 
> > The special dimensions provide an additional dimension along the default. The only place where this happens right now is the `type` column in `orgTransferPayments`. With the implementation I have at the moment it hasn't caused any problems
> 
> The design decisions I'm suggesting are about preventing/limiting future problems. Please check the motivation portion of my suggestion again, let me know if you want me to elaborate after that and we can get on a call some time!
> 
> Ok though, that complicates my suggestion slightly, but it's fine. How about allowing for index columns in table definitions to have an optional `groupings` property (leaning towards renaming "dimensions" to something like "groupings", for more descriptive/plainer language), working like this:
> 
> * `undefined` (default): the table class provides a dimension with the column's name, using the default of directly grouping by actual values in the corresponding column
> * Empty array: no groupings for this column
> * something like `{name: string, group_by_func: (table_data: Record<string, any>[]) => Record<string, any>[]}[]`: the table class provides dimensions with the provided name that apply the provided `group_by_func`. If a column wants a custom grouping option alongside the default, it just has to also redeclare the default group by
> 
> That's the most direct route to the overall idea I was suggesting, of pushing down the responsibility for handling special cases in to the table definitions themselves. There's definitely ways that suggestion could be improved/made cleaner, I just wanted to demonstrate _some_ possible implementation to get the ball rolling.


I could just have a `can_group_by=true` in every column that can be grouped by and keep the `custom_groupings` for columns that have the normal grouping and a custom grouping. I think `can_group_by=true` is more clear than the `groupings`

client/src/components/DisplayTable/DisplayTable.tsx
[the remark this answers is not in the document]
csong202: i guess I could just not set `unsorted_initial` to false and fix the whole sorting problem that way.

I initially wanted to sort the rows because of this:

> Thing is, if grouping doesn't affect sorting then some of the tables will have weird default orders. In Executive Level the Non-Ex row ends up be between the EX 04 and EX 05 rows. The Voted and Statutory Item rows will be in different orders on different tables. Geographic Region and Transfer Payments by Recipient Region will order the rows differently when grouped by region.

I could undo all my changes to the sorting and open up an issue for after this PR gets merged

client/src/rpb/granular_view.js
csong202: I didn't use `fully_qualified_name` because of this:

Screen Shot 2021-11-15 at 4 04 49 PM

csong202: I used `header` instead of `fully_qualified_name` because of this (on Transfer Payments):

Screen Shot 2021-11-15 at 4 21 23 PM

With using `header`, "Transfer Payments - Type" showed up as just "Type" and I thought that looked better at the time but i could use `fully_qualified_name` instead

client/src/components/DisplayTable/DisplayTable.tsx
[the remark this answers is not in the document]
Stephen-ONeil: The RPB tables specifically could choose to sort descending, although it's maybe better to leave them with the same default as everything else. Just want to be cautious of changing the global default on account of any specific use case like the RPB!

client/src/components/DisplayTable/DisplayTable.tsx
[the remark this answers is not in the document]
Stephen-ONeil: Hm, right, so making a selection in the group by menu also needs to directly set the new grouping column to be the sorting column. That responsibility needs to belong to the group by feature directly though, for the future maintainability reasons I gave.

client/src/rpb/granular_view.js
```diff
@@ -100,70 +111,70 @@ class GranularView extends React.Component {
...
+                  fully_qualified_name,
+                },
+                dimension
+              ),
               is_searchable:
                 is_searchable && !is_matched_undefined(non_dept_key_cols, nick),
               is_summable:
                 is_summable && !is_matched_undefined(data_columns, nick),
-              formatter: type,
+              formatter: nick === "dept" ? "wide-str" : type,
```
Stephen-ONeil, line 125: Is this necessary? Don't all the dept columns have type "wide-str" already?

Stephen-ONeil (changes requested):
Can you make a (careful) find and replace to rename "dimensions" to something like "groupings", for consistency? Not super important if that'd be too messy, ha.

More importantly, IMO, can we think up a better name for the "all" grouping? Grouping by "all" doesn't make sense. This matters more in the user-facing group by options, but again it's nice to keep the code side consistent/meaningful too. For the user facing name, it'd make more sense as something like "Default" or "No grouping".

I'm going to spot check a bunch of panels now, if all good then this is just about finished :tada:

Stephen-ONeil:
Spot checks all look good :tada:

client/src/rpb/granular_view.js
csong202: oh right I think I did that before I made `dept_and_legal_cols`

Stephen-ONeil (approved):
:tada: :+1:

Issue #1175, opened 2021-07-05
Rewrite WrappedNivoCircleProportion as an in-house chart


Stephen-ONeil:
Takes a _lot_ of hacky code to make this chart behave the way we want it, and we're not using the majority of the features of the nivo `circle-packing` package in the process. Just fighting it and hitting even more upgrade-bugs than usual with nivo. The only thing nivo's doing for us here is the functionality and common appearance for the tooltip.

Would be much better to just implement our own two-circle proportion chart (practically trivial) and figure out how to wrap it with nivo's tooltips. That second point's still hacky, since they don't seem to intend their tooltip package for external use. Probably going to create a different sort of nivo-upgrade-gotcha, but I think we want to keep that common look-and-feel to all our tooltips.

PR #1207, opened 2021-07-29
Implemented in-house circle graph and legend

patfire12345:
https://github.com/TBS-EACPD/infobase/issues/1175


client/src/charts/wrapped_nivo/WrappedNivoCircleProportion/WrappedNivoCircleProportion.js
```diff
@@ -119,101 +47,84 @@ export class WrappedNivoCircleProportion extends React.Component {
...
-                    tooltip_item.value / parent_value
-                  )})`}
-                </div>
-              </td>
-            </MediaQuery>
-          </Fragment>
-        )}
-      />
-    );
+    const Table = () => {
```
Stephen-ONeil, line 78: For consistency with other graphs (pie chart at least), use the `TabularLegend` component instead.

client/src/charts/wrapped_nivo/WrappedNivoCircleProportion/WrappedNivoCircleProportion.js
```diff
@@ -119,101 +47,84 @@ export class WrappedNivoCircleProportion extends React.Component {
...
+      const parent_radius = 20;
+      const parent_cx = 175;
+      const parent_cy = 25;
+
+      // child circle calculations
+      const child_percent = child_value / parent_value;
+      const child_radius = parent_radius * child_percent;
+
+      return (
+        <svg viewBox="0 0 350 350">
```
Stephen-ONeil, line 61: The SVG view box dimensions define the coordinate system inside of itself (to decouple it from the actual HTML size of the svg element itself). I'd suggest only hardcoding the viewbox length/height value (350) and calculating the parent radius, x, and y from that, so that there's no risk of them getting de-synced later.

Stephen-ONeil (changes requested):
Doesn't seem to be rendering right, [here](http://dev.ea-ad.ca/CustomNivoCircle/index-eng.html#orgs/dept/1/infograph/people/.-.-(panel_key.-.-'employee_last_year_totals)) at least.
image

client/src/charts/wrapped_nivo/WrappedNivoCircleProportion/WrappedNivoCircleProportion.js
patfire12345: Done!

client/src/charts/wrapped_nivo/WrappedNivoCircleProportion/WrappedNivoCircleProportion.js
patfire12345: Done!

patfire12345:
The resulting component is shown in the figure:
image_2021-08-04_162952

Turns out that the component was rendering correctly, but the ratio was too big, so as a result, the child circle was too small! To remedy this, I put a limit to how small the circle can be (10% of the parent circle radius).

client/src/charts/wrapped_nivo/WrappedNivoCircleProportion/WrappedNivoCircleProportion.js
```diff
@@ -119,101 +38,96 @@ export class WrappedNivoCircleProportion extends React.Component {
...
+      // child circle calculations
+      const child_percent = child_value / parent_value;
+
+      // precaution against smaller child_percent values
+      const true_child_radius = parent_radius * child_percent;
+      const default_child_radius = parent_radius * 0.1;
+      const child_radius =
+        true_child_radius > default_child_radius
+          ? true_child_radius
+          : default_child_radius;
```
Stephen-ONeil, line 58: Check the logic from the old component, it was correct. Want something like
```
const child_radius = _.max([
  parent_radius * Math.sqrt(child_percent), // do the math, make the actual area proportional,
  MIN_NODE_RADIUS,
]);
```

The square root is important! We want the areas of the circles to be proportional, which _isn't_ the same as using proportional radius (since the area itself is proportional to r^2). 

In the old case, `MIN_NODE_RADIUS` was `2` and was picked to be, I think, the smallest radius value that gave a visible inner circle (although this `2` could also be started relative to the svg viewbox dimensions`. We need to be very  careful with this sort of "fudging" of visualizations, misrepresenting the data is the last thing we want. Trying to pick the smallest possible visible radius for the lower limit is preferable as it's too small to actually communicate a proportion anyway, it just communicates that the inner value is not zero.

client/src/charts/wrapped_nivo/WrappedNivoCircleProportion/WrappedNivoCircleProportion.js
```diff
@@ -119,101 +38,96 @@ export class WrappedNivoCircleProportion extends React.Component {
...
+            " " +
+            minY.toString()
+          }
+        >
+          <circle
+            r={parent_radius.toString()}
+            fill={color_scale(parent_name)}
+          />
+          <circle
+            cy={(parent_radius - child_radius).toString()}
```
Stephen-ONeil, line 77: See the comment on the inner node y position calculation on the previous version, probably want to keep the same rule of adding a `- 1` offset (and keeping the comment explaining why).

Stephen-ONeil (changes requested):

client/src/charts/wrapped_nivo/WrappedNivoCircleProportion/WrappedNivoCircleProportion.js
```diff
@@ -119,101 +38,96 @@ export class WrappedNivoCircleProportion extends React.Component {
...
+      const child_radius =
+        true_child_radius > default_child_radius
+          ? true_child_radius
+          : default_child_radius;
+
+      return (
+        <svg
+          viewBox={
+            -parent_cx.toString() +
+            " " +
```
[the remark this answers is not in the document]
Stephen-ONeil: Also, want the offsets to be 0. Using cx and cy here is a bug, see image below
image

The svg extends down much further than it should. It's actually covering up the panel utility buttons on the panel below it, and preventing them from being clicked on directly.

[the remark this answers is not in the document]
patfire12345: Done!

client/src/charts/wrapped_nivo/WrappedNivoCircleProportion/WrappedNivoCircleProportion.js
```diff
@@ -119,101 +38,96 @@ export class WrappedNivoCircleProportion extends React.Component {
...
-            },
-          ]
-        : 0,
+    const Circles = () => {
+      // arbitrary parent values
+      const minX = 350;
+      const minY = 350;
+      const parent_radius = minX - 330;
+      const parent_cx = 175;
+      const parent_cy = 25;
```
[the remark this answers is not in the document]
patfire12345: Done!

client/src/charts/wrapped_nivo/WrappedNivoCircleProportion/WrappedNivoCircleProportion.js
patfire12345: Done! I think the formula provided is correct.

client/src/charts/wrapped_nivo/WrappedNivoCircleProportion/WrappedNivoCircleProportion.js
patfire12345: Done!

client/src/charts/wrapped_nivo/WrappedNivoCircleProportion/WrappedNivoCircleProportion.js
```diff
@@ -119,101 +38,96 @@ export class WrappedNivoCircleProportion extends React.Component {
...
-              name: child_name,
-              value: child_value,
-              color: color_scale(child_name),
-            },
-          ]
-        : 0,
+    const Circles = () => {
+      // arbitrary parent values
+      const minX = 350;
+      const minY = 350;
```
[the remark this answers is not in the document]
patfire12345: Done!

client/src/charts/wrapped_nivo/WrappedNivoCircleProportion/WrappedNivoCircleProportion.js
```diff
@@ -1,112 +1,31 @@
...
-        stroke={borderColor}
-        strokeWidth={borderWidth}
-        shapeRendering={"geometricPrecision"}
-        {...node_handlers}
-      />
-    </g>
-  );
-};
-
 export class WrappedNivoCircleProportion extends React.Component {
```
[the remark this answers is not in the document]
patfire12345: Done!

Stephen-ONeil (commented):
Double check the diff (remember self review!), I think it's possible the state of the branch isn't correct. At least, some of the change requests you commented done on from the last round don't appear to be made in what's currently pushed.

client/src/charts/CircleProportionGraph/CircleProportionGraph.js
```diff
@@ -0,0 +1,167 @@
...
+
+      // this y position will place the bottom of the inner circle just barely above the bottom of the outer circle.
+      // Easier to judge proptions than when it's centered, and the slight offset stops the the svg's edges from
+      // overlaping and looking jagged
+      const child_cy = parent_cy + parent_radius - child_radius - 1;
+
+      return (
+        <div
+          style={{ height: minDimension, width: minDimension, margin: "auto" }}
+        >
```
patfire12345, line 67: Not sure if this is the best way to do this, but it seems to fix the svg bug.

client/src/charts/CircleProportionGraph/CircleProportionGraph.js
```diff
@@ -0,0 +1,167 @@
...
+
+    const color_scale = scaleOrdinal().range(newIBCategoryColors);
+    const value_formatter = get_formatter(is_money, formatter, true, false);
+
+    const Circles = () => {
+      // arbitrary parent values
+      const minDimension = 180;
+      const parent_radius = minDimension / 2;
+      const parent_cx = minDimension / 2;
+      const parent_cy = minDimension / 2;
```
patfire12345, line 46: These parameters ended up working out well, it seems.

client/src/charts/CircleProportionGraph/CircleProportionGraph.js
```diff
@@ -0,0 +1,167 @@
...
+
+      // child circle calculations
+      const child_percent = child_value / parent_value;
+
+      // precaution against smaller child_percent values
+      const true_child_radius = parent_radius * Math.sqrt(child_percent);
+      const default_child_radius = parent_radius * 0.005;
```
patfire12345, line 53: I believe setting the smallest possible circle with a radius of 0.5% of the parent circle is appropriate.

client/src/charts/CircleProportionGraph/CircleProportionGraph.js
```diff
@@ -0,0 +1,167 @@
...
+      const minDimension = 180;
+      const parent_radius = minDimension / 2;
+      const parent_cx = minDimension / 2;
+      const parent_cy = minDimension / 2;
+
+      // child circle calculations
+      const child_percent = child_value / parent_value;
+
+      // precaution against smaller child_percent values
+      const true_child_radius = parent_radius * Math.sqrt(child_percent);
```
patfire12345, line 52: Seems like logic from the old component works, since Rc = Rp * Math.sqrt(Ac / Ap) mathematically.

client/src/charts/wrapped_nivo/WrappedNivoCircleProportion/WrappedNivoCircleProportion.js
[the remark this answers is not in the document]
patfire12345: To be more specific, I've renamed the file and it's folder into CircleProportionGraph.

client/src/charts/ProportionalNode.js
```diff
@@ -1,68 +0,0 @@
...
-        r={node_radius}
-        fill={node.color}
-        stroke={borderColor}
-        strokeWidth={borderWidth}
-        shapeRendering={"geometricPrecision"}
-        {...node_handlers}
-      />
-    </g>
-  );
-};
```
patfire12345: Not sure if we still need this now that the CircleProportionalGraph is made.

client/src/charts/CircleProportionGraph/CircleProportionGraph.js
Stephen-ONeil: I don't think you need the outer div here (the `Circles` component is wrapped in a div by its parent, the resulting html is `<div ...><div ...><svg /></div></div>`, so that div in the middle is unlikely to be necessary). I think the `svg` itself just needs an explicit `style={{height: "100%", width: "100%"}}`.

Stephen-ONeil (changes requested):

client/src/charts/ProportionalNode.js
Stephen-ONeil: Yup, this file vestigial.

client/src/charts/CircleProportionGraph/CircleProportionGraph.js
```diff
@@ -0,0 +1,167 @@
...
+                {value_formatter(item.value)}
+              </span>
+              {
+                <span className="infobase-pie__legend-data">
+                  <Format
+                    type="percentage1"
+                    content={
+                      item.value === parent_value
+                        ? item.value / parent_value
+                        : item.value / (parent_value + child_value)
```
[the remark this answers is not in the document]
patfire12345: Completed under the Fixed Percentage Calculation commit

client/src/charts/CircleProportionGraph/CircleProportionGraph.js
```diff
@@ -0,0 +1,167 @@
...
+    const value_formatter = get_formatter(is_money, formatter, true, false);
+
+    const Circles = () => {
+      // arbitrary parent values
+      const minDimension = 180;
+      const parent_radius = minDimension / 2;
+      const parent_cx = minDimension / 2;
+      const parent_cy = minDimension / 2;
+
+      // child circle calculations
```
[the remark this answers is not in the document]
patfire12345: I've removed the unnecessary comments

client/src/charts/CircleProportionGraph/CircleProportionGraph.js
```diff
@@ -0,0 +1,167 @@
...
+      const child_percent = child_value / parent_value;
+
+      // precaution against smaller child_percent values
+      const true_child_radius = parent_radius * Math.sqrt(child_percent);
+      const default_child_radius = parent_radius * 0.005;
+      const child_radius =
```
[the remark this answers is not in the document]
patfire12345: I've adjusted the logic to make this work

client/src/charts/CircleProportionGraph/CircleProportionGraph.js
```diff
@@ -0,0 +1,167 @@
...
+      disable_table_view,
+      table_name,
+    } = this.props;
+
+    const color_scale = scaleOrdinal().range(newIBCategoryColors);
+    const value_formatter = get_formatter(is_money, formatter, true, false);
+
+    const Circles = () => {
+      // arbitrary parent values
+      const minDimension = 180;
```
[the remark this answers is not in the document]
patfire12345: I've changed minDimension to viewbox_dimension.

client/src/charts/CircleProportionGraph/CircleProportionGraph.js
```diff
@@ -0,0 +1,160 @@
...
+    const Circles = () => {
+      const viewbox_dimension = 180;
+      const parent_radius = viewbox_dimension / 2;
+      const parent_cx = viewbox_dimension / 2;
+      const parent_cy = viewbox_dimension / 2;
+
+      const child_radius = _.max([
+        parent_radius * Math.sqrt(child_value / parent_value),
+        viewbox_dimension * 0.01, // roughly the smallest visible child radius
+      ]);
```
patfire12345, line 50: Screen Shot 2021-08-17 at 10 23 39 AM
I think 0.01 * viewbox_dimension is an appropriate value as the minimum child radius.

client/src/charts/CircleProportionGraph/CircleProportionGraph.js
patfire12345: Done!

Stephen-ONeil (approved):
:+1: :+1: 

I'll open the followup issue about cleaning up the wrapped_nivo_common dependency.

patfire12345:
Closes #1175

PR #1208, opened 2021-07-29
Drop HI tags

Stephen-ONeil:
TODO:
  - [x] coordinate with pipeline (and maybe titan) on dropping HIs from input data


Stephen-ONeil:
Paused until pipeline's other priorities allow for this.

PR #1210, opened 2021-07-30
Moved nav-header style into canada.scss and StandardLegend.scss

patfire12345:

### DIFF
--- a/client/src/charts/canada/canada.js
+++ b/client/src/charts/canada/canada.js
@@ -23,8 +23,6 @@ import { CanadaD3Component } from "./CanadaD3Component";
 
 import text from "./canada.yaml";
 
-import "./canada.scss";
-
 const { text_maker } = create_text_maker_component(text);
 
 const { provinces } = businessConstants;
--- a/client/src/charts/canada/canada.js
+++ b/client/src/charts/canada/canada.js
@@ -56,7 +56,7 @@ class CanadaGraphBarLegend extends React.Component {
 
     return (
       <Fragment>
-        <p className="mrgn-bttm-0 mrgn-tp-0 canada-nav-header centerer">
+        <p className="mrgn-bttm-0 mrgn-tp-0 title-p--canada centerer">
           {province_graph_title(prov)}
         </p>
         <WrappedNivoHBar
--- a/client/src/charts/canada/canada.js
+++ b/client/src/charts/canada/canada.js
@@ -195,23 +195,13 @@ export class Canada extends React.Component {
               checkBoxProps: { isSolidBox: true },
             }}
           />
-          <div
-            className="standard-legend-container"
-            style={{
-              maxHeight: "400px",
-              width: "100%",
-              overflowY: "hidden",
-              marginTop: "10px",
-            }}
-          >
-            <CanadaGraphBarLegend
-              prov={prov}
-              data={data}
-              alt_totals_by_year={alt_totals_by_year}
-              years={years}
-              formatter={formatter}
-            />
-          </div>
+          <CanadaGraphBarLegend
+            prov={prov}
+            data={data}
+            alt_totals_by_year={alt_totals_by_year}
+            years={years}
+            formatter={formatter}
+          />
         </div>
         <div className="col-12 col-lg-9" style={{ position: "relative" }}>
           {years.length > 1 && (
--- a/client/src/charts/canada/canada.js
+++ b/client/src/charts/canada/canada.js
@@ -21,6 +21,7 @@ import { secondaryColor, tertiaryColor } from "src/style_constants/index";
 import { CanadaD3Component } from "./CanadaD3Component";
 
 import text from "./canada.yaml";
+import "./canada.scss";
 
 const { text_maker } = create_text_maker_component(text);
 
@@ -55,7 +56,7 @@ class CanadaGraphBarLegend extends React.Component {
 
     return (
       <Fragment>
-        <p className="mrgn-bttm-0 mrgn-tp-0 nav-header centerer">
+        <p className="mrgn-bttm-0 mrgn-tp-0 canada-nav-header centerer">
           {province_graph_title(prov)}
         </p>
         <WrappedNivoHBar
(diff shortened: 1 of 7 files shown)


Stephen-ONeil (changes requested):
Having a shadow or using `text-transform: uppercase;` seem like legacy design approaches. I'd suggest
  - [ ] restyling these two cases from scratch to something else that looks good and fits the rest of the site's design
  - [ ] giving these more descriptive class names ("nav header" is not accurate in terms of html semantics for a `p`, and it's really how they're used on the page),

Stephen-ONeil:
Actually, the Canada graph case is intended to look the same as the legends, it's not actually a separate case just a hacky hard-coded copy of the legend markup/styling it seems. Might be worth exporting a `LegendContainer` component separate from the `StandardLegend` component, so that it can be re-used directly.

client/src/charts/canada/canada.scss
```diff
@@ -0,0 +1,6 @@
+@import "src/common_css/common-variables";
+
+.title-p--canada {
+  align-self: center;
+  width: 100%;
```
patfire12345, line 5: I think that this style looks alright, but I'm open to suggestions!

client/src/charts/canada/canada.js
```diff
@@ -57,7 +58,7 @@ class CanadaGraphBarLegend extends React.Component {
 
     return (
       <Fragment>
-        <p className="mrgn-bttm-0 mrgn-tp-0 nav-header centerer">
+        <p className="mrgn-bttm-0 mrgn-tp-0 title-p--canada centerer">
```
patfire12345, line 61: Trying to follow the BEM naming convention, I ended up with title-p--canada. Open to changing this into something that fits more.

client/src/charts/legends/LegendContainer.js
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+
+import { tertiaryColor } from "src/core/color_defs";
+
+export const LegendContainer = ({ title, legend_footer, children }) => (
```
Stephen-ONeil, line 5: Port this file to typescript please.

client/src/charts/canada/canada.js
Stephen-ONeil: Should switch to using `LegendContainer` in this file now, the title can just be passed in as a prop to that (so don't need `title-p--canada` either).

client/src/charts/legends/StandardLegend.scss
```diff
@@ -7,3 +7,9 @@
   border-radius: $standardBorderRadius;
   margin-bottom: 4px;
 }
+
+.title-p--standard-legend {
```
Stephen-ONeil, line 11: Proper BEM would be something like `standard-legend-container__title`

Think of BEM as block__element--modifier. Block is roughly analogous to the React element itself. Element is some specific sub-element of that (preferably identifiable as a concrete concept like "title"). Modifier is some variable state effecting either a block or element's style, such as `--active` etc.

The BEM names do not need to, and generally _shouldn't_, represent the actual corresponding HTML structure. HTML structure can be explicitly represented in the selectors if necessary. Otherwise, the BEM class names are a bit of an abstraction over top of the actual HTML implementation. In theory, there could be plenty of cases where HTML structure is edited but the corresponding BEM class names remain the same. E.g. the title could be moved around within the legend container or be changed from a `p` to a `div` or `a`, but it would still be, conceptually, a "title belonging to the legend container" and likely should still be styled the same.

Definitely not advised to ever put an html tag, like `p`, in the class name. If it's actually important to the selector, than you would write it as something like `p.<class name>` rather than `.<class name mentioning p>`.

Stephen-ONeil (changes requested):

client/src/charts/legends/StandardLegend.scss
patfire12345: Ok thanks for the advice!

client/src/charts/legends/LegendContainer.js
patfire12345: Done!

client/src/charts/canada/canada.js
```diff
@@ -56,10 +59,7 @@ class CanadaGraphBarLegend extends React.Component {
       .value();
 
     return (
-      <Fragment>
-        <p className="mrgn-bttm-0 mrgn-tp-0 nav-header centerer">
-          {province_graph_title(prov)}
-        </p>
+      <LegendContainer title={province_graph_title(prov)}>
```
patfire12345, line 62: I think Fragment can be removed now that LegendContainer is here.

client/src/charts/canada/canada.scss
```diff
@@ -0,0 +1,7 @@
...
+  color: #206bbd;
+}
```
Stephen-ONeil, line 7: Delete, clean up import of this file

Stephen-ONeil (commented):
Almost done, just a little cleanup left I think!
- [x] delete the legend-container styled div at line 202 in `canada.js`

client/src/charts/canada/canada.scss
patfire12345: Done!

Stephen-ONeil (changes requested):
> delete the legend-container styled div at line 202 in canada.js

Still need to do this. You've got double-containers on the legend right now.
image

Just noticed the conflict warnings. After you fix that, please rebase off master before I do the final review/approval. Thanks!

client/src/charts/canada/canada.js
```diff
@@ -197,23 +195,13 @@ export class Canada extends React.Component {
...
-            }}
-          >
-            <CanadaGraphBarLegend
-              prov={prov}
-              data={data}
-              alt_totals_by_year={alt_totals_by_year}
-              years={years}
-              formatter={formatter}
-            />
-          </div>
```
patfire12345, line 216: This should remove the double container, I must have overlooked the checkbox comment earlier. Sorry for having to repeat yourself.

Stephen-ONeil (approved):
:+1:

PR #1211, opened 2021-07-30
Moved copyable-hidden style to utils.scss

patfire12345:
Only found this style within result_flat_table.js, so it makes sense to move it. However, I am not sure where this file is exactly on the GC Infobase website.

### DIFF
--- a/client/src/common_css/common_css_index.side-effects.js
+++ b/client/src/common_css/common_css_index.side-effects.js
@@ -1,1 +1,2 @@
 import "./site.scss";
+import "./utils.scss";
--- a/client/src/common_css/site.scss
+++ b/client/src/common_css/site.scss
@@ -142,11 +142,6 @@ strong,
   }
 }
 
-.copyable-hidden {
-  position: absolute;
-  left: -999em;
-}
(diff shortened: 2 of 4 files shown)


Stephen-ONeil (changes requested):
It's the table in [this](http://dev.ea-ad.ca/master/index-eng.html#orgs/dept/1/infograph/results/.-.-(panel_key.-.-'results_flat_table)) panel, in the results infographics below the gov level.

As a generic utility class, I think it's reasonable for it to belong to the global styles. Only specific to that results table because we have nowhere else that needs it, but in theory it's usable anywhere. I'd prefer to keep it in `site.scss` (or move it to something like `utils.scss` if you want to start the reorganizing part of this cleanup).

patfire12345:
I've created the utils.scss file and placed the copyable-hidden style into it.

Stephen-ONeil (approved):
:+1: 

Approved, please keep moving other util classes in to here as you go. Also, before merging, could you rename the PR to be accurate to what's happening instead? Will make the PR history less confusing.

PR #1215, opened 2021-08-10
Drop client GraphiQL route, related webpack config cleanup

Stephen-ONeil:
In dev, we can just use the dev GraphQL server's GraphiQL end point. I could be wrong, but it's not been used much for dev link and prod deploys (but if so, we should look in to an alternative).

This was most likely going to happen inevitably in the much-delayed npm 7 upgrade, since GraphiQL (kind of ironic, facebook) is one of the big laggards on formally supporting React 17 as a peer dependency. 

TODO: 
  - [x] double check IE 11 once dev link is up, just to be careful
    - ok, currently broken but probably just need to make a small regex tweak. Leaving open for review in the meantime


Stephen-ONeil:
Ahh, haha, argh.

Couldn't figure out why ANY of my regex wasn't equivalent to the existing include/test rules. Shouldn't have been any reason for transpilation to have broken under my changes...

Until I finally noticed the trailing `|)` in the existing include rule:
` include:
        /node_modules\/(graphiql|graphql-language-service-.*|codemirror-graphql|codemirror|d3-scale|@nivo\/bar|@nivo\/circle-packing|@nivo\/core|@nivo\/line|@nivo\/pie|)/`
        
... so the whole hardcoded list was useless (and presumably unmaintained as a result), this was matching and transpiling ALL node modules. Turns out, we have other dependencies (often indirect, many are sub-dependencies of nivo) that don't support IE11.

So, I'm just going to drop the hardcoded list altogether, hit all node modules with babel when building with `target_ie11 = true`. It's what we've already been doing, and it will be easier to maintain.

Issue #1216, opened 2021-08-10
Dept name into program infographics navigation breadcrumbs, possibly also CR


ktw1016:
<img width="835" alt="Screen Shot 2021-08-10 at 4 24 14 PM" src="https://user-images.githubusercontent.com/29746626/128930051-1b8bb3e4-54f6-42ce-8510-1d969c0efd7c.png">

For program infographics, we should put department in there. Possibly also CR but let's see how it looks first for both

ktw1016:
Closed by #1222

PR #1222, opened 2021-08-11
Adding Department Names to Program Breadcrumbs

patfire12345:
https://github.com/TBS-EACPD/infobase/issues/1216

### DIFF
--- a/client/src/common_text/common_lang.yaml
+++ b/client/src/common_text/common_lang.yaml
@@ -1131,8 +1131,10 @@ infographic_for:
 
 department_of:
   transform: [handlebars]
-  en: (Department of {{subject.dept.applied_title}})
-  fr: (Département de {{subject.dept.applied_title}})
+  en: |
+    {{subject.name}} ({{subject.dept.applied_title}})
+  fr: |
+    {{subject.name}} ({{subject.dept.applied_title}})
 
 ##################################
 # footer text and links
--- a/client/src/common_text/common_lang.yaml
+++ b/client/src/common_text/common_lang.yaml
@@ -1129,6 +1129,11 @@ infographic_for:
   en: Infographic for {{subject.name}}
   fr: Infographie pour {{subject.name}}
 
+department_of:
+  transform: [handlebars]
+  en: (Department of {{subject.dept.applied_title}})
+  fr: (Département de {{subject.dept.applied_title}})
+
 ##################################
 # footer text and links
 ##################################
--- a/client/src/common_text/common_lang.yaml
+++ b/client/src/common_text/common_lang.yaml
@@ -1136,6 +1136,13 @@ infographic_title:
   fr: |
     {{subject.name}}
 
+crso_title:
+  transform: [handlebars]
+  en: |
+    {{subject.crso.name}}
+  fr: |
+    {{subject.crso.name}}
+
 department_title:
   transform: [handlebars]
   en: |
--- a/client/src/common_text/common_lang.yaml
+++ b/client/src/common_text/common_lang.yaml
@@ -1129,27 +1129,6 @@ infographic_for:
   en: Infographic for {{subject.name}}
   fr: Infographie pour {{subject.name}}
 
-infographic_title:
-  transform: [handlebars]
-  en: |
-    {{subject.name}}
-  fr: |
-    {{subject.name}}
-
-crso_title:
-  transform: [handlebars]
-  en: |
-    {{subject.crso.name}}
-  fr: |
-    {{subject.crso.name}}
-
-department_title:
-  transform: [handlebars]
-  en: |
-    {{subject.dept.applied_title}}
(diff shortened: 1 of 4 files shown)


client/src/common_text/common_lang.yaml
```diff
@@ -1122,6 +1122,13 @@ infographic_for:
   en: Infographic for {{subject.name}}
   fr: Infographie pour {{subject.name}}
 
+department_of:
+  transform: [handlebars]
+  en: |
+    {{subject.name}} ({{subject.dept.applied_title}})
+  fr: |
+    {{subject.name}} ({{subject.dept.applied_title}})
+
```
patfire12345, line 1131: Tried to truncate unnecessary words like "Department of" and "Infographic for" for programs, but some are still too long and get cut off on the website.

ktw1016 (changes requested):
As clarified on Slack, we'd want `InfoBase > DEPARTMENT href link > program` for better navigation

client/src/infographic/Infographic.js
```diff
@@ -422,8 +422,18 @@ const Infographic = ({
...
+    level === "program"
+      ? [
+          {
+            title: text_maker("department_title", { subject }),
+            department: true,
+            id: subject.dept.id,
+          },
+          { title: text_maker("infographic_title", { subject }) },
+        ]
+      : [{ title: text_maker("infographic_title", { subject }) }];
```
patfire12345, line 436: There could potentially be a better way to format this.

patfire12345:
Screen Shot 2021-08-23 at 10 33 59 AM
This is what it looks like with Infobase > Dept > CR > Program. Let me know if it looks too long!

client/src/infographic/Infographic.js
ktw1016: Was there a reason for putting them thru yaml? You can just do `subject.name` directly

client/src/infographic/Infographic.js
```diff
@@ -422,8 +422,19 @@ const Infographic = ({
       <Redirect to={infograph_href_template(subject_parent, bubble_id, "/")} />
     );
   }
-
   const title = text_maker("infographic_for", { subject });
+  const breadcrumb_program =
```
ktw1016, line 426: As far as I can see, you can directly pass in array of JSX elements. So you can pass in the `<a>` tags directly. Please remove the changes in NavComponents and try passing in react elements directly

client/src/core/NavComponents.js
```diff
@@ -102,8 +102,15 @@ class BreadCrumbs extends React.Component {
             <li className="infobase-home-breadcrumb-link">
               {_.isString(display) ? ( //allow strings or react elements to be used here (note that some strings may have the weird french apostrophe that needs to non-escaped)
                 <span dangerouslySetInnerHTML={{ __html: display }} />
+              ) : display.department ? (
+                <a
+                  href={`#orgs/dept/${display.id}/infograph/results`}
```
ktw1016, line 107: There are utils you can use in `infographic_link.js`. Look into that!

ktw1016 (changes requested):
Also, make sure CR is a href

client/src/infographic/Infographic.js
patfire12345: Since it was originally using a text_maker function, I decided to follow the format. I'll do it like how you suggested!

client/src/infographic/Infographic.js
```diff
@@ -422,8 +422,32 @@ const Infographic = ({
...
     );
   }
-
   const title = text_maker("infographic_for", { subject });
+  const breadcrumbs = (() => {
+    const get_breadcrumbs = (parent_subjects) =>
+      _.chain(parent_subjects)
+        .map((parent_subj) => (
+          <a
+            href={infograph_href_template(parent_subj, active_bubble_id)}
```
ktw1016, line 431: I was gonna make this active bubble id smart (services exist for program but not for crso, etc) but figured I'd leave it as is since typeahead behaves the same way and couldn't think of a fully automated way

Stephen-ONeil (changes requested):
Didn't have time to look at code, but here's some style feedback for now.

I'd suggest some DOM and CSS changes to make the longer breadcrumbs fit better. Something like this: 
image

Rather than this:
image

I'd recommend moving the >'s from separate `li`'s to `spans` inside the `li` of the corresponding item, and also removing the `text-overflow` and `white-space` styles on `#wb-bc .breadcrumb li`. That way the arrows stay on the same line as the corresponding item and the names can wrap lines when long instead of cutting off.
image

Also, noticed that the link to the home is black text while the intermediate links are blue. Should make that home link the same blue too for consistency with other links on the site/in the breadcrumb.

ktw1016:
Agreed, merging in with gov added into breadcrumbs

PR #1224, opened 2021-08-17
SortDirection converted from toggling to specific direction sorting

sranc046:
The arrows used to act like one toggling button, didn't matter which arrow you clicked, the sorting direction just toggled back and forth. Direction is now sorted per the arrow you click. Column header is no longer clickable but maybe we leave the header act as toggle? Will update the SortDirection stories to match new format.

### DIFF
--- a/client/src/EstimatesComparison/scheme.js
+++ b/client/src/EstimatesComparison/scheme.js
@@ -501,12 +501,13 @@ export class EstimatesExplorer extends AbstractExplorerScheme {
       };
     }
     if (type === "column_header_click") {
-      const { is_descending, sort_col } = state;
-      const clicked_col = payload;
+      const { sort_col } = state;
+      const clicked_col = payload.col_key;
+      const direction = payload.direction;
       const mods =
         clicked_col === sort_col
-          ? { is_descending: !is_descending }
-          : { is_descending: true, sort_col: clicked_col };
+          ? { is_descending: direction === "DESC" }
+          : { is_descending: direction === "DESC", sort_col: clicked_col };
       return { ...state, ...mods };
     } else {
       return state;
@@ -527,8 +528,11 @@ export class EstimatesExplorer extends AbstractExplorerScheme {
 
   @bound
   map_dispatch_to_props(dispatch) {
-    const col_click = (col_key) =>
-      dispatch({ type: "column_header_click", payload: col_key });
+    const col_click = (col_key, direction) =>
+      dispatch({
+        type: "column_header_click",
+        payload: { col_key, direction },
+      });
     const toggle_stat_filter = () => dispatch({ type: "toggle_stat_filter" });
     const toggle_legal_titles = () => dispatch({ type: "toggle_legal_titles" });
 
--- a/client/src/TagExplorer/resource_scheme.js
+++ b/client/src/TagExplorer/resource_scheme.js
@@ -229,8 +229,11 @@ export class ResourceScheme extends AbstractExplorerScheme {
 
   @bound
   map_dispatch_to_props(dispatch) {
-    const col_click = (col_key) =>
-      dispatch({ type: "column_header_click", payload: col_key });
+    const col_click = (col_key, direction) =>
+      dispatch({
+        type: "column_header_click",
+        payload: { col_key, direction },
+      });
 
     return {
       ...super.map_dispatch_to_props(dispatch),
@@ -273,14 +276,13 @@ export class ResourceScheme extends AbstractExplorerScheme {
     } else if (type === "set_hierarchy_scheme") {
       return { ...state, hierarchy_scheme: payload };
     } else if (type === "column_header_click") {
-      const { is_descending, sort_col } = state;
-      const clicked_col = payload;
-
+      const { sort_col } = state;
+      const clicked_col = payload.col_key;
+      const direction = payload.direction;
       const mods =
         clicked_col === sort_col
-          ? { is_descending: !is_descending }
-          : { is_descending: true, sort_col: clicked_col };
-
+          ? { is_descending: direction === "DESC" }
+          : { is_descending: direction === "DESC", sort_col: clicked_col };
       return { ...state, ...mods };
     } else if (type === "set_year") {
       return { ...state, year: payload };
--- a/client/src/components/DisplayTable/DisplayTable.tsx
+++ b/client/src/components/DisplayTable/DisplayTable.tsx
@@ -156,6 +156,8 @@ export class _DisplayTable extends React.Component<
       current_page: 0,
       show_pagination_load_spinner: false,
     };
+
+    this.sortCol = this.sortCol.bind(this);
   }
   static getDerivedStateFromProps(
     nextProps: _DisplayTableProps,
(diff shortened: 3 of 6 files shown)


Stephen-ONeil (changes requested):
TypeScript errors in build, take a look at fixing those then put it back up for review.

client/src/components/DisplayTable/DisplayTable.tsx
```diff
@@ -227,6 +229,15 @@ export class _DisplayTable extends React.Component<
     this.debounced_stop_loading.cancel();
   }
 
+  sortCol(column_key: string, dir: string) {
+    console.log("clicked");
```
ktw1016, line 233: Should be cleaned up

client/src/components/SortDirection/SortDirection.tsx
```diff
@@ -7,15 +7,21 @@ import "./SortDirection.scss";
 interface SortDirectionProps {
   sortDirection: string;
   active: boolean;
+  sortFunction: CallableFunction;
```
ktw1016, line 10: this should be like `(col: string, sortDirection: "ASC" || "DESC") => void`

ktw1016 (commented):
this is good!

client/src/EstimatesComparison/scheme.js
```diff
@@ -501,12 +501,13 @@ export class EstimatesExplorer extends AbstractExplorerScheme {
...
-      const { is_descending, sort_col } = state;
-      const clicked_col = payload;
+      const { sort_col } = state;
+      const clicked_col = payload.col_key;
+      const direction = payload.direction;
       const mods =
         clicked_col === sort_col
-          ? { is_descending: !is_descending }
-          : { is_descending: true, sort_col: clicked_col };
+          ? { is_descending: direction === "DESC" }
```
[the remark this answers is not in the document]
Stephen-ONeil: (well, if we were using `is_descending` I guess there could be a meaningful third option, which is when no sorting is applied, but that may be more useful as a separate prop for `is_sorting`, likely calculated by the consumer from something like `sort_col`, rather than as a special meaning for `is_descending === null`)

client/src/components/SortDirection/SortDirection.tsx
```diff
@@ -38,9 +45,24 @@ export const SortDirection = ({
   </button>
 );
 
-export const SortDirections = ({ asc, desc }: SortDirectionsProps) => (
+export const SortDirections = ({
+  asc,
+  desc,
+  sortFunction,
+  col,
+}: SortDirectionsProps) => (
```
Stephen-ONeil, line 53: IMO, `SortDirections` shouldn't need to know about `col`. `col` is a concept leaking down from the current use cases. It could be possible that this is used in a context where columns aren't meaningful (i.e. maybe sorting a list rather than a table). The consumer knows more about the context it's using the `SortDirections` in either way, and if it's one where there are multiple columns managed by one consumer than it can always smuggle that context in via the callback (i.e. `sortFunction={(direction) => this.sort(relevant_col, direction)}` passed to each `SortDirections` on render).

`sortFunction` sounds like it performs the sorting, but it shouldn't. Most likely it's just a state update. Maybe `onDirectionClick`? Conveys that it's mostly an `onClick`, but that it has some relation to the direction concept too. Alternatively, maybe just make the consumer provide two click callbacks, `onClickDescending` and `onClickAscending`? Again, the consumer will hold all the information and can smuggle whatever it wants in via the callbacks, so we might as well separate brains/behaviour and presentation as much as we can.

Stephen-ONeil (changes requested):

client/src/components/SortDirection/SortDirection.tsx
Stephen-ONeil: Unless part of the idea here is specifically to move behaviour/smarts down in to this component? When I asked for links to the previous discussions, that's what I was trying to figure out. There's some benefits if that's the plan (enforced consistent behaviour between uses of this component), but IMO also downsides (lose the clean smart vs presentational component design, likely going to mean leaks/coupling between this component and consumers).

Stephen-ONeil (changes requested):
Please fix that build error and re-flag it for review when the dev link is fresh.

client/src/TagExplorer/resource_scheme.js
```diff
@@ -273,14 +276,16 @@ export class ResourceScheme extends AbstractExplorerScheme {
     } else if (type === "set_hierarchy_scheme") {
       return { ...state, hierarchy_scheme: payload };
     } else if (type === "column_header_click") {
-      const { is_descending, sort_col } = state;
-      const clicked_col = payload;
-
+      const { sort_col } = state;
+      const clicked_col = payload.col_key;
+      const is_descending_bool = payload.is_descending_bool;
```
Stephen-ONeil, line 281: Not significant, but just wanted to note that you can declare these using object destructuring, including the renaming of `col_key` to `clicked_col`. Would look like this `const { col_key: clicked_col, is_descending_bool } = payload;`

client/src/EstimatesComparison/scheme.js
```diff
@@ -501,12 +501,13 @@ export class EstimatesExplorer extends AbstractExplorerScheme {
       };
     }
     if (type === "column_header_click") {
-      const { is_descending, sort_col } = state;
-      const clicked_col = payload;
+      const { sort_col } = state;
+      const clicked_col = payload.col_key;
+      const is_descending_bool = payload.is_descending_bool;
```
Stephen-ONeil, line 506: Minor note, a name like `is_...` implies a boolean generally (or possibly a getter or, less likely, a method that returns a boolean), or at least something that is truthy/falsey that can be treated as one if we're playing it loose with types. Either way, encoding type in the variable name is usually avoided/avoidable.

client/src/TagExplorer/resource_scheme.js
```diff
@@ -273,14 +276,16 @@ export class ResourceScheme extends AbstractExplorerScheme {
...
       const mods =
         clicked_col === sort_col
-          ? { is_descending: !is_descending }
-          : { is_descending: true, sort_col: clicked_col };
-
+          ? { is_descending: is_descending_bool === "DESC" }
```
Stephen-ONeil, line 284: Oh, is this some old code or is `is_descending_bool` sometimes string valued after all?

client/src/components/SortDirection/SortDirection.tsx
```diff
@@ -27,6 +30,7 @@ export const SortDirection = ({
       sortDirection === "ASC" ? "a11y_sort_asc" : "a11y_sort_desc"
     )}
     aria-pressed={active}
+    onClick={() => onDirectionClick(sortDirection === "DESC")}
```
Stephen-ONeil, line 33: How about something like `onDirectionClick(active ? sortDirection === "DESC" | sortDirection !== "DESC" )`, so that clicking on the active sort option can still toggle, rather than do nothing? Minor, but I think it's a little better.

Stephen-ONeil (changes requested):
Just a couple minor comments, nothing big to change so all good to also enable and fix up the storybook :+1:

Stephen-ONeil (approved):
:+1:

PR #1228, opened 2021-08-23
About page content update

sranc046:
Updating the content of the About page.

### DIFF
--- a/client/src/about/about.yaml
+++ b/client/src/about/about.yaml
@@ -11,7 +11,8 @@ principle_1_name:
   fr: Uniformité
 
 principle_1_desc:
-  en: GC InfoBase facilitates the analysis of previously hard to access information by aligning data from various public reports into a consisten framework based on Volume II of the Public Accounts. We help "connect the dots".
+  transform: [handlebars, markdown]
+  en: GC InfoBase facilitates the analysis of previously hard to access information by aligning data from various public reports into a consistent framework based on Volume II of the {{gl_tt "Public Accounts" "PA"}}. We help "connect the dots".
   fr: L'InfoBase du GC facilite l'analyse de l'information, auparavant difficile d'accès, en regroupant les données de divers rapports publics dans un cadre uniforme reposant sur le volume II des Comptes publics. Elle permet de faire le lien entre les différents éléments.
 
 principle_2_name:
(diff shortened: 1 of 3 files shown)


client/src/about/about.yaml
```diff
@@ -121,59 +119,77 @@ our_data_sub_title_2:
...
+    Alors que les ensembles de données sont accessibles aux utilisateurs sur le {{ext_link "Portail de données ouvertes" "https://ouvert.canada.ca/fr/donnees-ouvertes"}} et les rapports officiels du gouvernement du Canada, l'InfoBase du GC est la seule tentative d'intégration de ces ensembles de données dans un référentiel unique destiné aux parlementaires, aux fonctionnaires et à la population canadienne. À cette fin, l'InfoBase du GC évolue constamment avec l'ajout de nouvelles sources de données, de visualisations et d’explications. <a href="#contact">Communiquez avec nous</a> si vous avez des données, des réflexions ou des idées à partager!
```
sranc046, line 147: In the english section we changed it from "Canadian population" to "Canadians". Would we not want to do the same thing for the french section?

sranc046:
Fixed a few french typos here and there throughout the sections.

client/src/about/about.yaml
ktw1016: Yeah please do!

ktw1016:
Seems pretty trivial, someone who wrote it (I think Ross, or Val) should do a final review. Also, just reading About page made me wonder, why don't we have glossary tooltips? Ones I can think of that we can use are
Public Accounts
Estimates
Departmental Plans
Departmental Results Reports
Results

There could be more but just at a glance

sranc046:
> Also, just reading About page made me wonder, why don't we have glossary tooltips? Ones I can think of that we can use are
> Public Accounts
> Estimates
> Departmental Plans
> Departmental Results Reports
> Results
> 
> There could be more but just at a glance

Good point, unless there was a specific reason we don't have any I can add some in

client/src/about/about.yaml
```diff
@@ -4,82 +4,81 @@ about_page_title:
...
 principle_1_name:
-  en: Authoritative
-  fr: Source fiable, faisant autorité
+  en: Consistent
+  fr: Uniformité
 
 principle_1_desc:
-  en: Mirroring documents such as the Public Accounts, Estimates, and Departmental Plans in a timely manner. The data you find here will match what you find in public reports as they are tabled.
-  fr: InfoBase du GC reflète en temps opportun des documents, tels que les comptes publics, les budgets des dépenses et les plans ministériel. Vous y trouverez des données qui correspondent à celles des rapports publics au fur et à mesure de leur dépôt.
+  en: GC InfoBase facilitates the analysis of previously hard to access information by aligning data from various public reports into a consisten framework based on Volume II of the Public Accounts. We help "connect the dots".
```
ktw1016, line 14: PA tooltip here also

client/src/about/about.yaml
```diff
@@ -121,59 +119,77 @@ our_data_sub_title_2:
...
     croissants des utilisateurs.
 
 behind_scenes_sub_title_2:
+  en: HOW DOES THE INFORMATION ON GC INFOBASE DIFFER FROM OFFICAL REPORTS?
+  fr: COMMENT LES RENSEIGNEMENTS DANS L'INFOBASE DU GC DIFFÈRENT-ILS DES RAPPORTS OFFICIELS?
+
+behind_scenes_sub_text_2:
+  transform: [handlebars, markdown]
+  en: |
+    Whenever possible, we align the information on GC InfoBase to the universe of organizations and concepts that are presented in Volume II of the {{gl_tt "Public Accounts" "PA"}}. As a result, organizations that are presented separately in other reports may be grouped together on GC InfoBase to match the {{gl_tt "Public Accounts" "PA"}}.
```
ktw1016, line 176: I think gl_tt once per section of the same word would be pretty reasonable so that there is just one PA gl_tt for behind_scenes, once for our_data

ktw1016 (approved):
I just made some small cleanups commit, everything looks good to me

PR #1229, opened 2021-08-25
historical years to exp, fte for CR

### LINKED ISSUE #1227: Add CR level panel for actual resources by program, clarify existing planned resources panel
CRs have panels for planned spending and planned FTEs by program ([e.g.](https://www.tbs-sct.gc.ca/ems-sgd/edb-bdd/index-eng.html#orgs/crso/HRSD-BGM00/infograph/financial/.-.-(panel_key.-.-'crso_by_prog_fte))). For some reason, these panels are specifically only for planned resources. I presume they were made the first year we had CRs and no one ever updated them after...

Should create separate actual resources panels and clarify the names and text of the existing planned resources panels for now. Some day we'll have to do a review of all financial infographic levels to try and enforce some consistency and sanity, but not a short term priority.

ktw1016:
closes #1227


client/src/panels/panel_declarations/finances/drr_dp_resources/crso_by_prog.js
```diff
@@ -144,10 +167,29 @@ class PlannedProgramResources extends React.Component {
...
+              {...{
+                ...(gap_year && {
+                  markers: [
+                    {
+                      axis: "x",
+                      value: gap_year,
+                      lineStyle: {
+                        stroke: tertiaryColor,
+                        transform: `translate(${
+                          0.3 * window_width_last_updated_at * 0.08 // Prob need media query to better handle
```
ktw1016, line 184: Ehh doesn't really work all that well, I can try putting in media query to adjust better for 992px and smaller

ktw1016: But maybe could just remove this line

client/src/panels/panel_declarations/finances/drr_dp_resources/crso_by_prog.js
```diff
@@ -142,13 +196,30 @@ class PlannedProgramResources extends React.Component {
...
+            </MediaQuery>
+            <MediaQuery
+              minWidth={minExtraSmallDevice}
+              maxWidth={maxSmallDevice}
+            >
+              {get_nivo_bar_graph(17)}
+            </MediaQuery>
+            <MediaQuery maxWidth={maxExtraSmallDevice}>
+              {get_nivo_bar_graph(9)}
+            </MediaQuery>
```
Stephen-ONeil, line 222: It works well enough, but all these hardcoded values are brittle against other nivo props/nivo internals (e.g. if they change the whole component's viewbox or something, whoch I don't trust them _not_ to do). Its a bit low value to make it more resilient to be fair, although if merged as is it would need a follow up issue to do something about some day (if nothing else, it will remind us to keep checking whether it's broken yet in the mean time, haha).

Stephen-ONeil (commented):
- [ ] I think the  `footnotes` panel args need to be updated, probably still just for `PLANNED_...`, right?
  - [ ] panel text args need to be reworked, possibly the text too
    - right now, it always uses `planning_year_1`

Stephen-ONeil:
Oops, was still writing that comment about panel text, will edit it to finish my thoughts.

ktw1016:
I made the text separate for most recent historical year and planning year 1. But it's identical sentence😢

client/src/panels/panel_declarations/finances/drr_dp_resources/crso_by_prog.js
ktw1016: Nivo always giving us pain😢

client/src/panels/panel_declarations/finances/drr_dp_resources/crso_by_prog.js
```diff
@@ -322,9 +322,23 @@ const get_calculate_func = (is_fte) => {
...
 
-    const relevant_data = is_fte ? fte_data : exp_data;
-    const first_year_top_2_programs = _.chain(relevant_data)
-      .sortBy(({ data }) => _.first(data))
+    const relevant_data = is_fte ? queried_fte.data : queried_exp.data;
+    const most_recent_top_2_programs = _.chain(relevant_data)
+      .flatMap((program_data) => ({
+        prgm: program_data.prgm,
+        value:
+          program_data[is_fte ? "{{pa_last_year}}" : "{{pa_last_year}}exp"],
```
Stephen-ONeil, line 330: I don't have an example in mind, but theoretically a CR will die and then hang around, so `pa_last_year` might not always be the "most recent" year for actual spending.

Fine to hardcode planning to `planning_year_1` because, if it has any planned data, it will exist.

Part of the pain points of putting both in to one graph, it needs to work in a lot of cases:
  1) dead CR with historical actual spending, no planned
  1.5) not sure if this _can_ happen, but might be a non-dead CR with no planned some day :shrug:
  2) ongoing actual spending\*, planned spending\**
  3) planned spending only

\* anywhere from one to five years, but can assume they are consecutive and include the most recent year
\** should always be three years out, should have data for at least planning year 1 but might be dead after that

client/src/panels/panel_declarations/finances/drr_dp_resources/crso_by_prog.js
```diff
@@ -180,29 +309,68 @@ const get_calculate_func = (is_fte) => {
...
+      }))
+      .sortBy("value")
+      .takeRight(2)
+      .reverse()
+      .value();
+    const most_recent_planned_top_2_programs = _.chain(relevant_data)
```
Stephen-ONeil, line 336: Small nitpick, but referring to planed spending as "most recent" was a little confusing to me at first. If this is hardcoded to planning year 1 then just call it that IMO.

Stephen-ONeil (changes requested):
I think the text tweeks should be pretty easy. They can be nearly the same sentence, the first just needs to refer to the resources as "actual" and use past tense and the second needs to refer to "planned" and use future tense.

If it's easier, feel free to simplify all the cases about picking top 2 and just refer to the top most program in the text. That was likely there to give more substance to the panel, but (in most cases) we'll have enough just from presenting both planned and actual now.

client/src/panels/panel_declarations/finances/drr_dp_resources/crso_by_prog.js
ktw1016: Okay, if I change the hard coded `pa_last_year` to be dynamic, I should be covered for all 3.5 cases I think

Stephen-ONeil (changes requested):
- [x] First sentence should be past tense, second future tense
- [ ] I'd suggest asking for feedback from the team in case anyone has suggestions on more useful text. Might invalidate the above item, if anyone has a good enough idea

Otherwise, I'm fine moving this along. Stand by my opinion that the brittle media query setup will need an immediate (low priority) followup issue to keep track of it for when a future layout change breaks it.

PR #1230, opened 2021-08-27
Glossary tooltip

csong202:
Issue: #1016 

- Created `TooltipManager` component which is a direct child of ErrorBoundry.
  - renders `TooltipActivator` or `TooltipModalDelegator` depending on `is_a11y_mode`
- Created `TooltipModalDelegator` component 
  - Will handle replacing tooltips with a new `ModalButton` component
    -  the original `ModalButton` (the one that uses withRouter) has been renamed to `ModalButtonWithRouter`
    - the new modal component handles the modal closing and returns a `StatelessModal`
    - Decided to use `StatelessModal` because `ModalButton` uses `withRouter` and the `TooltipModalDelegator` is outside the `Route` component
 
TODOs:
- [ ] fix the modal body layout
- [ ] maybe storybook the new `ModalButton`, it's pretty similar to the other modals though


Stephen-ONeil:
Superseded by #1270. Expect some of the work from here to be cherry picked over to that branch though. Delete this branch once that has occurred.

PR #1231, opened 2021-08-30
ErrorBoundary catch unhandled promise rejections

Stephen-ONeil:
Allows `ErrorBoundary` to see unhandled promsie rejections, which are expected to be (and treated the same as) errors. Previously, these wouldn't trigger the `ErrorBoundary`. Will need to do a fairly thorough pass on the site just to make sure there weren't any "quiet" unhandled promise rejections that were previously allowed. 

No IE 11 support for `unhandledrejection` (unless babel is polyfilling promises in our prod builds, haven't double checked but it probably isn't since IE 11 has promises already). With its ever decreasing browser share, I think t hat's fine.

### DIFF
--- a/client/src/core/ErrorBoundary.js
+++ b/client/src/core/ErrorBoundary.js
@@ -21,6 +21,14 @@ const state_from_error = (error) => ({
   testing_for_stale_client: true,
 });
 
+const log_error = (error) =>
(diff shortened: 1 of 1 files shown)


Stephen-ONeil:
Oops, branch name's too long.

PR #1238, opened 2021-09-02
Cypress error tracking

sranc046:
Added functionality so that Cypress can track error logs on console and fail tests accordingly.

### DIFF
--- a/client/cypress/support/commands.js
+++ b/client/cypress/support/commands.js
@@ -17,3 +17,18 @@ Cypress.Commands.add("terminalLog", (violations) => {
 
   cy.task("table", violationData);
 });
+
+Cypress.on("window:before:load", (win) => {
+  cy.stub(win.console, "error", (msg) => {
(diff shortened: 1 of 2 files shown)


client/cypress/support/commands.js
```diff
@@ -17,3 +17,18 @@ Cypress.Commands.add("terminalLog", (violations) => {
 
   cy.task("table", violationData);
 });
+
+Cypress.on("window:before:load", (win) => {
+  cy.stub(win.console, "error", (msg) => {
+    if (msg.includes('No reducer provided for key "app"')) {
+      return null;
```
ktw1016, line 24: so when null is returned, does it skip over?

ktw1016 (approved):
Seems good to me

client/cypress/support/commands.js
sranc046: Yeah it'll ignore this specific error that we currently always have in our console

PR #1239, opened 2021-09-10
DisplayTable dropdown filter

### LINKED ISSUE #1223: Dropdown filter for display table
Screen Shot 2021-08-13 at 12 31 00 PM

Screen Shot 2021-08-13 at 12 29 12 PM

So I'm imagining when you click the arrow on the first pic, dropdown (like 2nd pic) would come down for filtering
This would be pretty cool

ktw1016:
closes #1223 

- [x] Filter Data placeholder text to be grey
- [x] Use react virtualized
- [x] Figure out right or left of dropdown
- [x] Experiment search within dropdown filter


ktw1016:
Example route that I was using: https://dev.ea-ad.ca/dp_dropdown_filter/index-eng.html#rpb/.-.-(table.-.-'orgSobjs.-.-subject.-.-'gov_gov.-.-columns.-.-(.-.-'*7b*7bpa_last_year_5*7d*7d.-.-'*7b*7bpa_last_year_4*7d*7d.-.-'*7b*7bpa_last_year_3*7d*7d.-.-'*7b*7bpa_last_year_2*7d*7d.-.-'*7b*7bpa_last_year*7d*7d))

Doesn't work for all rpb, we'll need to be selective where to use it for. Also need to go through other tables to check if it can be used anywhere

sranc046:
Just a thought, not sure which is best, but maybe the filter should be unbound from the table (speaking in terms of visuals)? For example, if you currently unselect all, the filter gets cut off by the height of the table. I think having the full length of the filter still visible could look good (maybe cut it off and have it scrollable if there are too many options)?

ktw1016:
For getting cut off, it's actually been already happening with column select button but yeah it should be improved nonetheless. I think we could add big padding "no data, check filter" if filtered data is empty. And also have it scrollable

sranc046:
Looks good! Last visual thing I would just tweak is when there is no data to filter, the first two columns stretch out leaving some awkward empty white space between the filter button and the edge of the column (not sure if I'm explaining that correctly...). Would it be possible for the columns to keep their usual width?

sranc046 (approved):
Looks good to me

ktw1016:
RPB (tables and columns that don't work):
- Authorities and Expenditures (vote/stat description)
- Spending by Program (program)
- Transfer Payments (transfer payment name)
- Tabled Estimates (vote/stat description)
- Full-Time Equivalents (FTEs) by Program (program)
- Program Expenditures by Authority Type (program)
- Program Expenditures by Standard Object (program)

Works well for all people tables

client/src/rpb/granular_view.js
```diff
@@ -67,6 +67,11 @@ class GranularView extends React.Component {
     } = this.props;
 
     const non_dept_key_cols = _.reject(sorted_key_columns, { nick: "dept" });
+    const dropdown_filterable_cols = _.reject(
+      sorted_key_columns,
+      ({ nick }) =>
+        nick === "dept" || nick === "prgm" || nick === "desc" || nick === "tp"
+    );
```
ktw1016, line 74: This is covering:
Authorities and Expenditures (vote/stat description)
Spending by Program (program)
Transfer Payments (transfer payment name)
Tabled Estimates (vote/stat description)
Full-Time Equivalents (FTEs) by Program (program)
Program Expenditures by Authority Type (program)
Program Expenditures by Standard Object (program)

ktw1016:
Once merged, I have more tables to use this for Service Inventory

Stephen-ONeil (changes requested):
How about always showing a filter icon for the filter menu button, displaying it as grey or just an outline when no filters are active? Communicates more than the down arrow in the default state

 - [ ] Current implementation leaves us two separate but overlapping column filtering options within the `DisplayTable`. Could we unify them? If we're taking cues from spread sheet programs, the search-based filtering is usually within, and applied to, the filter dropdown itself, not directly to the column. Does mean a lot more clicks to effectively get the same search-based filtering functionality, but from a "power" (spread sheet) user perspective it does give more options overall
   - bonus, in this case we don't need special logic/hardcoding for when to show what combination of search/dropdown based filtering. I think the "hidden logic" of the current approach (sometimes neither, sometimes search, sometimes search with dropdown) will just be frustrating to users
 - [ ] Alternative to making the functionality more coupled, we could make them more visually distinct and, importantly, make them exclusive (i.e. both can exist for a given column, but the user can only apply one or the other at a given time). Being able to apply both at once, as-is, is more confusing than useful IMO
   - even in this case, I'd prefer always showing both types of filter for every column over only sometimes showing the dropdown

ktw1016:
> * [ ]  Current implementation leaves us two separate but overlapping column filtering options within the `DisplayTable`. Could we unify them? If we're taking cues from spread sheet programs, the search-based filtering is usually within, and applied to, the filter dropdown itself, not directly to the column. Does mean a lot more clicks to effectively get the same search-based filtering functionality, but from a "power" (spread sheet) user perspective it does give more options overall
>   
>   * bonus, in this case we don't need special logic/hardcoding for when to show what combination of search/dropdown based filtering. I think the "hidden logic" of the current approach (sometimes neither, sometimes search, sometimes search with dropdown) will just be frustrating to users

This would impact performance, and also as you mentioned, usability. For performance, experimenting dropdown filter for columns in RPB cause big slow downs. For usability, to search for one specific item, a user will have to unselect all, then search, then select the item. But I think performance would be the biggest issue here

I personally don't think the search bar is confusing (searching the column vs within dropdown menu) because when the dropdown menu is open, clicking on the search bar closes the dropdown menu. That being said, I can understand the concern of confusion. How about putting another search bar within the dropdown menu and make it clear that this is searching within dropdown menu options?

I think the "hidden logic" is the biggest concern, I'd definitely prefer to fully unify them as you suggested for this problem but due to performance issue, I think we'll have to make them exclusive. I think making them exclusive could be even more confusing though. Like if a user has a dropdown filter applied, then after starts searching, resetting the dropdown filter seem odd and unnatural

Stephen-ONeil:
So I think we agree about some of the issues, just not how/if we can fix them. Putting distance between the simple (direct input) and advanced (dropdown, with or without it's own internal input filter) seems like it's not ruled out though, just needs to be explored/clarified further.

How about something, roughly, like this (with a good bit more style cohesion):
image
Leave the current, simple and probably more directly useful, filter as the default. Let users toggle a column over to the more advanced (but more clicks to use) spread sheet style filter. Avoids the confusing state where both are applying at once, and probably justifies enabling it for all filterable columns always.

Stephen-ONeil (changes requested):
So I think we agree about some of the issues, just not how/if we can fix them. Putting distance between the simple (direct input) and advanced (dropdown, with or without it's own internal input filter) seems like it's not ruled out though, just needs to be explored/clarified further.

How about something, roughly, like this (with a good bit more style cohesion):
image
Leave the current, simple and probably more directly useful, filter as the default. Let users toggle a column over to the more advanced (but more clicks to use) spread sheet style filter. Avoids the confusing state where both are applying at once, and probably justifies enabling it for all filterable columns always.

Stephen-ONeil:
Once toggled to "advanced" mode, the text input could be replaced with something like a dropdown menu with a summary label like "Showing X of Y values".

ktw1016:
So the problem we're solving is to either remove, or make it clear to the user that there is hidden filters applied. How about leaving the filter icon and dropdown menu as-is, but instead of "Use advanced filter" checkbox, simply a small grey italicized text below the line, "Filters applied" or something?

I just pushed a commit to showcase it, I think it's pretty neat

ktw1016:
Note: items down the list for react-virtualized list is not ctrl+F 'able

Stephen-ONeil (changes requested):
Mostly some more misc UI/UX thoughts, code all looks good (but I'll do one more pass after any of these you agree with have been done):
  - [ ] should it also display "Filters applied" when the stand alone search input is applying something?
  - [x] there's no on-hover style change to the filter icon (unless there's some active filter)
  - [x] I think Select All should only be checked if everything is selected, i.e. if you deselect anything else then select all should also switch to unchecked
  - [ ] the filter dropdown is a little annoying to use when the table is too short (in particular, you need to do some annoying scrolling immediately after de-selecting all items)
  - [ ] I know we talked about having the options in the dropdown also be filtered by the non-dropdown input, in addition to the dropdown-specific input. If items are filtered out of the table by the stand alone search, it doesn't seem useful to still show them in the dropdown list. Was there a reason not to do that? If we had one I've already forgotten it, haha
  - [ ] hmm, not set on it either way but I think the search inputs should be consistent on whether or not they use the magnifying glass icon. Adding it to the outer text input might help clarify that it's a text input, otherwise users might only try clicking the filter icon?
  - [x] last design thought I think, maybe the text for the stand alone input could be "Search this column" and the text input inside the dropdown could be "Search these filter options"? Just looking to make them more descriptive but also more consistent (right now one says search and one says filter, even though they do basically the same thing to their respective data; leaning towards calling it filter so that we can use that term for the dropdown menu/list of options itself)

ktw1016:
oops forgot to put awaiting for review label!

Stephen-ONeil (changes requested):
Still just the design comments/ongoing debates to settle, code's all good!

Stephen-ONeil (changes requested):
Some comments, couple bugs. I think it's stable enough to share with the team, up to you if you fix these first or not.
 
 - [ ] may have missed this comment, I slipped it in outside of a review, still applies I believe
> Oh, small thing, if you toggle off one item in the dropdown then toggle it back on a) the select all stays unchecked, b) the filters applied text stays displayed. I think select all shouldn't have it's own state, it should be derived from/control the other checkboxes.

Bugs:
  - [x] something's funky with the border radius on the tables
image
  - [ ] search column, open filter menu and interact with any check boxes, clear original search. The filtering doesn't properly clear (in fact, the search input no longer functions at this point)
  - [ ] doesn't always seem to be recalculating the virtualized list correctly after applying a search to the column (searching within the filter dropdown always seems to update it correctly though)
image

client/src/components/DisplayTable/DisplayTable.tsx
```diff
@@ -125,13 +159,19 @@ const get_default_state_from_props = (props: _DisplayTableProps) => {
...
+  );
+
   return {
     visible_col_keys,
     sort_by,
     descending: false,
     searches,
     initial_props: props,
     page_size: props.page_size_increment,
+    dropdown_filter,
```
Stephen-ONeil, line 174: Minor naming point, but I'd assumed `dropdown_filter` was a function. Something like `filter_options_by_column` would be more descriptive.

client/src/components/DisplayTable/DisplayTable.tsx
```diff
@@ -59,6 +62,9 @@ interface _DisplayTableState {
   initial_props: _DisplayTableProps;
   visible_col_keys: string[];
   searches: { [keys: string]: string };
+  dropdown_filter: {
+    [col_keys: string]: LegendItemType[];
```
Stephen-ONeil, line 66: These don't wind up passed to a `LegendList`, shouldn't be coupled to their type.

client/src/components/DisplayTable/DisplayTable.tsx
```diff
@@ -531,25 +586,66 @@ export class _DisplayTable extends React.Component<
...
+                                );
+
+                                this.setState({
+                                  searches: updated_searches,
+                                  dropdown_filter: {
+                                    ...dropdown_filter,
+                                    [column_key]: get_default_dropdown_filter(
+                                      col_configs_with_defaults,
+                                      data
+                                    )[column_key],
```
Stephen-ONeil, line 619: Why do we want to reset the state of the filter options when the out dropdown changes? Is that desired behaviour (specifically resetting `active` status)?

client/src/components/DisplayTable/DisplayTable.yaml
```diff
@@ -19,6 +19,25 @@ items_per_page_description:
...
+  fr: TODO
+search_filter_options:
+  en: Search these filter options
+  fr: TODO
+filter_data:
+  en: Filter data
+  fr: Filtrer les données
+no_data_table:
+  en: No data available for this table. Please check any filters applied.
+  fr: Aucune donnée n'est disponible pour ce tableau. TODO
```
Stephen-ONeil, line 40: Just a reminder to double check all these, make sure the wording's good, and get translation.

`select_all` is not used anymore pretty sure, can delete.

Stephen-ONeil (changes requested):
Questions and change requests.

Didn't comment on this one in-line, but the text use for the button to close the column filter dropdown is a little off. Maybe it's just the capitalization, but I also think it's not super clearly worded. Just saying "Close" would be clearer (since the user doesn't ever see the menu called "Filter menu" anywhere else, and may or may not make the connection quickly).
image

I think that's something in the default behaviour for the `DropdownMenu`, probably reusing the a11y title or something. I think you can either over ride that with custom close button text, but I didn't go looking at that components API my self.

Stephen-ONeil:
Out of scope, but I hate how much functionality is all crammed in to `DisplayTable` (I've also _always_ been annoyed by the name, haha). It's going to need a big cleanup and reorganization before we go adding any new features after this!

client/src/components/DisplayTable/DisplayTable.tsx
```diff
@@ -76,6 +82,7 @@ const get_column_config_defaults = (index: number) => ({
   is_sortable: true,
   is_summable: false,
   is_searchable: false,
+  show_dropdown_filter: true,
```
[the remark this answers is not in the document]
ktw1016: Left it in as a feature, not currently used. We could want it searchable but not have dropdown filter🤷‍♂️

client/src/components/DisplayTable/DisplayTable.tsx
```diff
@@ -96,13 +103,40 @@ type ColumnConfig = Partial<ReturnType<typeof get_column_config_defaults>> & {
   formatter: FormatKey | ((val: CellValue) => React.ReactNode);
 };
 
+const get_default_dropdown_filter = _.memoize(
+  (col_configs_with_defaults: ColumnConfigs, data: DisplayTableData[]) =>
```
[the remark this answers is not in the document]
ktw1016: Oh hmmm, wouldn't having `data` as first arg make caching sufficient? So that on the same set of data (which would mean on the same table), this would always be cached. When the table data changes, cache will change so it'll be fresh default dropdown filter with the new data.

client/src/components/DisplayTable/DisplayTable.tsx
ktw1016: Hmmm so my thought was that when a user checks off an item, then search for that in the outer search and if the item doesn't show up, that might be confusing.. But now that I think of it, filter being reset on outer search would be unexpected behaviour and would be more confusing. I'll remove this

client/src/components/DisplayTable/DisplayTable.tsx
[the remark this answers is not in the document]
Stephen-ONeil: Doesn't feel like a safe assumption to make. It's possible that either/or of the configs or data could be updated from outside the table (if not currently, then still possible in the future). Not doing this carefully right now could be a future (and hard to spot) bug.

client/src/components/DisplayTable/DisplayTable.tsx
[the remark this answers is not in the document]
Stephen-ONeil: I'd rather remove it. 

From the dev side, this component already has a bloated API, and this is a tricky one because it's meaningless unless is_searchable` is also true. 

From the user side, I don't want the inconsistency of sometimes having the filter button and sometimes not, even if it's just in one or two obscure cases. I can't think of a reason we'd disable it, which probably means if we ever have one it'll be something really obscure in which case the user probably would have no chance to work out our hidden logic for why the filter is missing. They'd just see it as a bug.

client/src/components/DisplayTable/DisplayTable.yaml
```diff
@@ -19,6 +19,22 @@ items_per_page_description:
...
+filters_applied:
+  transform: [markdown]
+  en: _Filters applied_
+  fr: TODO
+search_column:
+  en: Search column
+  fr: TODO
+search_filter_options:
+  en: Search these filter options
+  fr: TODO
```
Stephen-ONeil, line 31: Text translation reminder

Stephen-ONeil (changes requested):
Response to two comments from previous PR above, reminder abut text translation. Think that's it and we'll be good!

client/src/components/DisplayTable/DisplayTable.tsx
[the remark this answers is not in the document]
ktw1016: How would I go about writing the resolver? I can only think of stringify'ing everything and concatenating them but that can't be the most efficient way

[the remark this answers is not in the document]
Stephen-ONeil: That's most robust, but if you don't mind the extra maintenance burden then you can make it smarter by just keying on characteristics that you know change the output of to the memoized function (may or may not be more efficient, mostly better if different inputs can result in the same output when similar n known ways). Optimization often means sacrificing ease of maintenance/doing something hacky :shrug:. That's at least a localized problem, better than writing it in a way that could produce bugs if assumptions about other code are broken.

Is the memoized function particularly slow? Could it be optimized in any way?

client/src/components/DisplayTable/DisplayTable.tsx
```diff
@@ -124,6 +158,13 @@ const get_default_state_from_props = (props: _DisplayTableProps) => {
     .pickBy((col) => col.is_searchable)
     .mapValues(() => "")
     .value();
+  const t0 = performance.now();
+  const filter_options_by_column = get_default_filter_options_by_column(
+    data,
+    col_configs_with_defaults
+  );
+  const t1 = performance.now();
+  console.log(t1 - t0);
```
Stephen-ONeil, line 167: So how was the performance on the bigger RPB tables? Fine to drop the memoization?

Assume you meant to clean this up before flagging for review, haha.

client/src/components/DisplayTable/DisplayTable.tsx
```diff
@@ -96,13 +106,37 @@ type ColumnConfig = Partial<ReturnType<typeof get_column_config_defaults>> & {
...
+  (data: DisplayTableData[], col_configs_with_defaults: ColumnConfigs) =>
+    _.chain(col_configs_with_defaults)
+      .map((config, col_key) => ({ ...config, col_key }))
+      .filter((config) => config.is_searchable as boolean)
+      .map(({ col_key, plain_formatter }) => [
+        col_key,
+        _.chain(data)
+          .map(col_key)
+          .uniq()
+          .sort()
```
Stephen-ONeil, line 119: Only just caught this, but we need to extract and reuse the display table sorting logic here. A searchable column might be providing a custom `sort_func`, or just rely on a `plain_formatter` (by default "plain" formatted values are used for sorting).

Take the covid measure column in any covid inforgraphic table. The filter dropdown doesn't appear sorted (at least not alphabetically) because the raw column data is actually the IDs and it's the formatter's job to supply plain language names.

Stephen-ONeil (changes requested):
Cleanup, and a bug for edge case columns that provide custom sorting/rely on their plain formatters to be human readable.

client/src/components/DisplayTable/DisplayTable.tsx
ktw1016: Haha oops, lodash native caching without resolver is slightly faster but still fast enough

Stephen-ONeil (approved):
Approved, acknowledging that none of us have a satisfactory solution to #1309 at this time. Let's not sleep on the issue though, would like that fixed asap!

PR #1244, opened 2021-09-15
Search services

ktw1016:
TODO:
- [x] Typeahead "refreshing" when data loads
- [ ] Disable stemming ([here](https://docs.mongodb.com/manual/reference/operator/query/text/#stemmed-words) and [here](https://docs.mongodb.com/manual/core/index-text/))
  - I've tried ```  ServiceSchema.index(
    { name_en: "text", name_fr: "text" },
    { default_language: "none", language: "none", language_override: "none" }
  );```
  - `name_en: { ...str_type, index: true, language: "none" },
    name_fr: { ...str_type, index: true, language: "none" },`
  -  `Service.find({
          $text: { $search: `"${name_query}"`, $language: "none" },
        }),`
  - For e.g.: Try searching `agr` (No services), then `agri` (search properly). Also `publi` (no services), then `public` (search properly)
- I've tried using simple `.find({name: input_value})` but I can't get it to search as "contains" (I've tried regex). Case insensitivity and diacritics insensitivity should be do-able with [collation](https://docs.mongodb.com/manual/reference/collation/) but I haven't tried it with that yet

### DIFF
--- a/client/src/components/Typeahead/Typeahead.tsx
+++ b/client/src/components/Typeahead/Typeahead.tsx
@@ -75,12 +75,13 @@ export class Typeahead extends React.Component<TypeaheadProps, TypeaheadState> {
 
   constructor(props: TypeaheadProps) {
     super(props);
+    const { query_value, min_length } = props;
 
     this.menu_id = _.uniqueId("typeahead-");
 
     this.state = {
-      input_value: props.query_value,
-      may_show_menu: false,
+      input_value: query_value,
+      may_show_menu: query_value.length >= min_length,
       selection_cursor: default_selection_cursor,
     };
   }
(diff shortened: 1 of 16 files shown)


client/src/components/Typeahead/Typeahead.tsx
```diff
@@ -75,12 +75,13 @@ export class Typeahead extends React.Component<TypeaheadProps, TypeaheadState> {
...
     super(props);
+    const { query_value, min_length } = props;
 
     this.menu_id = _.uniqueId("typeahead-");
 
     this.state = {
-      input_value: props.query_value,
-      may_show_menu: false,
+      input_value: query_value,
+      may_show_menu: query_value.length >= min_length,
```
Stephen-ONeil, line 84: What's the case where this is necessary? Seems to imply there's a case where the Typeahead is initialized with a query active and we want the menu to open, without focus on the input?

Stephen-ONeil:
Commits merged via #1293, which finishes off the async search result loading. Forthcoming PR will work out the service mono-side searching.

PR #1265, opened 2021-10-08
eslint plugin jsx-a11y

ktw1016:
closes TBS-EACPD/infobase-archive#1533

### DIFF
--- a/client/src/components/WriteToClipboard/WriteToClipboard.js
+++ b/client/src/components/WriteToClipboard/WriteToClipboard.js
@@ -32,7 +32,6 @@ export class WriteToClipboard extends React.Component {
       button_description,
       IconComponent,
       icon_color,
-      title,
     } = this.props;
 
     const { copy_status_message, keyboard_navigation_detected } = this.state;
@@ -44,7 +43,7 @@ export class WriteToClipboard extends React.Component {
         <button
           className={button_class_name}
           onClick={() => {
-            debounced_log_standard_event("PANEL_LINK_COPIED", title);
+            debounced_log_standard_event("WRITE_TO_CLIPBOARD", text_to_copy);
             clipboard
               .writeText(text_to_copy)
               .then(() =>
@@ -86,7 +85,13 @@ export class WriteToClipboard extends React.Component {
             </Fragment>
           }
           subtitle={modal_active && !copy_success && copy_status_message}
-          body={modal_active && !copy_success && text_to_copy}
+          body={
+            modal_active &&
+            copy_success && (
+              /* eslint-disable jsx-a11y/no-noninteractive-tabindex */
+              <div tabIndex={0}>{text_to_copy}</div>
+            )
+          }
           dialog_position="left"
           auto_close_time={
             !is_a11y_mode && modal_active && copy_success && 1900
--- a/client/src/panels/panel_declarations/results/result_components.js
+++ b/client/src/panels/panel_declarations/results/result_components.js
@@ -324,59 +324,31 @@ const StatusIconTable = ({
   onClearClick,
   active_list,
 }) => (
-  <div>
-    <div className="status-icon-table">
-      <VisibilityControl
-        items={_.map(ordered_status_keys, (status_key) => ({
-          key: status_key,
-          active:
-            active_list.length === 0 ||
-            _.indexOf(active_list, status_key) !== -1,
-          count: icon_counts[status_key] || 0,
-          text: (
-            <span
-              className="link-unstyled"
-              tabIndex={-1}
-              aria-hidden="true"
-              data-toggle="tooltip"
-              data-ibtt-glossary-key={status_key_to_glossary_key[status_key]}
-            >
-              {text_maker(status_key)}
-            </span>
-          ),
-          aria_text: text_maker(status_key),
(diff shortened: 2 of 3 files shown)


client/src/components/WriteToClipboard/WriteToClipboard.js
```diff
@@ -83,10 +83,7 @@ export class WriteToClipboard extends React.Component {
             </Fragment>
           }
           subtitle={modal_active && !copy_success && copy_status_message}
-          body={
-            modal_active &&
-            !copy_success && <div tabIndex="0">{text_to_copy}</div>
-          }
+          body={modal_active && copy_success && text_to_copy}
```
Stephen-ONeil, line 86: Dropped a `!` on the `copy_success`. 

Also, I think this tab-able content was intentional because the purpose of this is to let a user manually copy some text when the clipboard access fails, I think a non-screen reader keyboard navigator needs to be able to focus on the text to easily copy just it?

I could be wrong on that, probably worth a little research.

client/src/components/glossary_components.tsx
```diff
@@ -45,7 +45,6 @@ export const GlossaryTooltipWrapper = ({
     <span
       className="nowrap glossary-tippy-link"
       style={no_bottom_border ? { borderBottom: "none" } : undefined}
-      tabIndex={0}
       data-ibtt-glossary-key={id}
```
Stephen-ONeil, line 48: Focusing this is necessary to open the tooptip.

client/src/components/modals_and_popovers/StatelessModal.tsx
```diff
@@ -96,7 +96,6 @@ export class StatelessModal extends React.Component<StatelessModalProps> {
           {body && <Modal.Body>{body}</Modal.Body>}
           {footer_content && <Modal.Footer>{footer_content}</Modal.Footer>}
         </div>
-        <div tabIndex={0} onFocus={this.closeModal} />
       </Modal>
```
Stephen-ONeil, line 99: I think this was important, so that "leaving" the modal (with any keyboard navigation) would close it and send focus back to wherever it had been before.

client/src/core/injected_build_constants.ts
```diff
@@ -24,4 +24,4 @@ export const cdn_url = typeof CDN_URL !== "undefined" && CDN_URL;
 
 export const local_ip = is_dev && typeof LOCAL_IP !== "undefined" && LOCAL_IP;
 
-export const services_feature_flag = false;
+export const services_feature_flag = true;
```
Stephen-ONeil, line 27: Typo for this branch, or were you testing how some change applied to the service inventory?

client/src/explorer_common/explorer_components.js
```diff
@@ -158,6 +157,7 @@ export const ExplorerNode = ({
         )}
       </div>
       <div className="ExplorerNode__ContentContainer">
+        {/* eslint-disable jsx-a11y/click-events-have-key-events */}
```
Stephen-ONeil, line 160: Worth listening to this one?

client/src/glossary/glossary.js
```diff
@@ -146,7 +143,7 @@ const Glossary_ = ({ active_key, items_by_letter }) => (
                   <div className="glossary-top-link-container">
                     <a
                       className="glossary-top-link"
-                      href="#"
+                      href="#glossary"
                       tabIndex="0"
```
Stephen-ONeil, line 147: This `tabIndex` isn't needed, I think it was there because an anchor with no href isn't focusable by default, probably from a time before this had the placeholder `href="#"`.

client/src/panels/panel_declarations/results/result_components.js
```diff
@@ -360,7 +360,6 @@ const StatusIconTable = ({
               <span
                 className="link-unstyled sr-only"
                 aria-hidden="true"
-                tabIndex={0}
                 data-toggle="tooltip"
```
Stephen-ONeil, line 363: Another case where it needs to take focus for the tooltip... but combining aria-hidden with a tabIndex is bad. If it remains focusable, it should at least not render in the text-only version

client/src/components/modals_and_popovers/StatelessModal.tsx
ktw1016: Hmm, I'm not sure if tabbing to leave modal should be the correct behaviour, especially because user would've had to purposely tab-past "cancel" button.

Without this div, tab goes to address bar, then comes back into the modal. User has to either press esc or trigger cancel to leave modal. Focus goes back to where it should go regardless

But I'll just disable rule and leave the div if you prefer!

client/src/panels/panel_declarations/results/result_components.js
ktw1016: Hmm actually, what is the purpose of this table anyways? This seems totally unnecessary

If this is for a11y, it should only be displayed in a11y mode, and also if anything, it should be incorporated into `VisibilityControl` component. But actually checking with screen reader, screen reader can't even read this properly because `aria-hidden="true"`. Not tab navigable either, and screen reader seems to be reading `VisibilityControl` component just fine. 

So I suggest that we remove this `<table>` entirely

PR #1266, opened 2021-10-08
Cypress update

Steph-Rancourt:
Fixing an issue left in the cypress tests and upgrading it

### DIFF
--- a/client/cypress/integration/InfoBase/route_tests.spec.js
+++ b/client/cypress/integration/InfoBase/route_tests.spec.js
@@ -254,26 +254,24 @@ describe("Route tests", () => {
             `http://localhost:8080/build/InfoBase/index-${app}.html#${routes.route}`
           );
 
+          cy.get(".leaf-spinner__inner-circle", { timeout: 10000 }).should(
+            "not.exist"
+          );
+
+          //Once basic routes a11y critical issues are fix, can remove this if statement
           if (!routes.skipAxe && (app == "eng" || app == "fra")) {
             cy.injectAxe();
-            cy.get(".leaf-spinner__inner-circle", { timeout: 10000 }).should(
-              "not.exist"
-            );
             cy.checkA11y(
               null,
               { includedImpacts: ["critical"] },
               cy.terminalLog,
               false
             );
-          } else {
-            cy.get(".leaf-spinner__inner-circle", { timeout: 10000 }).should(
-              "not.exist"
-            );
           }
 
           cy.on("fail", (e, test) => {
             if (routes.expect_to_fail) {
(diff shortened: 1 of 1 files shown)


client/cypress/integration/InfoBase/route_tests.spec.js
```diff
@@ -247,38 +247,37 @@ const route_load_tests_config = [
...
+            );
           } else {
             cy.get(".leaf-spinner__inner-circle", { timeout: 10000 }).should(
               "not.exist"
             );
           }
+
+          cy.on("fail", (e, test) => {
+            if (routes.expect_to_fail) {
+              console.log("Test expected to fail.");
```
Stephen-ONeil, line 276: Might be worth logging what it failed with (not re-throwing, just as a log). If those always-failing routes are still flaky then that'd help us figure out why

client/cypress/integration/InfoBase/route_tests.spec.js
```diff
@@ -247,38 +247,37 @@ const route_load_tests_config = [
...
+            cy.injectAxe();
+            cy.get(".leaf-spinner__inner-circle", { timeout: 10000 }).should(
+              "not.exist"
+            );
+            cy.checkA11y(
+              null,
+              { includedImpacts: ["critical"] },
+              cy.terminalLog,
+              false
+            );
```
Stephen-ONeil, line 267: Is tit necessary to inject axe, make the leaf spinner check, and then do the a11y check in that order? Just wondering if the common `cy.get...` between the two branches of this `if` could be moved up right after the `cy.visit` so that
  1. the operation isn't duplicated and we don't risk having them go out of sync
  2. so that the `cy.get` part runs under the same conditions whether or not the axe checks also runs after (don't know if the `cy.injectAxe()` call could impact the spinner test, but seems worth being cautious)

client/cypress/integration/InfoBase/route_tests.spec.js
```diff
@@ -247,38 +247,37 @@ const route_load_tests_config = [
...
-              );
-              cy.checkA11y(
-                null,
-                { includedImpacts: ["critical"] },
-                cy.terminalLog,
-                false
-              );
-            }
+
+          if (!routes.skipAxe && (app == "eng" || app == "fra")) {
```
Stephen-ONeil, line 257: Can you remind me why axe is skipped in both basic versions right now? Probably worth capturing in a comment directly on this condition, along with a note on what needs to be done before it can be run on all versions (an issue would be good too, if there isn't already one)

Stephen-ONeil (approved):
:+1:

PR #1267, opened 2021-10-09
Wrap the GQL client's fetch in retry_promise

Stephen-ONeil:
TODO:
  - [x] add `retry_promise` to the `fetch` used by the GQL client
  - [x]  `retry_promise` could use a few more features
    - [x] some growth factor on the intervals (probably exponential)
    - [x] maybe an option for an upper limit on the delay, to reign in the growth factor  
    - [x] random jitter on each timeout (... well, `setTimeout` naturally has some uncertain delay while the event loop works around to its callback, but some option to control an additional amount is easy enough)
  - [x] consider the impact on client logging. As is, the total request time will include the retries, but the logging won't know if, or how many, retries occurred. Maybe that's fine, or maybe that will throw off future usage of the analytics, hmm
    - shuffled things around, managed to get retry count info in to the logging

In September, 48 API queries errored (of 32773). 26 of those were failures to even connect to the server, a handful occurred after the initial preflights, and a few were actually in the processing of the received data (TODO: look in to those more). Small portion, but every one of them is unhandled and results in an error page for a user. Currently our highest source of them by far.

These aren't new! The trend goes back further, monthly API error counts on this order of magnitude have been long standing. What's new (as of mid August) is that unhandled error rejections actually hit the error boundary (overdue on that, oops). Before these probably resulted in endless spinners for users, now they result in the full error page. They also now end up in the primary error report, instead of only showing up in the specialized API error reports, which is admittedly the only reason I'm not noticing them.

Anyway, not enough information to positively diagnose the cause(s), and unlikely to be able to remedy them ourselves (e.g. if they're environmental on the user's end or across the network). Best I can tell, it's not 100% of requests failing for the impacted users. For now, letting the requests retry will hopefully knock out a lot of those error-screen producing cases.

### DIFF
--- a/client/src/InfoBase/common_app_component_utils.js
+++ b/client/src/InfoBase/common_app_component_utils.js
@@ -4,7 +4,7 @@ import React from "react";
 import { log_standard_event } from "src/core/analytics";
 import { cdn_url, is_dev } from "src/core/injected_build_constants";
 
-import { retry_promise } from "src/general_utils";
+import { retrying_promise } from "src/general_utils";
 
 // Link tags for stylesheets should all have non null sheet properties
 const linked_stylesheets_loaded = () => {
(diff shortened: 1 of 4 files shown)


Stephen-ONeil:
> Is the changes in `populate_results` unrelated? Anyways LGTM!

Tangential, yeah, see 029e8c3

Issue #1253, opened 2021-09-27
Add glossary side menu


sranc046:
Example from USASpending: https://www.usaspending.gov/disaster/covid-19/?publicLaw=all. Clicking on glossary term will bring up the side menu.

The menu will allow users to browse the whole content of the glossary without losing their place on the InfoBase.

Stephen-ONeil:
With the right focus management, this would also address #1016 (although #1230 might beat this idea to the punch).

Issue #1240, opened 2021-09-11
Potentially applicable popper.js optimization


Stephen-ONeil:
See [this ](https://atfzl.com/don-t-attach-tooltips-to-document-body) article (worth a read from everyone, even if turns out to not be applicable).

The article's author notes that the source of the performance issue could be fixed on popper's side, and that the author's experience was with an older version of popper. Possible it was, or has since (article's from March), been corrected for up stream. Double check that, and also double check that popper.js is up to date on our end.

Issue #999, opened 2021-04-07
Tooltip arrows not showing in IE


GarvinH:
related #997 - if it can be fixed without using the `display: inline-block` might as well remove it 

For some reason the tooltip arrow (below) doesn't work on IE:
![image](https://user-images.githubusercontent.com/25855114/113925259-d0ee8000-97b8-11eb-84ea-478692689f78.png)

My guess:
It's not working because of pseudo-selectors (ie `::before`).  When viewing the `.tippy-arrow` styles, it will show all the `::before` classes as crossed out INCLUDING the style `content: ""`. Previously when I was replacing the old tooltip with tippy, I noticed that without `content: ""`, the arrow will not display!

Give it a try in chrome by disabling the `content` style, the arrow should disappear.

I've tried one of the methods online suggesting to add `display: inline-block` but it doesn't do anything 😕

PR #1270, opened 2021-10-15
Glossary sidebar

sranc046:
#1253

- [x] Separate into appropriate components
- [x] Small screen support
- [x] Search terms
- [x] Link words throughout app
- [ ] Plain vs official definition (to be discussed)


Stephen-ONeil:
Should close #1016 too.

Stephen-ONeil (changes requested):
Review comments from on-site testing, will follow up with code feedback separately. 

- [x] on hover cursor for `.glossary-tippy-link` should show a pointer, to communicate click interactivity
  - [ ] while you're at it, look at renaming/possibly reorganizing any classes (or other) references to tippy
- [x] make sure the additional term items like these also work as expected 
image
- [x] I think a back arrow icon would help this navigation stand out better. I feel like it needs something more at least
image
- [x] think about the behaviour of the menu's search bar when an item is active. Currently it doesn't function. Should it not be present if the user isn't viewing the full list? Or should editing it while viewing a term return the user to the full list with updated results? In that second case, when opening the menu by clicking on an item, probably the input should start populated with the clicked term
- [x] I can't seem to scroll the glossary list all the way down, stuck just above the W's
image
- [ ] haven't tested with a screen reader, but normal keyboard navigation is unable to access the glossary "list". It's made of divs that can't take focus. Proper HTML syntax is important, make it an actual list and use unstyled buttons or possible anchors for the items (will dig in deeper on proper a11y in a later review)
- [x] minor quirk, the `.glossary-title` class used on the menu's header has a :hover underline, assume this is a class that had some other purpose and was just reused here
- [x] lots to fix with layout and positioning here
image

Just a reminder that the handful of non-glossary tooltips, like these from the covid authorities panel, need alternative tooltip functionality if we're dropping the tooltip activator, tippy, etc in this branch. I'm in favour of dropping that junk, so I think once this branch is closer to being finished I'll make a sub-branch to tackle these remaining tooltips myself. Might be able to push through a replacement with simpler footnotes or something.
image

client/src/glossary/GlossaryMenu.tsx
```diff
@@ -0,0 +1,149 @@
...
+              </div>
+            </div>
+          </div>
+          <div className="glossary-sidebar-content-wrapper">
+            <SidebarContent
+              title={this.props.item.title}
+              def={this.props.item.def}
+              results={this.state.results}
+              closeItem={() => this.closeItem()}
+              openItem={(item: Record<string, unknown>) => this.openItem(item)}
```
Stephen-ONeil, line 142: RE suggestion to just pass around glossary keys around rather than `item`s: here's an example of where that'd make things easier. Simple string type, instead of needing (I assume just for dev expediency) a mystery record.

Stephen-ONeil (changes requested):
Scatter shot of suggestions, a couple questions. Haven't looked at or commented on every single line since it's just a mid-point review and work is ongoing. Probably worth a follow up mid-point review once the majority of this, and the previous visual comments, are addressed. Let me know any time you want that!

ktw1016 (commented):
Some of my thoughts:

- "Search for a term used in GC InfoBase" seem too long? Can't read the full text. How about "Search for a definition" or just "Search for a term"?
- `glossary-letter` should be more emphasized, probably just bolding it is fine
- `item-title` should also be more emphasized, I'd definitely bold it. But could it even have different background color or a font color? Can't think of a good combination but that'd be nice
- Any reason why "Glossary" button doesn't appear right away before scrolling? And also on home route? I'd prefer that button to always be there regardless on route/scrolling
- "Back" button:
  - It's hard to guess what this button is for after clicking on a glossary item in text. Should be reworded to something like "See the full glossary"? or "See other definitions (or terms)" something along those lines, I think
  - How about giving different background color (primary color) to all area below the definition, so all area including and below back button? It'd give more separation between the back button and definition and visually can emphasize both.. I think, but it could look messy
- When pressed "Back", scrolling to where that definition is, would be nice

ktw1016 (commented):
I kept looking and thought of few more lol
- When clicking "Glossary" button, should it always open the full list of definitions? Or a definition it was previously on? I'm leaning towards always full list of definitions. Something to think about
- for gl_tt items: on Enter, it doesn't open the sidebar. Hence the screen reader doesn't work either

ktw1016 (commented):
I think it mostly looks good on the small screen, just one small thing, I think "Glossary" button should be small button to the right side of the screen. Right now, it's taking full screen width on mobile screen.

Still far way to go on a11y:
- Need a11y text on `gl_tt` item, "Open to see the definition" or something.
- Also, when clicked, focus needs to shift to the glossary definition.
  - Similarly, when definition is clicked from "Full list", focus should go to definition
- Will probably need invisible, screen reader purpose "close glossary sidebar" button after the definition
- Horizontal splitter should prob be a11y hidden

Stephen-ONeil:
Directly closes #1253

By removing the old tooltips, closes #1240, closes #1016, and closes #999

client/src/glossary/SideBarActivator.js
```diff
@@ -0,0 +1,133 @@
...
+import { BackToTop } from "src/components";
+
+import { GlossaryMenu } from "./GlossaryMenu";
+
+const ROUTES_WITHOUT_GLOSSARY = {
+  "/start": true,
+  "/glossary": true,
+};
+
+const SidebarActivator = withRouter(
```
Stephen-ONeil, line 16: The "activator" in the `TooltipActivator` came from it's (hacky) job of detecting and "activating" newly rendered tooltip targets, since we needed to run code to initialize them as they were inserted in to the DOM. Hard to give this a specific name because it's both the event handler for all glossary links AND the glossary menu itself at the same time. Maybe something like `GlossaryMenuSystem` to indicate that it's more than one thing ("system" isn't great, this is just a suggestion).

client/src/glossary/SideBarActivator.js
```diff
@@ -0,0 +1,133 @@
...
+          <BackToTop
+            focus={() => {
+              document.querySelector(".glossary__search-bar > input").focus();
+            }}
+            handleClick={() => {
+              this.setState({ showGlossary: true });
+            }}
+            text={"glossary_button"}
+            showWithScroll={false}
+          />
```
Stephen-ONeil, line 126: Ah, this is at odds with my upstream changes to the `BackToTop` component, haha. A little hacky to be using it this way though. Maybe extract the parts of `BackToTop` responsible for the floating button styles/catching on the header footer/etc in to something that can be shared between `BackToTop` and this use case (the fact both move focus is a bit coincidental, but I think a better way to move focus to the input bar will be possible when things are reorganzied).

client/src/glossary/GlossarySidebarContent.tsx
```diff
@@ -0,0 +1,213 @@
...
+      this.props.results.length == 0
+        ? glossaryEntryStore.get_all()
+        : this.props.results;
+
+    const glossary_items_by_letter = _.chain(glossary_items)
+      .groupBy((item) => {
+        const first_letter = item.title[0];
+        if (_.includes(["É", "È", "Ê", "Ë"], first_letter)) {
+          return "E";
+        }
```
Stephen-ONeil, line 127: `_.deburr`, it's a miracle worker for us

client/src/glossary/GlossarySidebarContent.tsx
```diff
@@ -0,0 +1,213 @@
...
+        case "open":
+          if (item) {
+            this.openDefinition(item);
+          }
+          break;
+      }
+    }
+  }
+
+  get_glossary_items_by_letter() {
```
Stephen-ONeil, line 116: This code looks like it's copied from `glossary.js`, right? Extract it to `src/glossary/glossary_utils.ts` (along with any other common code)

Stephen-ONeil (changes requested):
Haven't gotten to the search config related stuff in this round, but let's start by going through the `SideBarActivator`/`GlossaryMenu`/`GlossarySidebarContent`.

Start by reading [this](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0) classic. Don't take the note about not taking it too seriously _too_ seriously, hahaha. It's still a very useful way to think about things when building complex systems of components like this (but don't take it as dogmatic, it's just a helpful tool). I'm going to use the terms "brain" and "face" components, because I think it's funny to keep expanding the list of alternate lingo for this. 

Next, refresh on the [single-responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle) (and apply the same warning about not taking it as dogma, again, it's just a tool).

Look over those components again and think about what's brain, what's face, and what's both. Also think about what each component is responsible for (and where the responsibilities might be tangled up between them).

To be honest, I haven't worked any of that out fully my self. I think it'll be much easier for you, as the original author, though. Still, I'm pretty confident that basically all of the state/brains could be pulled up in to what's currently called the `SidebarActivator`. I also think everything else could be sorted out in to simple, single responsibility, face components. 

Importantly, I think most, if not all, of those face components could be direct children of the brain part and lose their state. I believe most of the state/smarts of each component is to figure out _what_ needs to be rendered, _when_ to rerender, and _whether_ that rerender is enough of a change to do some on-change behaviour. I think the brain should have all the information to decided what to render, and I think rerendering/on-change behaviour will naturally sort itself out to only on-mounts (because I think the flow will work out to be, once everything's single-responsibility and flat, that those cases will be unmounting/remounting between changes instead of persisting statefully in multiple roles).

Could be wrong, would basically have to start refactoring it my self to test these assumptions. Possibly over-optimistic, maybe there's some complexity I missed that means the ideals don't apply here (remember, both of those linked resources are tools, not rules). Even in that case though, aiming for the ideal here should be a good start on reorganizing it! 

Let me know if you want further clarification or to pair program on this or anything.

Stephen-ONeil:
Meta-comment: just went back to the previous review to try and figure out what comments had been resolved/were actually outdated vs ones that were just flagged as outdated because a lot of code had changed since. I think everything was addressed or no longer applicable. Just wanted to clarify, whoever left the review comment is in charge of marking it resolved (so they have a chance to review that what they meant to communicate got across), BUT you can do a lot to help out with that by leaving a comment, or even just a thumbs up, on ones that you believe you've addressed. I'd appreciate it!

Also, a lot of them were questions. I marked them resolved because they no longer applied with the new code, but to be clear I'm not asking those rhetorically, haha. Please respond to them, I might have had a piece of follow up advice depending on the answer (or, pretty often, I was just actually curious).

client/src/glossary/GlossaryMenu.tsx
```diff
@@ -0,0 +1,191 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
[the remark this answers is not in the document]
sranc046: These just happens when I'm first testing out some code and gradually work out the types, probably just forgot to re-enable, my bad

client/src/glossary/SideBarActivator.js
sranc046: Hmm okay yeah, makes sense

client/src/glossary/GlossaryMenu.scss
```diff
@@ -0,0 +1,191 @@
...
+    border: 2px $tertiaryColor solid;
+    border-radius: 1rem;
+    height: 3.5rem;
+    margin-top: 10px;
+    padding: 0 1rem;
+    background: $backgroundColor;
+
+    & > input {
+      width: 100%;
+      font-size: 16px;
```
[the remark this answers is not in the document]
sranc046: There were some conflicting styles and an extra unused class leftover but yeah the idea is that input is (now) inside .glossary-sb__search-bar

client/src/glossary/SideBarActivator.js
```diff
@@ -0,0 +1,133 @@
...
+
+    handleWindowKeyDown = (e) => {
+      if (e.code === "Enter") {
+        this.handleWindowClick(e);
+      }
+    };
+
+    componentDidMount() {
+      window.addEventListener("click", this.handleWindowClick, {
+        capture: true,
```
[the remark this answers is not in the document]
sranc046: Not sure I understand your question... but the use of capture was to deal with an order of events issue in the DOM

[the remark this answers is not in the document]
Stephen-ONeil: I was just curious because capturing phase event listeners are pretty rarely used. Was curious why it was used here (assumed maybe you saw an article recommending them for the event delegation pattern or something). What was the specific bug?

Capturing's usually not recommended. Practically, that's because it _is_ uncommon; devs will be thinking of events in terms of bubbling by default so it can be confusing when debugging down the line. More theoretically, it's not recommended because it gives priority to handlers further away from the actual target which is less logical than bubbling. 

Doesn't mean you can't use it, but maybe treat it as a "hack" and leave a comment here on why it was necessary (so that  future devs don't blindly copy the pattern or anything).

client/src/components/SidebarButton/SidebarButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+import { FloatingButton } from "src/components/FloatingButton/FloatingButton";
+
+interface SidebarButtonProps {
+  open_sidebar: () => void;
+  left: boolean;
+  focus: () => void;
+}
+
+export class SidebarButton extends React.Component<SidebarButtonProps> {
```
Stephen-ONeil, line 13: Doesn't seem useful as-is. If the caller has to provide a `focus` call back anyway then it might as well just handle focusing from the `open_sidebar` callback. Outside of that it's just `FloatingButton` with some renaming and enforced defaults.

I suggested a component for this, but what I had in mind was for it to be responsible for all "sidebar" behaviour. Interface for the simple, stateless, version could look something like this
```
const Sidebar = ({is_open, close_callback, children}) => /* 
  Sidebar is responsible for:
    - standard styling of the sidebar
    - a close button
    - hiding it when not open
    - moving focus in when opening (returning focus, if necessary, is the parent's job via the `close_callback`). 
    
  The parent is responsible for:
    - holding the open state
    - some arbitrary control to open the sidebar
    - supplying arbitrary children to render inside the sidebar
    - potentially handling focus return on close
 */
```

A separate more special case `SidebarButton` component could extend that stateless version, taking on the responsibility for everything but the the actual child content.

Stephen-ONeil: Out of scope for now though, can drop this component in this PR and save that work for whenever we need a separate sidebar (... so probably right away, just in the branch for the table of contents).

Stephen-ONeil (changes requested):
First chunk of review, will continue after lunch, just giving you some comments to start in case something comes up and I don't get through the rest today.

Side note: I left a response on one of the previous comments within the hidden block... not sure GitHub will do a good job of letting you know about it, so if you haven't seen it yet you'll have to expand those hidden comments to find it :roll_eyes:

ktw1016 (changes requested):
A few more screen reader tingz:
- After pressing "Full list", focus should go back to where the definition was at
- Should have an invisible button for screen reader at the end (after "full list"), to close the side bar
- Any thoughts on rewording "Full list" to "View Glossary"? (Glossary itself is a "list of definition")

client/src/glossary/GlossaryMenuController.js
```diff
@@ -0,0 +1,156 @@
...
+    itemClick = (e) => {
+      const target = e;
+      const glossary_item_key = target.dataset.ibttGlossaryKey;
+      const glossary_item = glossaryEntryStore.lookup(glossary_item_key);
+
+      this.setGlossaryItem(glossary_item.id);
+      this.setList(false);
+
+      this.toggleGlossary(true);
+      document.querySelector(`.glossary-sb__search-bar > input`).focus();
```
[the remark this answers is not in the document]
Stephen-ONeil: Ah, wait, I see why this is tricky. That glossary search bar is rendered via the typeahead and is located in a totally different part of the code base...

client/src/glossary/GlossaryMenuController.js
```diff
@@ -0,0 +1,156 @@
...
+  class GlossaryMenuController extends React.Component {
+    constructor(props) {
+      super(props);
+      this.state = {
+        show_sidebar: true,
+        showGlossary: false,
+        glossaryItem: null,
+        showList: true,
+        results: [],
+        query: "",
```
Stephen-ONeil, line 29: Some of this state (`results` at least, maybe `showList`) can be derived inside child components rather than existing as top-level state here, I believe. Don't think `results` needs to be state _anywhere_, the child components can just derive it from the `query` value (as long as that's memoized, no problem at all).

client/src/glossary/GlossaryMenuController.js
```diff
@@ -0,0 +1,156 @@
...
+          <GlossaryMenu
+            show={this.state.showGlossary}
+            toggle={(value) => this.toggleGlossary(value)}
+            item={this.state.glossaryItem}
+            setGlossaryItem={(key) => this.setGlossaryItem(key)}
+            showList={this.state.showList}
+            setList={(value) => this.setList(value)}
+            setResults={(data) => this.setResults(data)}
+            setQuery={(query) => this.setQuery(query)}
+            results={get_glossary_items_by_letter(this.state.results)}
```
Stephen-ONeil, line 140: Grouping results by letter should be `GlossaryMenu`'s responsibility (or this prop should have a name like `resultsByLetter` or something, but that's more clumsy).

Stephen-ONeil (changes requested):
- [x] lots of sidebar specific glossary components directly under `glossary/` (with kind of generic names). Better organization to move them to a sub-directory (e.g. `glossary/GlossarySidebar/`), maybe with an index file that specifically only exports the intended entry component (since a lot of them are specifically sub-components not to be used directly) IMO
  - [x] I'm having a bit of trouble untangling search related changes/additions
    - [x] is `SearchConfigSidebar.js` vestigial/can be deleted?
    - [x] otherwise, I'd like to get on a call some time to work though this and maybe do a bit of pair programming on it. Let's link up on slack once you've got some of the other feedback addressed

Stephen-ONeil (changes requested):
Edit, oops, hit shift + enter when I wanted a newline, ha, posted directly. Ignore this.

Stephen-ONeil (commented):
Edit: wow, I did it again, my new line entry has been miss-trained by slack! I'm going to just edit this to list out other mix bugs, and then get to code comments. The comment editing view gives me a wider display than the PR comment box anyway, so let's pretend this was intentional :laughing: 

Misc bugs/design considerations:
  - [ ] tab navigation enters glossary side menu content right after tabbing out of the header. No focus rings anywhere, takes a number of tab presses to get out of it and down to actual page content
    - to start, it shouldn't be reachable that early in the tab order, if directly navigable then it should probably be near the site footer in order
    - tab navigation should directly land on the glossary menu button, with some clear visible indication (default should be fine, since it's a button)
    - relevant prior comment: https://github.com/TBS-EACPD/infobase/pull/1270/files#r780360661
  - [ ] another tab navigation note, once the menu is opened I think it should act like a modal and trap focus. Right now you can tab out of the open menu back in to the site header. I think it's better user experience to require key board navigators to close the menu with the X button to get back out (so that we can then manage returning focus to where it came from originally, i.e. either the button or some glossary target in the page content)
  - [ ] opening the glossary side menu from glossary target text should move focus in to the glossary menu. Also the above not on returning focus to where it came from afterwards
    - hmm, might be a tricky combination of focus management. Maybe only manage focus this way when the target text is interacted with via a keyboard event? Otherwise, since a mouse user will expect to be able to continue to interacting with the page while the side bar is open, this focus management could be awkward for them
      - ... actually, it closes on any mouse click events outside of it, so the only awkward case is that mouse users can still scroll the window (and a keyboard user could do this too, e.g. page up/down). The specific behaviour to be careful with is if the original target is scrolled in to view when focus is returned
  - [ ] Should the glossary menu close when a user clicks on content outside of it? If so, maybe it actually should be a modal rather than a sticky sidebar? The sticky sidebar, on significantly wide screens, has the benefit of allowing the user to keep navigating the site with the menu open, but if any clicks outside the menu close it then they're limited to just scrolling which isn't too much value
    - [ ] related, _if_ we keep the close on outside click behaviour, it should be for mouse down only I think. A mouse up outside the menu could result from an interaction that began inside it (e.g. selecting text inside the menu, but releasing the click after moving the mouse outside)
  - [x] the position of the glossary menu button doesn't respond well to screen resizes (I don't believe this is the same as the general need for better display on mobile devices, this is specifically about the positioning of the floating glossary button when a screen is resized mid-session)
  - [x] searching something with no matches does nothing (continues showing the full list), feels more like a bug than something intentional (whether or not it is). If nothing matches, should display that as a message instead of the full list
  - [x] search field input isn't reapplied when returning to the list after viewing an item. Have to edit the search input again to reapply it. Search input should be known and used consistently by the list

Stephen-ONeil (changes requested):
I've probably not covered everything I could at this point but between these specific code comments and the bugs/missing features listed above I think this is more than enough for now.

If there was anything specific you wanted me to help with that I've missed, just let me know!

I think we should discuss ways to move this along, try and split up the work or something. Didn't suggest it earlier because this isn't a priority, but the longer it's open as a PR the harder to review/move forward it's going to get!

client/src/components/DebouncedTextInput/DebouncedTextInput.js
```diff
@@ -32,8 +31,7 @@ class DebouncedTextInput extends React.Component {
         id={unique_id}
         type="text"
         aria_label={a11y_label || trivial_text_maker("text_input")}
-        className={classNames("form-control", inputClassName)}
-        style={style}
+        className={classNames(inputClassName)}
         placeholder={placeHolder || ""}
```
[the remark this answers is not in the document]
sranc046: I was certain I had made the proper adjustments, I think they got lost in a rebase somewhere

client/src/components/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,86 @@
...
+              </aside>
+            </div>
+          </CSSTransition>
+        </FocusLock>
+        <FloatingButton
+          button_text={button_text}
+          showWithScroll={false}
+          handleClick={() => this.props.callback(true)}
+          tabIndex={0}
+        />
```
Stephen-ONeil, line 82: Is a floating button inherent to a sidebar? We might not have a use case, right now at least, for a sidebar that is opened by something other than a floating button, but I don't think the base `Sidebar` component should package them both. 

PR's old enough that it's getting hard to keep track of my own past advice, haha, but I believe you may have done this in response to my comments on the previous `SidebarButton`. I was saying there that something called `SidebarButton` should include a sidebar in addition to a button, and probably also manage it's own state. I also said, though, that I didn't think we needed that component to exist right now. I didn't mean to suggest that it be pushed down in to `Sidebar` itself though, I was thinking that rendering the `FloatingButton` that controls the glossary sidebar should be `GlossarySidebarController`'s responsibility.

A common pattern you'll see with our components is to exist in two versions. One stateless version that's just presentation and as much inherent behaviour as can be encapsulated without needing to own too much state, and one version with state, and possibly additional controls, that is a fully usable implementation. We sometimes call the second type "auto" as opposed to "stateless", but I don't think I've always held consistent on that. The point of this approach is to have both a ready to use/reference implementation for the most default use case, as well as a version that gives up more control/responsibility to the consumer for non-standard use cases.

Circling back, the more flexible, stateless, `Sidebar` shouldn't render its own control. That should be something the consumer can pick/is responsible for. If you want to/anticipate us using an auto version, then you can make that a separate module that imports the stateless `Sidebar` and packages it with a `FloatingButton` (and self-managed open/closed state). Since the glossary sidebar use case needs special behaviour, opening on glossary link click, it won't be using the auto version either way.

Stephen-ONeil: While I'm at it though, there _is_ some code in `GlossarySideBar` controller that should belong to even the stateless `Sidebar`, I believe. I think we consider closing on click outside to be inherent to a sidebar, right? The stateless version should probably manage window click events and call it's closing callback as necessary. Pushing that inherent behaviour down to this level will save us duplicating it across all sidebars going forward.

If we decide there's a sidebar that _shouldn't_ have that behaviour in the future, then just add a prop to disable it. Still better for it to be something encapsulated at the lowest level.

Stephen-ONeil (changes requested):
Getting there! You might want to fix any more obvious dev link notes/bugs below first, but I think you should present this at the next team meeting and share it around for broader feedback :+1:. Further code comments can mostly wait, in case feedback leads to any significant chnages.

Dev link review:
  - [ ] while the glossary side bar is open, clicking and dragging to select text doesn't work
  - [ ] after opening the glossary from a "glossary link" and then closing it:
    - [ ] 1) focus isn't returned to clicked link, just scrolling (focus necessary, scrolling is potentially not wanted; tricky to balance the keyboard navigation vs mouse navigation behaviour)
    - [ ] 2) the glossary menu continues to scroll to that item in to view on close, on subsequent open/close toggling via the direct glossary button
  - [ ] right clicking on the glossary link also opens the menu?
  - [ ] there's some empty divs with the attribute `data-focus-guard` in the DOM right above the glossary button, at least one of these is tab-focusable despite being empty. Took a quick look, they seem to come from the `react-focus-lock` library, probably shouldn't render the `FocusLock` component in `Sidebar` when the sidebar is closed
  - [ ] it's possible for glossary entries to contain links, poor contrast against the light blue background. The only example of this at the moment is the entry for "Gender Based Analysis Plus (GBA Plus) Indicators", to be fair

client/src/glossary/GlossarySidebar/GlossarySidebarController.tsx
```diff
@@ -0,0 +1,170 @@
...
+      const currentPage = this.props.location.pathname;
+
+      if (routes_without_glossary.includes(currentPage)) return null;
+
+      return (
+        <>
+          <GlossarySidebar
+            glossary_item_key={this.state.glossary_item_key}
+            focus_item_key={this.state.focus_item_key}
+            set_glossary_item={(key) => this.setGlossaryItem(key)}
```
Stephen-ONeil, line 141: The old advice (still applies in classes, but function components with hooks just give up on it far as I know) is to avoid call backs and event handlers that are declared inside a component's `render`, so that React can tell when those props actually are changing. Inlined arrow functions are re-declared on each render! The pattern here should just be `set_glossary_item={this.setGlossaryItem}` etc.

Stephen-ONeil: This occurs across other files in this PR, so (as always) do your own pass after addressing the specific things I've pointed out in this review round :+1:

Stephen-ONeil (changes requested):
I think the design and architecture is good enough for an initial release. I think some small things could be better but I also think they can wait, this has been open for a long time and it's blocking a lot at this point.

Trying just to focus on specific code comments that I think serve as important lessons or point out (potential) bugs.

Aside from extra polish, maybe after a break and some other small scale branches, I'd like you to write unit tests for these new components in a follow up PR. Eventually good coverage will be expected in anything but the more urgent hotfix branches, but we're not there yet.

client/src/components/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+import { CSSTransition } from "react-transition-group";
+
+import { trivial_text_maker } from "src/models/text";
+
+import { IconX } from "src/icons/icons";
+
+interface SidebarProps {
+  is_open: boolean;
+  callback: (value: boolean) => void;
```
Stephen-ONeil, line 14: More descriptive name would be handy, like `open_close_callback` (of even splitting in to distinct open and close callbacks).

client/src/components/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,124 @@
...
+  children: React.ReactElement;
+  title_text: string;
+  sidebar_toggle_target: string;
+  return_focus_target: HTMLElement | undefined;
+  keydown_close: () => void;
+  keydown_close_value: boolean;
```
Stephen-ONeil, line 20: Best I can tell, this pair of props just exist so that the state for `keydown_close_value` can be owned by the `GlossarySidebarController`. In the process, it creates a lot of tricky to following passing down (and then via the call back, up) of props/information across three different components. Ouch! Really complicates this particular use case, and will definitely make future use/maintenance of `Sidebar` a challenge.

So, in case I missed something, the `Sidebar` itself could just manage this all internally, right? If I'm right then I think I've got some follow up advice, so let me know and we can talk further :+1:

Stephen-ONeil (changes requested):

client/src/components/Sidebar/Sidebar.tsx
sranc046: Attempted a different approach to keep this all internal, let me know if that works!

client/src/handlebars/register_helpers.side-effects.js
```diff
@@ -669,8 +669,15 @@ HandlebarsWithPrototypeAccess.registerHelper("gl", function glossary_link(key) {
...
+    const glos_item = glossaryEntryStore.lookup(key);
     return new HandlebarsWithPrototypeAccess.SafeString(
-      `<span class="nowrap glossary-tippy-link" tabindex="0" data-ibtt-glossary-key="${key}" data-toggle="tooltip">${display}</span>`
+      `<button class="nowrap glossary-sidebar-link" data-ibtt-glossary-key="${key}" aria-label="${
+        display +
+        `, ` +
+        trivial_text_maker("open_glossary_definition") +
+        ` ` +
+        glos_item.title
+      }" data-toggle="glossary_sidebar">${display}</button>`
```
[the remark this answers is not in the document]
sranc046: Updated it. If this approach gets approved I will apply it to the remaining files

client/src/glossary/glossary.js
```diff
@@ -185,7 +170,23 @@ export default class Glossary extends React.Component {
...
         </h1>
         <ScrollToTargetContainer target_id={active_key}>
-          {!is_a11y_mode && <BackToTop />}
+          {!is_a11y_mode && (
+            <FloatingButton
+              button_text={text_maker("back_to_top")}
+              aria_label={text_maker("back_to_top")}
+              showWithScroll={true}
+              handleClick={() => scroll_to_top()}
+              tabIndex={-1}
```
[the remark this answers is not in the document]
sranc046: This was just mimicking the original setup for the BackToTop component. I believe the idea is that the button isn't keyboard focusable because of its placement in the DOM and every glossary definition has a hidden link back to the top when navigating via keyboard.

client/src/components/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+    const target = (e.target as HTMLElement).closest(sidebar_toggle_target);
+    if (!target && is_open) {
+      this.closeSidebar(e);
+    }
+  };
+
+  closeButtonClick = (e: React.MouseEvent) => {
+    if (e.detail == 0) {
+      this.props.keydown_close();
```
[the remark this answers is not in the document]
sranc046: Would there be a better approach to this?

client/src/glossary/GlossarySidebar/GlossarySidebarController.tsx
sranc046: Can discuss this in our dev meeting, I have a few questions, thanks!

client/cypress/integration/InfoBase/cypress-snapshots.js
```diff
@@ -127,20 +131,18 @@ module.exports = {
       "Tested on index-eng.html#infographic/gov/gov/covid": {
         "Axe violations allow list": {
           "aria-hidden-focus, serious": [
-            // Should be fixed by PR #1315
-            "<span class=\"link-unstyled\" tabindex=\"0\" aria-hidden=\"true\" data-toggle=\"tooltip\" data-ibtt-html=\"true\" data-ibtt-container=\"body\" data-ibtt-text=\"<p>",
-            "<span class=\"link-unstyled\" tabindex=\"0\" aria-hidden=\"true\" data-toggle=\"tooltip\" data-ibtt-html=\"true\" data-ibtt-container=\"body\" data-ibtt-text=\"<p>",
-            "<span class=\"link-unstyled\" tabindex=\"0\" aria-hidden=\"true\" data-toggle=\"tooltip\" data-ibtt-html=\"true\" data-ibtt-container=\"body\" data-ibtt-text=\"<p>"
+            "<span class=\"link-unstyled\" tabindex=\"0\" aria-hidden=\"true\" data-toggle=\"glossary_sidebar\" data-ibtt-html=\"true\" data-ibtt-container=\"body\" data-ibtt-text=\"<p>",
```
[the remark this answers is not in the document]
sranc046: Oops, yeah my bad. Thought I could just run the test for a specific route and it would only update that case without affecting the rest of the file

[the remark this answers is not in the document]
Stephen-ONeil: All good. I left a line in the output that directs devs to update the file by running it locally. Could be clearer there, that's only for the worst case (something creates a false positive across a large number of routes) and requires care. Otherwise, updating lines manually works well.

client/src/components/Sidebar/Sidebar.tsx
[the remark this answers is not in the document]
Stephen-ONeil: `screenX` and `screenY` of 0,0, as mentioned above, should work in IE (work-ish, it's a hack, does't help if the user's mixing keyboard and mouse navigation for one). Like I said, not working in IE is minor and stops being a concern of ours relatively soon, we can ignore support for it here IMO.

Questions about the double equals still stand.

client/cypress/integration/InfoBase/cypress-snapshots.js
```diff
@@ -16,7 +16,11 @@ module.exports = {
     },
     "Glossary": {
       "Tested on index-eng.html#glossary": {
-        "Axe violations allow list": null
+        "Axe violations allow list": {
+          "color-contrast, serious": [
+            "<button class=\"btn btn-ib-primary floating-button floating-button--fixed\" tabindex=\"-1\" aria-label=\"Back to top\" style=\"top: auto;\">Back to top</button>"
```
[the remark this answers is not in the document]
sranc046: Forgot to mention this in our dev meeting, but can't seem to even locate the error with the AXE tools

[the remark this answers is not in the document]
Stephen-ONeil: Hm, it does seem that the AXE engine running in cypress isn't always in sync with the version that runs in the browser plugin. I think the browser plugins auto-update, while the cypress plugin will only update when our package locks do. Aside from that, I also don't know if the maintainers keep them in sync or use the same default configuration.

Aside from that, there's some potential differences in the page as examined in CI. The main one is the dev link banner being inserted (seems unrelated here). Might also be some other difference on the page, such as if you've scrolled/interacted with the page before running AXE locally. 

Outside of that, we've not been using this setup very long, so it could be some other kind of gotcha I've not accounted for. Worth looking in to it a bit more, since we need to get more acquainted with it, learn what the gotchas might be.

client/src/common_text/common_lang.yaml
```diff
@@ -148,6 +148,13 @@ glossary_link_title:
 back_to_top:
   en: Back to top
   fr: Retour au début
+glossary_button:
+  en: Glossary
+  fr: Glossaire
+open_glossary_definition:
+  transform: [handlebars]
+  en: "{{display}}, click to open glossary defintion for {{title}}"
+  fr: appuyez pour ouvrir la définition du glossaire de
```
Stephen-ONeil, line 157: Are the quotes on the English intended? On the French missing the templating.

client/src/components/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,115 @@
...
+    });
+  }
+
+  closeSidebar = (e: Event) => {
+    const sidebar_node = this.sidebar_ref.current;
+    if (
+      this.props.is_open &&
+      sidebar_node &&
+      !sidebar_node.contains(e.target as HTMLElement)
+    ) {
```
Stephen-ONeil, line 47: The logic for determining if the click is outside the sidebar should probably be inside `handleWindowClick`. After that, both `handleWindowClick` and `closeButtonClick` should call `closeSidebar` so that any closing logic is applied consistently.

client/src/glossary/GlossarySidebar/GlossarySidebarList.tsx
```diff
@@ -0,0 +1,106 @@
...
+    const { search_phrase } = this.state;
+    const { open_definition } = this.props;
+
+    const results = glossary_search_config.query_sync(search_phrase);
+
+    const items_by_letter = get_glossary_items_by_letter(results);
+
+    return (
+      <>
+        {results.length == 0 && search_phrase != "" ? (
```
Stephen-ONeil, line 70: Should be strict comparisons. As always, if a type-insensitive comparison _is_ the right choice, then a) it's probably better to just explicitly write out the different expected cases and b) if it really makes sense to use a type coercing comparison then you should comment on it.

client/src/glossary/glossary.js
```diff
@@ -185,11 +170,24 @@ export default class Glossary extends React.Component {
...
-          {!is_a11y_mode && <BackToTop />}
           <Glossary_
             active_key={active_key}
             items_by_letter={items_by_letter}
           />
+          <FloatingButton
+            button_text={text_maker("back_to_top")}
+            aria_label={text_maker("back_to_top")}
+            showWithScroll={true}
+            handleClick={() => scroll_to_top()}
```
Stephen-ONeil, line 181: Equivalent to `handleClick={scroll_to_top}`

client/src/glossary/glossary.yaml
```diff
@@ -12,4 +12,36 @@ glossary_translation:
 
 glossary_related_data:
   en: 'Related Data:'
-  fr: 'Données connexes :'
\ No newline at end of file
+  fr: 'Données connexes :'
+
+glossary_placeholder: 
+  en: "Search for a term"
```
Stephen-ONeil, line 18: Outer quotations not necessary, or standard for the project, in text maker yaml. I'm surprised to see that they get stripped out somewhere and aren't appearing in the app!

client/src/glossary/GlossarySidebar/GlossarySidebar.tsx
```diff
@@ -0,0 +1,110 @@
...
+      >
+        <div className="glossary-sb__header-wrapper">
+          <div className="glossary-sb__header">
+            <div className="glossary-sb__search-wrapper">
+              <div className={"glossary-sb__search-bar"}>
+                <div className={"glossary-sb__icon-container"}>
+                  <span aria-hidden="true">
+                    <IconSearch
+                      width="30px"
+                      color="#2C70C9"
```
Stephen-ONeil, line 74: Should import and use the `secondaryColor` variable, since that appears to be where this is from.

Stephen-ONeil (changes requested):
Just some minor clean up points, just about done!

Stephen-ONeil:
After that, #1315 is flagged for your review, to merge down in to this branch. We should all kick the tires on it after that, then it should be good to release!

Stephen-ONeil (changes requested):
- [ ] the list of glossary items should use list(s), rather than a bunch of styled divs. I think one list per letter, with the letter using some appropriate heading level. You can keep the same styles without too much work (to start, we have a sass mixin for `unstyled-list `)
  - [ ] can't select text inside the side bar (aside from the main header and anything I type in to the search bar)
  - [ ] when you select a glossary item from the sidebar list, focus should probably end up on `glossary-sb__item-def`, since the button in the list tells screen reader users it will open the definition (jumping back to the close button like it does will be unhelpful)
  - [ ] review #1422 for merging in to this branch
  
At least one of these may be a repeat from earlier, just remind me if there's a good reason to either not do it, or potentially to leave it for a follow-up issue!

Stephen-ONeil (changes requested):
Hm, I see what you meant about the `set_done_animating`/`done_animating` pattern being a rough one. Here's some approaches that spring to mind and my thoughts on them:

1. what you did, keep the `Sidebar` stateless and rendering arbitrary `children`, but give it some callback so that the parent can hold that state and hook in to the `Sidebar`'s `onEntered` and `onExited`.
    - only change I would make if going with this would be to tweak the naming and make the props on `Sidebar` more transparent. The current `set_done_animating: (boolean) => void` is very prescriptive against what the parent is using the callback for, suggesting that parent has some `done_animating: boolean` state. In reality, the parent could be doing anything it wanted in this callback, and that' not really `Sidebar`'s problem! I'd just give `Sidebar` pass through props for arbitrary `on_entered` and `on_exited` events. That way the prop's name isn't coupled to how consumers choose to use it (the power of the callback argument is that it's totally up to the consumer what it does)
    - pro: flexible, everything's up to the consumer
    - con: tedious, error prone. Consumers need to wrap this in something stateful that implements something like the `is_animating` state and callbacks, then pass that state through all the actual `Sidebar` children so that focus management can be aware of animating. Lots of stuff for every user of the `Sidebar` to forget to do, or do wrong
    
2. type `Sidebar`'s children prop as `children: (is_animating: boolean) => React.ReactNode;` The parent would call as something like 
```
<Sidebar>
  { (is_animating) => 
    <SomeOtherComponentOrWhatever is_animating={is_animating} />
  }
</Sidebar>
```
  - pro: at least some of that tedious repeated state management is now implemented once, and always, within the `Sidebar` (it owns `is_animating`)
  - con: maybe less flexible, the consumer can't have arbitrary on enter/exit behaviour (without getting even hackier and injecting that as side effects of the `children` callback, or even behvaiour of the rendered children components themselves? Ouch, hurts to think about). Still might have to do a lot of passing down of that prop through the rendered children
  - mixed: `children` needing to be a function might be annoying if there's some case where `is_animating` isn't needed, but I think it might always be given the need to manage focus when using the sidebar, so this works as a reminder to use it when rendering the children
  
3. [Context API](https://reactjs.org/docs/context.html)! `Sidebar` declares a context that can provide the `is_animating` state to any children that need to consume it
  - pro: no weird tricks, no potential need to pass the prop down through a bunch of levels. Only the `Sidebar` and, potentially, _some_ specific child needs to know
  - con: probably the least visible, context doesn't show up on the component's API, you need to open the file to see the context provider being exported. Easy to miss when looking at an example use case too, might be consumed somewhere totally different. Also, what ever component ultimately _does_ consume that context provider is coupled to always being a child of a `Sidebar` (although potentially needing a specific prop and even passing that prop through multiple layers is even worse, in terms of infecting a bunch of components with some `Sidebar` specific implementation detail). 
  
I'd probably suggest 3 myself, but maybe that's just because it's kind of rare to have a good reason to use contexts, in my experience, haha.

Stephen-ONeil:
I'm going to look further in to the broken text selection/the need for `tabIndex={-1}` everywhere and get back to you. Guessing it's something `FocusLock` is causing. Can't have that long-term, since it means anyone using the `Sidebar` needs to know/remember to be throwing `tabIndex={-1}` anywhere there's text that should be selectable, ouch.

Stephen-ONeil:
RE: above comment on `persistentFocus=false`, looks like I was off! I was testing that prop, to see if it was related to our problem, very quickly at the very end of the day, so rather than find the relevant `tabIndex={-1}`s I just removed all across src for expedience. Tried to recreate the "fix" with more care just now, and it turns out it _wasn't_ setting `persistentFocus=false` that fixed anything (it _is_ false by default, turns out), it was removing the `tabIndex={-1}` on the `#app-focus-root` element.

Now that's a bother, as far as I can tell from skimming [the spec](https://html.spec.whatwg.org/multipage/interaction.html#the-tabindex-attribute), an ancestor's `tabindex` shouldn't matter to any descendants. Sounds like this is effectively the same case for `react-focus-lock` as described [here](https://github.com/theKashey/react-focus-lock/issues/94#issuecomment-589330581), so might be a feature-not-a-bug kind of headache.

Mixed blessing, `#app-focus-root` appears to never receive focus via JS (and click focus on it doesn't seem to have a purpose) so it might not need that `tabIndex={-1}` at all. Preferably we'd dig up the history of _why_ it has it right now before we remove it though. Probably also drop that `app-focus-root` id from it too, since it will lose/has lost meaning, although make sure to update the related CSS selector when you do.

client/src/InfoBase/App.js
```diff
@@ -132,114 +130,107 @@ export class App extends React.Component {
   render() {
     const { outage_msg, showSurvey } = this.state;
     return (
-      <div
-        tabIndex={-1}
-        id="app-focus-root"
-        className={`app-focus-root--${is_a11y_mode ? "a11y" : "standard"}`}
-      >
-        {is_a11y_mode && <NoIndex />}
```
Stephen-ONeil, line 140: Lost this line during the rebase `{is_a11y_mode && <NoIndex />}` (and deleted the import in the following cleanup). Bring that back, thanks!

sranc046: Oh oops, my bad

Stephen-ONeil: Just move it in to the `ErrorBoundary` this time, it never really needed to be outside of it.

Stephen-ONeil (approved):
Final approval, looks good to merge to me! :tada: :tada:

PR #1272, opened 2021-10-18
General retrying fetch pattern

Stephen-ONeil:
Follow up to #1267. Even though the GQL queries were the majority of our fetches and the most likely to have issues (more complex and varied, especially with bathing, than other fetches, much less likely to be cached), the retrying fetch pattern has only caught a couple problems in them since last Friday. This PR (mainly) extends the retrying fetch pattern (with all its extra logging) to all fetch requests made in the client. If nothing else, the logging should make the actual sources of these issues far more visible.

Some small tangential cleanup in `src/InfoBase/bootstrapper.js` got in here. Some necessary, some not. Biggest change there was shifting some polyfills around and making them dependent on feature detection. Means this branch needs IE testing for sure.


ktw1016 (approved):
👍👍

PR #1288, opened 2021-11-01
Split cypress CI jobs in half

Stephen-ONeil:
Not using cypress' parallelization "support", since it
  1) requires per-test config files, which we don't do for the route load tests
  2) requires you to use (and potentially pay for) their dashboard service. Not something we need or care about right now

Just a small tweak to `route_tests.spec.js` and the CI config to allow batching our route load tests configs. The outputs aren't reconciled, so you could potentially need to look in two places to find all route load errors after your CI job fails. Probably not worth the complexity to try and do more than this though, and it's a good 2-3 minute CI optimization!

After this merges, open an issue to look in to these libraries
  - https://github.com/Shelex/cypress-parallel-specs-locally lightweight, just some scripts to make it easy to manage running multiple cypress processes in parallel locally (i.e. within the same container). Would need to be mindful of memory usage, haven't been logging that on the cypress jobs
  - https://github.com/sorry-cypress/sorry-cypress heavier, more focused on fully replacing the cypress dashboard service (including it's parallel test orchestration component). Since we don't really need or want the first party dashboards anyway, not too likely we want to set this up either right now

If we start writing more robust per-route cypress tests (probably should eventually, and gradually replace the route load tests with them), might need to change the approach up. This is a specific solution based on how we do the route load test configs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -217,7 +217,7 @@ jobs:
 
       - run: (cd server && npm run snapshot_tests)
 
-  test_end_to_end:
+  cypress_tests:
     docker:
       - image: eaadtbs/ib-ci-cypress:4.0
         <<: *dockerhub_auth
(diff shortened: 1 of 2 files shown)


sranc046:
Nice changes, definitely seems to be cutting down on CI time! Seems straightforward enough (although putting my lodash knowledge to work lol)

sranc046 (approved):

Stephen-ONeil:
> Nice changes, definitely seems to be cutting down on CI time! Seems straightforward enough (although putting my lodash knowledge to work lol)

Haha, it's not a very typical lodash construction. Ask me on slack if you want it explained more.

PR #1289, opened 2021-11-02
Fix typescript progress script

Stephen-ONeil:
It didn't know about the root of the `/client` dir being configured as the base root, so non-relative paths couldn't fully resolve (and, since we dropped the .js/.ts prefixes from the import statements, not resolving meant not knowing what type of file it was).

Bunch of cleanup while I was at it, reorganized a bunch of scripts.


Stephen-ONeil:
Self approving, just minor dev script fixes/reorg.

PR #1293, opened 2021-11-05
Async SearchConfigTypeahead

Stephen-ONeil:
Still a fair bit to do, will map out remaining steps next week. Basic idea is that anywhere we're mixing API data with local data, we should force the local data to return as promises too. Most of it is intended to become API data some day, and in the mean time we only have to deal with a single API/avoid mixed sync and async code.


Stephen-ONeil (commented):
Self reviewed/approved.

PR #1294, opened 2021-11-10
Client Jest configuration

Stephen-ONeil:
... taking over from #936, this one's actually getting finished though, I swear.

TODO:
  - [x] unit test configuration
    - [x] simpler files are testable
    - [x] need a transform or mock for static content
    - [x] @testing-library jest-dom and react
  - [x] a couple unit test examples for
    - [x] non-react
      - [x] `src/general_utils.ts`, as easy as it gets, nothing to mock even
        - not going to take the time to test all of the exports in this PR, it's a big mess of utils
      - [x] `src/request_utils.ts`, slightly more complicated example, some mocking involved
    - [x] react
      - [x] `AlertBanner`, simple and nothing to mock
      - [x] `HeaderBanner`, not much more complicated but includes mocking a text maker
  - [x] CI
      - [x] add CI job
      - [x] see what can be cached with respect to coverage report generation. Initial run was much slower than subsequent runs locally. If this works out, look in to repeating pattern in `server` and `form_backend`
  - [x] documentation
    - [x] _start_ collecting client unit testing standard and patterns
      - expect to grow/refine this doc as we write more tests for sure, this PR isn't going to run in to every testing case we have
    - [x] integration tests are out of scope for this initial PR, but should be getting my thoughts in order while I'm at it. Flesh out what needs to be done for this/what they should cover and document it
  - [x] before merging, brief everyone on build script name tweaks, and testing setup in general


Stephen-ONeil:
Oh, interesting but a bit confusing, those last two commits that are marked as failing aren't from this PR, they're commits on a different branch that _reference_ this PR. Have to read a little carefully to notice, IMO GitHub should make the distinction clear (and maybe not show the CI status for them). This branch _is_ passing.

sranc046 (approved):

ktw1016 (approved):
awesome👍

PR #1298, opened 2021-11-15
Cypress error tracking cleanup

sranc046:
Removed a bandaid fix for cypress's error tracking. There was a specific case earlier on that was causing the Error Boundary route to throw a console error which doesn't fail cypress tests by default. This is no longer the case and the extra tweaks to cypress were sometimes cluttering actual errors.


Stephen-ONeil (approved):
:+1:

PR #1299, opened 2021-11-15
Service inventory

ktw1016:


Stephen-ONeil:
TODO:
  - [ ] ... some of the services datasets aren't unique by ID, that's gonna cause problems
    - e.g. service searching returns multiple instances of a service (one per year)... but they all navigate to the same "single service route", which has an id arg but no year arg. Are prior year service versions actually used anywhere? Do we just need the most recent, per-ID?
  - [ ] URL construction for services in search produces bad URLs when searching from within a specific infographic section belonging to any non-service. The URL tries to route to the corresponding (always non-existent) section of the service infographic. Something goofy there (I think that the ID for the service has been shoved in to the URL param for section, instead of the URL for the subject ID, get's clobbered by the search behaviour; otherwise it would fall back to the default and work as expected?)
    - ... why is it constructed this way, are service IDs not unique between departments or something?
  - [ ] maybe a won't-fix, but the loading spinners used in service panels are inconsistent with other infographic panel's in terms of size

ktw1016:
> * [ ]  ... some of the services datasets aren't unique by ID, that's gonna cause problems
>   
>   * e.g. service searching returns multiple instances of a service (one per year)... but they all navigate to the same "single service route", which has an id arg but no year arg. Are prior year service versions actually used anywhere? Do we just need the most recent, per-ID?

So, service route should show the service for the most recent submission year. For the older submission years, it's not being used yet but we're planning to use it (for e.g., panel "Online status of services", showing historical change for the data)

> * [ ]  URL construction for services in search produces bad URLs when searching from within a specific infographic section belonging to any non-service. The URL tries to route to the corresponding (always non-existent) section of the service infographic. Something goofy there (I think that the ID for the service has been shoved in to the URL param for section, instead of the URL for the subject ID, get's clobbered by the search behaviour; otherwise it would fall back to the default and work as expected?)
>   
>   * ... why is it constructed this way, are service IDs not unique between departments or something?

Service IDs are unique and I agree URL should be changed so that it's service ID only, without department ID

Stephen-ONeil:
TODO:
  - [ ] don't love the (inconsistent) sentence in the intro panels about using the feedback form. Doesn't clearly direct the users to it, "feedback form" is a somewhat internal terminology for it, and it's full of questions unrelated to the service inventory anyway. Would prefer a service inventory specific template, directly linked from the intro panels.
  - [ ] inconsistent service counts. Most panels filter to the latest year, but the panel listing all services includes dead services as well (importantly, doesn't make that very clear, so it just looks like the counts are not consistent)
    - maybe present historical services in a separate table under the first, counted separately   
  - [ ] service overviews should include start/end year, or at least flag dead services

Things to follow up on:
  - [ ] are all years in the service data actually fiscal years? If not, want a double check on which are which
  - [ ] going to want to refactor all the models along the same lines as the covid models for multiple years of service data, eventually
  - [ ] why no CRSO level services infographics?

PR #1302, opened 2021-11-19
Server side service searching

Stephen-ONeil:
Taking over from #1244

TODO:
  - [x] lots of cleanup from first rough proof of concept commit
  - [x] extract server searching pattern, make `search_text` a model utility
    - maybe a field that derives from a list of other text fields, deburing, sorting, and uniq-ing their words
    - maybe with producing en, fr, AND bilingual fields
    - if we pre-lowercase the field and constructed regex, making it an index [should](https://docs.mongodb.com/manual/reference/operator/query/regex/#index-use) help optimize 
  - ~probably want to share `search_phrase_to_all_words_regex` between `client` and `server`~
    - hmm, woud mean writing it in JS and maintaining types on the side I guess, not perfect 
    - skipping for now, figuring out a good way to share common code is probably its own PR
  - [x] careful testing (ha, wish #1294 was merged already), make sure new searching retains old search's behaviour
  - [x] could be a small optimization to pre-process the search phrase in to sorted unique words before sending the request to the client, more chances to his cache that way
  - [ ] first search can be particularly slow, given the function wake up time. Thought I had the initial load waking the server up with an empty query? Is that query being cached or something, or am I forgetting how that was set up?

### DIFF
--- a/client/src/core/injected_build_constants.ts
+++ b/client/src/core/injected_build_constants.ts
@@ -24,4 +24,4 @@ export const cdn_url = typeof CDN_URL !== "undefined" && CDN_URL;
 
 export const local_ip = is_dev && typeof LOCAL_IP !== "undefined" && LOCAL_IP;
 
-export const services_feature_flag = is_dev && false;
+export const services_feature_flag = is_dev && true;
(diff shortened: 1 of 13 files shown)


ktw1016 (approved):
Code looks good to me.

My only concerns are client, server resolver and regex construction being tightly coupled but as of now, it's minor and for optimization reasons, also well commented. My other minor thought is when `search_{something}` will become too many where our graphql tree (root) will get messy but not a concern at the moment

Thanks for this, great work!!

PR #1308, opened 2021-11-23
Early GraphQL wake up

### LINKED ISSUE #1306: Wake up cloud functions during initial load
Noticed during the work on #1302, first search from the homepage can be particularly slow given the function wake up time. Thought I had the initial load waking the server up with an empty query? Is that query being cached or something, or am I forgetting how that was set up? Look in to that, worth poking the server to wake it up regardless, negligible cost and will at least sometimes speed things up in prod.

Stephen-ONeil:
Closes #1306


Stephen-ONeil:
Very minor, self-approved.

PR #1312, opened 2021-11-29
fix z-index of dropdown menu and its svg

ktw1016:
Fixed the problem where SVG floats on top of other dropdown menu

image (1)

but not sure what's going on with the weird style on IE. Been playing around with IE DOM inspector and the dropdown button style seem to be changing somewhat randomly


Stephen-ONeil (approved):
:+1:

PR #1315, opened 2021-11-30
New Tooltip component

Stephen-ONeil:
To be merged with #1270, provides a replacement for the tooltip system that that PR will be dropping. Don't want to merge sooner than that as these new tooltips, intentionally, are not styled exactly the same as our old tooltips were.

TODO:
  - [x] confirm screen reader usability
    - didn't work directly, minor but serviceable workaround by adding a screen reader only copy of the intended tooltip content in-line with the tooltip target. Welcome to suggestions on that though, I know it's not perfect
  - [x] write some unit tests, want to assert the expected behaviour against upstream changes to the library/possible replacements in the future


Stephen-ONeil:
Somewhat of a bigger problem, but I noticed that the table cells that contain these tooltips are just read as "blank" by screen readers. Also happens with cells containing svgs in other tables (e.g. department level results). Need to dig in to that more deeply.

ktw1016:
> Somewhat of a bigger problem, but I noticed that the table cells that contain these tooltips are just read as "blank" by screen readers. Also happens with cells containing svgs in other tables (e.g. department level results). Need to dig in to that more deeply.

It seems to work exactly same as non table cell tooltips for me, which is working fine

ktw1016 (approved):
👍👍

Stephen-ONeil:
Paused pending #1270. This is unnecessary (and unwanted) until the glossary side menu is ready to merge (if merged sooner, we'll have conflicting tooltip display/behaviour).

sranc046 (changes requested):
Two comments:
- Some tooltips contain links which are hard to read because of the colour
- Sometimes a tooltip stays open when moving on to another via keyboard, it'll only disappear when the page automatically scrolls or when using the mouse

Stephen-ONeil:
> * Sometimes a tooltip stays open when moving on to another via keyboard, it'll only disappear when the page automatically scrolls or when using the mouse

Hmm, not sure I'm able to recreate this. Can you try again and, if you can, note where the mouse cursor is located at the time? It might have been left over top of where the tooltip opened, which would then keep that tooltip open until the page scrolls and it  out from underneath or the cursor is moved directly.

sranc046:
> Hmm, not sure I'm able to recreate this. Can you try again and, if you can, note where the mouse cursor is located at the time? It might have been left over top of where the tooltip opened, which would then keep that tooltip open until the page scrolls and it out from underneath or the cursor is moved directly.

Yeah it was the cursor position, completely missed that while I was testing, my bad... 🤦‍♂️

PR #1318, opened 2021-12-05
Making DisplayTable Util Buttons more Obvious

csong202:
Issue: #1277


client/src/components/DisplayTable/DisplayTable.tsx
```diff
@@ -486,6 +497,8 @@ export class _DisplayTable extends React.Component<
       .reverse()
       .value();
 
+    const footer_content = {};
```
ktw1016, line 500: Was this supposed to be cleaned up? Doesn't seem used

client/src/components/DisplayTable/DisplayTable.tsx
```diff
@@ -509,6 +522,12 @@ export class _DisplayTable extends React.Component<
           util_components_with_defaults && "display-table-container--with-utils"
         )}
       >
+        <div className={"display-table-utils"}>
+          <div>{selectPageUtil}</div>
```
ktw1016, line 526: Minor but this `div` doesn't seem needed

client/src/components/DisplayTable/DisplayTable.tsx
```diff
@@ -533,19 +552,11 @@ export class _DisplayTable extends React.Component<
...
-                          selected={page_size}
-                          on_select={this.change_page_size}
-                          page_size_increment={page_size_increment}
-                          num_items={_.size(sorted_filtered_data)}
-                          num_options_max={page_size_num_options_max}
-                        />
-                      )}
-                    </div>
                     <div className={"display-table-container__utils"}>
                       <button
```
ktw1016, line 561: This button should actually be the first tabbable element, so it should go before `selectPageUtil`, so all of this div can be cleaned up and the button can move up

ktw1016 (changes requested):

client/src/components/DisplayTable/DisplayTable.tsx
csong202: it was used in the "footer" of the table under the table closing tag. I just removed it since there's nothing we want in the "footer" right now

client/src/components/DisplayTable/DisplayTable.tsx
csong202: I ended up keeping it so I could put the skip-to-data button before `selectPageUtil`

ktw1016 (approved):
Looks good to me !

PR #1319, opened 2021-12-06
Fix infographic table of contents focus management

### LINKED ISSUE #1317: Move on-page focus when selecting link in Table of Contents
Hello!

Great site, looks generally super cool.

Just noticed a minor issue with your "table of contents". When I am using <kbd>Tab</kbd> to move around the page, I can open the Table of contents, select a subheading and then press <kbd>Enter</kbd> to actually move the window down, but if I keep using <kbd>Tab</kbd>, then it whips me back up to the next item in the table of contents.

Clicking/activating a table of contents link will visually move focus, but the real browser focus stays inside of the Table of Contents.

I have a little gif of this, although you can't really see when I press  <kbd>Tab</kbd>, so you have to imagine it happening.

---

Screen Recording 2021-12-03 at 13 33 18

Stephen-ONeil:
Fixes #1317

Funny quirk, the table of contents shared its scroll to mechanism with the functionality for directly linking to a panel. Directly linking to a panel DID (appear) to manage focus as desired, which is probably why no one checked that the table of contents did as well. That was a fluke, the `.focus()` call didn't actually do anything (targeted an unfocusable element), linking to the panel only _seemed_ to manage focus as expected because, when nothing on the screen has focus yet, the first keyboard navigation lands on the first focusable thing _on screen_, so scrolling the view was enough (on Chrome, not testing widely, not sure if this is in the spec or not). Ha.

### DIFF
--- a/client/src/infographic/Infographic.js
+++ b/client/src/infographic/Infographic.js
@@ -235,6 +235,8 @@ class Infographic extends React.Component {
               />
             )}
           <TableOfContents
+            subject={subject}
+            active_bubble_id={active_bubble_id}
             panel_titles_by_key={_.chain(filtered_panel_keys)
               .map((panel_key) =>
                 PanelRegistry.lookup(panel_key, subject.subject_type)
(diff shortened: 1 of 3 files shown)


ktw1016 (approved):
Great👍

PR #1322, opened 2021-12-06
Scroll to and focus utility

Stephen-ONeil:
Generic solution to the general case of #1317, we had a number of scroll/focus management snippets and they were inconsistently friendly to tab navigation. This fixes that, regardless of the specific element/tabindex value of the scroll + focus target.


ktw1016 (approved):
👍

PR #1324, opened 2021-12-06
Drop storybook, once again

### LINKED ISSUE #64: storybook-ing panel-subject combination cases (e.g. programs w/ out spending, etc. )
* subject-mocking API ? 
 * tests will help move to graphql
 * doesn't need to use storybook, could simply beef up the panel inventory
 * good for business knowledge and exposes bulk of the InfoBase code

### LINKED ISSUE #1241: Thought: component stories should be good starting points for untit tests
It's the main takeaway from [this](https://marmelab.com/blog/2021/08/31/testing-react.html) piece. Authour has a good point. We should be writing stories to cover all of the (valid, at least) use cases for a given component. Positive unit tests should import and run against these stories. Good interplay there, as we then have an easy way to visually validate the behaviour that's under test.

Most likely, this probably obviates the need/utility of integrating cypress in to the storybooks (... pending us ever untangling the environment issues that are currently stone walling a solid jest setup in the client; shouldn't be too hard just need the time for it).

Stephen-ONeil:
Up for debate over on slack, although maybe the conversation should move here.

Summary of my point of view: I don't think we're making good use of it, and I don't see us making enough use of it going forward to make the necessary continued investment/ongoing maintenance worthwhile. I'd happily throw this away and give all that (theoretical) time to writing good types and good tests.

Closes #1241
Closes #64 (ha, this was from the attempt at using storybooks from before this go around; I'm not getting talked in to it a third time so talk me out of dropping it now if you care)


Stephen-ONeil:
Discussion occurred on slack/during the dev meeting. Decided that it wasn't getting enough use to be worth the maintenance cost (breaking on unrelated webpack changes and package upgrades on top of needing to write/maintain the stories themselves). Not useful for a small team that isn't publishing any of its components. 

If any use cases for a given component aren't represented in the actual application, then it's more useful to consider simplifying the component to drop those than it is to try and maintain a sandbox to demonstrate them in, ha.

PR #1328, opened 2021-12-08
2021 people data

Stephen-ONeil:


Stephen-ONeil:
Ha, oops, invalid branch name (for dev link cloud function naming). Maybe that should be checked in a pre-push git hook...

PR #1331, opened 2021-12-08
Improved route load tests

Stephen-ONeil:
Just the resulting fixes from #1325

Recursively await resolution of _all_ spinners.

Extra sanity check assertion that the resulting page is not the error boundary.

### DIFF
--- a/client/cypress/integration/InfoBase/route_tests.spec.js
+++ b/client/cypress/integration/InfoBase/route_tests.spec.js
@@ -278,17 +278,31 @@ const run_tests_from_config = ({
           `http://localhost:8080/build/InfoBase/index-${app}.html#${route}`
         );
 
-        cy.get(".leaf-spinner__inner-circle", { timeout: 10000 }).should(
-          "not.exist"
-        );
-        cy.wait(1000); // temporary hack to make sure long-tail spinners are all caught and waited on
-        cy.get(".leaf-spinner__inner-circle", { timeout: 10000 }).should(
-          "not.exist"
-        );
-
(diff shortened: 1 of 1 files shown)


client/cypress/integration/InfoBase/route_tests.spec.js
```diff
@@ -278,9 +278,31 @@ const run_tests_from_config = ({
...
+            // would have been sufficient, but there's cases where a new spinner takes a split second to render
+            .wait(250) // Note: if any routes with lots of loading layers start flaking, may need to increase this (or optimzie those routes!)
+            .then(() =>
+              cy.document().then((document) => {
+                if (document.querySelector(spinner_selector)) {
+                  return wait_on_all_spinners();
+                } else {
+                  // Seems some errors that can be caught by the error boundary may skate by cypress. For an extra
+                  // sanity check, make sure we've not wound up on the error boundary at the end of all loading states
+                  // TODO: try and better identify what case causes this, see if we can make cypress see and display them directly
```
Stephen-ONeil, line 298: @sranc046 any guess on this? Might have something to do with why we were, at one point, using a custom console output watcher?

sranc046: Hmm maybe... Do we have a specific example of this type of error?

Stephen-ONeil: Check out ac8ea1d1, run an A11Y build, and visit the gov/any dept level service infographic.

Stephen-ONeil:
Want the regain test confidence before the upcoming burst of releases, self approving.

PR #1332, opened 2021-12-09
Convert VisibilityControl to typescript

sranc046:
Leaving out story conversion for now since not worth the trouble if we are removing them...

### DIFF
--- a/client/src/components/VisibilityControl/VisibilityControl.tsx
+++ b/client/src/components/VisibilityControl/VisibilityControl.tsx
@@ -1,6 +1,6 @@
 import classNames from "classnames";
 import _ from "lodash";
-import React from "react";
+import React, { Fragment } from "react";
 
 import { lang } from "src/core/injected_build_constants";
 
@@ -11,7 +11,7 @@ import "./VisibilityControl.scss";
 
 interface VisibilityControlProps {
   items: itemInterface[];
-  item_component_order: Record<string, unknown>;
+  item_component_order: ("count" | "icon" | "text")[];
   click_callback: (key: string) => void;
   show_eyes_override: boolean;
 }
@@ -21,10 +21,12 @@ interface itemInterface {
   aria_text: string;
   count: number;
   key: string;
-  icon: HTMLElement;
-  text: HTMLElement;
+  icon: React.ReactNode;
+  text: string;
   is_filtered: boolean;
 }
+
+type itemType = "count" | "icon" | "text";
 
 export class VisibilityControl extends React.Component<VisibilityControlProps> {
   render() {
@@ -40,7 +42,7 @@ export class VisibilityControl extends React.Component<VisibilityControlProps> {
       <div className="visibility-control">
         {_.map(items, (item) => {
           const item_components: {
-            [key: string]: boolean | JSX.Element | Element;
+            [key in itemType]: React.ReactNode;
           } = {
             count: !_.isUndefined(item.count) && (
               <div className="visibility-control__count_area" key="count">
@@ -101,10 +103,9 @@ export class VisibilityControl extends React.Component<VisibilityControlProps> {
                 )}
               </div>
               <div className="visibility-control__components">
-                {_.map(
-                  item_component_order,
-                  (component: string) => item_components[component]
-                )}
+                {_.map(item_component_order, (component, index: number) => (
+                  <Fragment key={index}>{item_components[component]}</Fragment>
+                ))}
               </div>
             </button>
           );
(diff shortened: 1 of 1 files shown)


client/src/components/VisibilityControl/VisibilityControl.tsx
```diff
@@ -9,7 +9,24 @@ import { secondaryColor, tertiaryColor } from "src/style_constants/index";
...
+  show_eyes_override: boolean;
+}
+
+interface itemInterface {
+  active: boolean;
+  aria_text: string;
+  count: number;
+  key: string;
+  icon: HTMLElement;
+  text: HTMLElement;
```
Stephen-ONeil, line 25: `icon` and `text` should be `React.ReactNode` (basically anything that is valid for the curly brackets inside of a JSX block to evaluate to; more JSX, a string, null, etc). 

Looks like the component also allows for them to be `undefined`, which the current typing doesn't accept (coincidentally, a `React.ReactNode` would) 

Maybe `text` should actually be a string, depends if the styles from `.visibility-control__text` make for a usage that should be constrained to text content.

Arguably count _could_ also be of type `React.ReactNode`, based on usage, although I think number is better. Just double check that that (and the rest of the typing) aligns with the current actual usage of this component\*.

\* important to check existing usage while typing, _but_ there's a balance to be struck between matching current usage and taking this as a chance to enforce tighter typing. Play it by ear.

client/src/components/VisibilityControl/VisibilityControl.tsx
```diff
@@ -22,7 +39,9 @@ export class VisibilityControl extends React.Component {
     return (
       <div className="visibility-control">
         {_.map(items, (item) => {
-          const item_components = {
+          const item_components: {
+            [key: string]: boolean | JSX.Element | Element;
```
Stephen-ONeil, line 43: `[key: "count" | "icon" | "text"]: React.ReactNode`

If you've addressed my first comment then `"count" | "icon" | "text"` is now duplicated, so probably extract that to it's own type to DRY things out.

client/src/components/VisibilityControl/VisibilityControl.tsx
```diff
@@ -84,7 +103,7 @@ export class VisibilityControl extends React.Component {
               <div className="visibility-control__components">
                 {_.map(
                   item_component_order,
-                  (component) => item_components[component]
+                  (component: string) => item_components[component]
```
Stephen-ONeil, line 106: I think if the above typing is sorted out, this should be able to figure out that it's `"count" | "icon" | "text"` automatically, won't need to be told it's a string.

Stephen-ONeil (changes requested):

client/src/components/VisibilityControl/VisibilityControl.tsx
```diff
@@ -9,7 +9,26 @@ import { secondaryColor, tertiaryColor } from "src/style_constants/index";
 
 import "./VisibilityControl.scss";
 
-export class VisibilityControl extends React.Component {
+interface VisibilityControlProps {
+  items: itemInterface[];
+  item_component_order: ("count" | "icon" | "text")[];
```
Stephen-ONeil, line 14: Use `itemType` here

client/src/components/VisibilityControl/VisibilityControl.tsx
```diff
@@ -22,7 +41,9 @@ export class VisibilityControl extends React.Component {
     return (
       <div className="visibility-control">
         {_.map(items, (item) => {
-          const item_components = {
+          const item_components: {
+            [key in itemType]: React.ReactNode;
+          } = {
             count: !_.isUndefined(item.count) && (
               <div className="visibility-control__count_area" key="count">
```
Stephen-ONeil, line 48: Can remove `key=count"` etc, given at line 107

Stephen-ONeil (changes requested):
Minor comments

Stephen-ONeil (changes requested):
Sorry, didn't highlight the other cases because they're outside of the current diff range, but
> Can remove key=count" **etc**, given at line 107
that "etc" was referring to all items in `item_components`, I believe at least one other has an explicit (and no longer necessary) key.

sranc046:
> Sorry, didn't highlight the other cases because they're outside of the current diff range, but
> 
> > Can remove key=count" **etc**, given at line 107
> 
> that "etc" was referring to all items in `item_components`, I believe at least one other has an explicit (and no longer necessary) key.

Oh my bad, completely missed that

Stephen-ONeil (approved):
:+1:

Issue #1326, opened 2021-12-08
Service Inventory release follow-up items


Stephen-ONeil:
From [this comment](https://github.com/TBS-EACPD/infobase/pull/1299#issuecomment-984716377)

Things to follow up on:
  - More urgent:
     - [ ] [this](https://app.circleci.com/pipelines/github/TBS-EACPD/infobase/5677/workflows/136631b0-6572-48af-80f1-debe6f105a70/jobs/42266) job should NOT have passed, a11y versions of the dept level service route (at a minimum) were throwing errors!
      - testing on #1325
    - [ ] in calculating percentage of service available online end-to-end, should NAs be excluded from the denominator?
    - [ ] these labels are a mess in French
![image](https://user-images.githubusercontent.com/11301919/144552522-46ec6d4f-86f2-41c3-8255-7426fe0bcba8.png)
  - Less urgent:
    - [ ] are all years in the service data actually fiscal years? If not, want a double check on which are which
      - [ ] either way, should be numbers, not strings, for consistency with other API years
    - [ ] going to want to refactor all the models along the same lines as the covid models for multiple years of service data, eventually
    - [ ] why no CRSO level services infographics? 
    - [ ] nulls vs 0s in data, are those meaningful and where? Those are a big source of bugs client side, if necessary would prefer to default them to 0 and, if necessary 
      - orrrr revisit the question of making the `DisplayTable` "understand" nulls be default...
      - didn't refresh my memory on this too closely, it may be that the default sorting explicitly rejects mixed content columns. A column of ALL empties may be fine? Double check this stuff when coming back to it
    - [ ] what's with the name "Customizable Service Graph"? Weak
    - [ ] are `first_active_year`/`last_active_year` useless?
      - ... thought I could use first_active_year/last_active_year to identify dead services but looks like dead services may have no 2019 submission and, in 2018, have no last_active_year. So last_active_year isn't going to ever be set?
      - all these first_active_years of 2018... can't be right, this must be more like first year reported on, which isn't really worth displaying to users
    - [ ] strange horizontal ticks [here](https://www.tbs-sct.gc.ca/ems-sgd/edb-bdd/index-eng.html#infographic/program/INDSC-BXL01/services/.-.-(panel_key.-.-'application_channels_by_services)), shouldn't all be 5,000,000. Also generally have the potential to get cut off in this panel
[...]

PR #1333, opened 2021-12-09
Fix french label for digital status panel and more margin on application channel graph

ktw1016:
Fixing part of #1326


Stephen-ONeil (approved):
:+1: 

For the sake of future searching, can you give the PR a more specific name?

PR #1348, opened 2021-12-20
Route load test anti-redirect check

### LINKED ISSUE #1307: Possible route load test enhancement, verify that window location is as expected
Somewhat for the sake of the test, but also to make maintaining the route test config easier to remember, one of the steps in asserting that a route loaded without error should be confirming that the route wasn't redirected/failed back to home.

Motivated by 417cbf7a4740fb69a290536a941277f527112643, we still had an item in the test config for the lab route, months after that route was deprecated. Was essentially just re-testing the home page, because the route fell back to `#start` after failing to match.

Might not want the test to be "is the final url the same as the configuration route", some routes might be intended to redirect/update parameters while loading. By default, asserting that the route is _not_ the home route once loading is done is probably enough. In case we want to use a test to assert that some route DOES redirect home, might as well add an optional confg parameter to disable that assertion.

Stephen-ONeil:
Closes #1307

TODO:
  - [x] was motivated to jump on this now because of cc6424db25fd685e2808c823287f74f5e988b6bb... but that instance doesn't actually redirect currently. Update infographic route to actually change the url (to have no  `active_bubble_id`) when trying to visit an invalid bubble
  - [x] identify routes that redirect but are still valid
    - [x] potentially make select config urls more specific
    - ~probably add a config option to skip the redirect test~
      - not necessary for current route tests, and the option for `expected_route_pattern` could be used if it was
    - [x] maybe add an optional config option for a custom final URL validation function


ktw1016 (approved):
Can't follow what's going on with subject stuff but href construction looks good to me !

Stephen-ONeil:
Took a bit of a trick, but the point was so that, e.g., `Program.subject_type` had type `"program"` instead of just `string`. There's a lot of benefits to the actual literal string value types on those.

PR #1338, opened 2021-12-13
Next SI update

ktw1016:
- [ ] On single service page, somehow group and show: `last_improve_feedback`, `last_gender_analysis` together
- [ ] On single service page, show  `last_accessibility_review` under website visits
- [ ] On single service standards panel: a stacked bar chart by fiscal year above the table like:
  2018-19: |---5 met---| |---10 not met---|
  2019-20: |---6 met---| |---8 not met---|
- [ ] Services standards: dropdown year selector

PR #1349, opened 2021-12-20
Hotfix for depts that didnt submit services in the most recent year

ktw1016:
Was gonna post this in next SI update but I guess we need to hotfix this. (Commit in the other PR which I'll remove https://github.com/TBS-EACPD/infobase/pull/1338/commits/e4aa70c878f9665b7ed9f30e8db176387ce822f5)

### DIFF
--- a/server/src/models/services/populate.js
+++ b/server/src/models/services/populate.js
@@ -556,53 +556,82 @@ export default async function ({ models }) {
 
   const org_summary = _.chain(filtered_service_rows)
     .groupBy("org_id")
-    .flatMap((services, org_id) => ({
-      id: org_id,
-      service_general_stats: {
-        report_years: get_years_from_service_report(services),
-        number_of_services: services.length,
-        number_of_online_enabled_services:
-          get_number_of_online_enabled_services(services),
-        pct_of_online_client_interaction_pts:
-          get_pct_of_online_client_interaction_pts(services),
-        pct_of_standards_met_high_vol_services:
-          get_pct_of_standards_met_high_vol_services(services),
-        num_of_programs_offering_services: _.chain(services)
-          .reduce(group_by_program_id, {})
-          .size()
-          .value(),
-      },
-      service_channels_summary: get_service_channels_summary(services),
-      service_digital_status_summary: _.flatMap(digital_status_keys, (key) =>
-        populate_digital_summary_key(services, org_id, "dept", key)
-      ),
-      service_standards_summary: get_final_standards_summary(services, org_id),
-      subject_offering_services_summary: get_subject_offering_services_summary(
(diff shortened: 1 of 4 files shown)


Stephen-ONeil:
Large impact to data, changes in pretty much every panel. Master on the left, this branch (local) on the right.
image

Are these new stats some how more correct, or is this an accident?

server/src/models/services/populate.js
```diff
@@ -148,14 +148,14 @@ export default async function ({ models }) {
...
-  const most_recent_submission_year = _.chain(raw_service_rows)
+  const absolute_most_recent_submission_year = _.chain(raw_service_rows)
     .map("dept_submissions__document__year")
     .uniq()
     .sort()
     .last()
     .value();
 
-  const service_rows = _.chain(raw_service_rows)
+  const filtered_service_rows = _.chain(raw_service_rows)
```
Stephen-ONeil, line 158: These aren't filtered

ktw1016: Isn't this filtering for the last submission year?
```
    .map((service_across_years) =>
      _.chain(service_across_years)
        .sortBy(({ submission_year }) => _.toInteger(submission_year))
        .last()
        .value()
    )
```

ktw1016: line 275

Stephen-ONeil: Ohh, yeah, my bad. Yeah, that rejects all duplicates from old years. I still don't think of this as "filtered", since its what ends up as the canonical rows in the DB. The filtering is just to reject data we aren't currently set up to do anything with.

ktw1016: Still, "rows" in `service_rows`, is different from the "rows" represented in raw data (`services.csv`) because it "filters" out service rows that are not the most recent submission year. IMO just `service_rows` doesn't clearly show that.
Is this rejecting previous years temporary so that you just want this to be called `service_rows`?

Stephen-ONeil: The DB will only ever have one document ("row") per service. The end state will be by-year sub-documents off that to hold the year data. 

It's a matter of perspective here, and I guess whether you were referring to csv rows or DB "rows" in the name. It's filtered from the perspective of the csv rows, but the content of the variable represents eventual database "rows" (from that perspective it's not filtered, it is the full set of rows the database will know about).

Talking about rows in mongo is inaccurate (and just confusing here anyway, haha). I think I usually refer to the data as "records" once it's been transformed from the raw csv rows, but I'm sure I've not always been consistent on that.

Stephen-ONeil (approved):
Everything checks out on the dev link (broken program routes now work, other routes are unchanged).

PR #1350, opened 2021-12-22
Typescript 4.5 

Stephen-ONeil:
TODO:
  - [x] upgrade TS (and Prettier)
    - [x] and eslint
      - [ ] ... and fix `no-unused-vars` rule, tons of linting for that
      - [x] while I'm at it, better project wide linting script, lint everywhere
      - [x] maybe a side-job in CI to check global linting, right now only client src code is actually regularly checked by CI
  - [ ] Identify mixed JS/type imports, add `type` flags (see [here](https://devblogs.microsoft.com/typescript/announcing-typescript-4-5/#type-on-import-names))
    - this is the only TS 4.5 feature we _should_ immediately port for, don't have a good plan to identify these yet though... 
      - [ ] [consistent-type-imports](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/consistent-type-imports.md)
        - also https://github.com/typescript-eslint/typescript-eslint/issues/4338
    - [ ] also make sure everyone's in the loop on using these going forward

PR #1351, opened 2021-12-22
Typescript 4.5 upgrade, loads of new linting

### LINKED ISSUE #1068: Have a CI step that checks that webpack output is clean
We should have a check that the webpack output has no errors or warnings. This is going to ensure that static typing and linting both pass. It's also going to speed up PR reviews.

This should not hold back dev-link creation or other CI steps

### LINKED ISSUE #1213: Add combined eslint fix + prettier write script
Want a shorthand to do them together, I've seen cases where the result of an eslint fix run produces output that prettier in turn doesn't like (e.g. eslint import reordering can leave newlines that prettier wants to remove). 

It should be easy to pass a single target file path to both via this script, so it probably can't be a one-liner directly in the root package.json.

Stephen-ONeil:
#1350 re-pushed with CI-friendly branch name :roll_eyes: 

Closes #1213
Closes #1068

TODO:
  - [x] upgrade TS (and Prettier)
    - [x] and eslint
      - [x] ... and fix `no-unused-vars` rule, tons of linting for that
      - [x] while I'm at it, better project wide linting script, lint everywhere
      - [x] maybe a side-job in CI to check global linting, right now only client src code is actually regularly checked by CI
        - bonus, extended this job to check prettier and types as well, removed prettier from the local pre-push hook
  - [x] Identify mixed JS/type imports, add `type` flags (see [here](https://devblogs.microsoft.com/typescript/announcing-typescript-4-5/#type-on-import-names))
    - this is the only TS 4.5 feature we _should_ immediately port for, don't have a good plan to identify these yet though... 
      - [x] [consistent-type-imports](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/consistent-type-imports.md)
        - also watch this and related issues https://github.com/typescript-eslint/typescript-eslint/issues/4338


sranc046 (approved):
Seems good, mostly just a CI job change and some linting (didn't look at all those files too too closely)

PR #1352, opened 2021-12-24
SVG icon aria-label and remove title

ktw1016:
I can change the prop name to `aria_label` if this looks ok


Stephen-ONeil (approved):
Approved as a hotfix for the sake of screen reader on Mac. Not fixed for the alternate screen reader I've been testing, orca, which seems to have a larger issue with content inside table cells.

PR #1296, opened 2021-11-13
TabContent and TabLoadingWrapper to ts

csong202:
- also converted the TabLoadingWrapper story to ts and made a story for TabContent

PR #1353, opened 2022-01-05
TabbedContent cleanup

Stephen-ONeil:
TODO:
  - [x] allow clicking anywhere in tab, not just on the label text
  - [x] drop export/use of the `TabbedControl` component? Having this exist separate from the content panel (which belongs to `TabbedContent`) will make it a real burden to get the aria attributes right/keep them maintained
      - maybe other uses of `TabbedControl` could just use `TabbedContent`? If not, should they use the `RadioButtons`? Hmm, radio button component doesn't do well with arbitrary number of options
      - I think replacing both with one more generic component would do a better job. Drop `tab_pane_contents`, make parent's responsible for rendering the correct child via callback
        - probably going to drop `TabLoadingWrapper` in the process, parents will be responsible for spinners as necessary. If that happens, will deprecate #1296
  - [x] use proper aria attributes `role="tablist"`, `role="tab"`, `role="tabpanel` etc
  - [x] give label an on focus ring, to help with tab navigation
  - [x] implement spec-compliant keyboard navigation
  - [x] ... test with screen readers. The tab aria spec's young/not finalized, want to make sure they at least make sense
    - voice over on Mac works well
    - orca's pretty bad with all tabs (not just ours)
      - the more I use it, the more orca seems to work poorly with lots of things, oof
  - [x] ~consider still allowing tab navigation between tabs. Goes against the spec, but I think it's not useful to expect users to know the spec/expect these to be spec compliant tabs (GitHub's tab-looking navigation, for example, is just a fancy link list and doesn't follow the tab spec; maybe ok for screen reader users, who will be told whether it _is_ a tab vs nav, etc, but keyboard navigators are probably frustrated by that, hmm)~
    - managed to change my mind again, sticking 100% to the spec
  - [x] write some unit tests while I'm here
    - [x] ... maybe revisit these tests. They're good enough, but it would be theoretically better if I tested `TabbedContent` and `TabbedContentStateful` in isolation, in case `TabbedContent` gains features in the future outside of `TabbedContentStateful`'s use cases
  - [x] tangent: use of tab role on `DisplayTable` pagination controls is poor, either:
    - ~remove that role~
    - ~fix (add tablist to parent, tabpanel to table content, tedious custom keyboard navigation...)~
    - [x] or open an issue to deal with this later #1356
  - [x] kind of want to return to working out a typing that enforces that open key prop is actually a valid key in the tab options, but not sure that's possible with current typescript features (outside of requiring consumers to declare things `as const` for it to enforce, possibly?)


Stephen-ONeil (commented):
Sat cold long enough to have fresh eyes my self (and to check off all my stretch goals, ha), so I'm going to self review.

PR #1355, opened 2022-01-08
FAQs for the Report a Problem form

### LINKED ISSUE #1334: Maintain a set of FAQs within the Report A Problem form
We get a lot of repeat  reports of non-issues. The general solution would be to start collection those along with a short explanation and putting it inside the Report A Problem form. If putting them in there passively doesn't resolve it, we could actively require users reporting a problem to acknowledge the FAQ before allowing them to submit.

In particular, we receive frequent reports that the previous year's program spending information is missing... seems the existing footnotes/warnings about it not being available till Public Accounts near year-end aren't being seen. Making those existing footnotes even more up-front is an alternative solution to this specific case, but I assume this is a bit of a survivorship bias problem; other users probably _do_ read that footnote (and thus don't report a problem), users who don't and end up reporting it are probably less likely to read a footnote no matter how prominent (but _are_ likely to interact with the report a problem form).

csong202:
Related to #1334 

Stephen edit:
  - [x] add functionality for forms to include additional arbitrary HTML
    - [x] use to attack FAQ items to an acknowledgement checkbox, with the content (and display) owned by the backend
  - [x] tweak how the form backend marks fields which are hidden from the user, explicit `"is_user_hidden": true` property rather than ``key === "meta" | !form_type``
  - Note: leaving report a problem FAQ + acknowledgement hidden for now, haven't yet written content for it, that'll be followup


client/src/components/PinnedFAQ/PinnedFAQ.js
```diff
@@ -26,47 +26,59 @@ export class PinnedFAQ extends React.Component {
       TM,
       is_initially_expanded,
       background_color,
+      no_pin,
```
csong202, line 29: seems unintuitive to have a `no_pin` option for a component called "PinnedFAQ" but it doesn't make sense to have a pin option inside the form since I can't pin things inside the modal

client/src/components/PinnedFAQ/PinnedFAQ.js
```diff
@@ -26,47 +26,59 @@ export class PinnedFAQ extends React.Component {
...
-                        <TM key={"a"} k={a_key} args={{ subject }} />,
-                      ]
-                    )}
-                  />
-                </div>
-              </AutoAccordion>
-            </ButtonToolbar>
-          </div>
-        </PinnedContent>
+        {(_matches) =>
```
csong202, line 64: not sure why, but when I don't include the `(matches) => ` the PinnedFAQ panel doesn't show up

client/src/FormFrontend/FormFrontend.yaml
```diff
@@ -6,6 +6,13 @@ form_frontend_privacy_ack:
   en: I accept
   fr: J'accepte
 
+form_frontend_faq_note:
```
csong202, line 9: can someone check over my french?

Stephen-ONeil (changes requested):
`FormFrontend` is generic, used for arbitrary templates from the `form_backend` server. Without the flag, these FAQs would be on the feedback, survey popup, etc too, right?

I think the Q/A content for a given form should belong to the same source as the content of the form, i.e. as part of the templates in `form_backend/templates`. I didn't have the idea when we were discussing things earlier, but I think form FAQs could be just another (special) type of field available for use in the templates/display by the `FormFrontend` component.

Assume, maybe, it looks something like this within the form templates:
```
"faq": { // this label has no special meaning, form_type is what indicates it's a FAQ for the frontend
    "required": true | false, // determine whether to show a check box, if false just default to true for field's value
    "value_type": "boolean",
    "form_type": "form_faq", // what indicates the client should display it as a faq
    "form_label": { 
      "en": "whatever label, I think it's optional though?",
      "fr": ...
    }
  },
```


Bonus, this fixes that first problem of knowing what FAQs to display in what instance of `FormBackend` :+1:

csong202:
>This is an old change, can be dropped, right?

yep, forgot to delete that

client/src/FormFrontend/FormFrontend.js
```diff
@@ -297,6 +296,42 @@ class FormFrontend extends React.Component {
...
+              <div style={{ fontWeight: 700 }}>
+                {text_maker("form_frontend_faq_note")}
+                {required_asterisk}
+              </div>
+              <FancyUL>
+                {_.map(field_info.faq_content, ([q, a], idx) => {
+                  return (
+                    <div key={idx}>
+                      <div style={{ fontWeight: "bold" }}>{text_maker(q)}</div>
+                      <TM k={a} />
```
Stephen-ONeil, line 311: The Q&A text itself should belong to the template on the server side. It's a leak for the backend to even know that the `text_maker` ecosystem exists on the frontend; it would add one more place where maintenance/deployment of both are, indirectly, coupled.

The two _are_ coupled in a number of ways. Ideally that'd be strictly limited to what explicitly is communicated through the backend's API. It largely is, _but_ the client does need to correctly implement behaviour for all possible form types or else it risks failing the backend's validation. I think that's unavoidable though, since the responsibilities for display and behaviour clearly belong to the client. The content of the form templates shouldn't live in both places though.

Stephen-ONeil: Possibly subtle, but it's fine for text keys like `"form_frontend_faq_note"` to belong to the client because that's something that is an implementation detail of the client's generic `"form_faq"` type field, rather than the content of any specific template owned by the backend.

client/src/FormFrontend/FormFrontend.js
```diff
@@ -297,6 +296,42 @@ class FormFrontend extends React.Component {
...
+                active={faq_acknowledged}
+                disabled={disable_forms}
+                onClick={() =>
+                  this.setState({
+                    faq_acknowledged: !faq_acknowledged,
+                    completed_template: {
+                      ...completed_template,
+                      [field_key]: true,
+                    },
+                  })
```
Stephen-ONeil, line 327: As mentioned above, this should use `mergeIntoCompletedTemplateState`, like the other fields. Might need to toggle the value between `true` and `null` rather than `true` and `false`, so that the validation works as expected. Double check that.

form_backend/src/template_utils/validate_completed_template.js
```diff
@@ -46,6 +46,8 @@ const verify_values_are_expected_and_match_value_types = (
                 );
               })
               .value();
+          case "boolean":
+            return submitted_value;
```
Stephen-ONeil, line 50: Should check if the value's a boolean, don't want to accept other values based on truthy/falsey-ness.

Stephen-ONeil (changes requested):

csong202:
> Reminded by the code cov report, the changes in validate_completed_template.js need additional coverage from the corresponding unit test!

no need to do this anymore since I'm using the "json" case instead creating of the "boolean" one

client/src/FormFrontend/FormFrontend.yaml
```diff
@@ -6,6 +6,13 @@ form_frontend_privacy_ack:
...
+  en: Please read and acknowledge the FAQ before submitting.
+  fr: S'il vous plaît lisez et reconnaissez les FAQ avant de soumettre le formulaire.
+form_frontend_faq_ack:
+  en: I acknowledge that I read the FAQ
+  fr: Je reconnais que j'ai lu les FAQ
```
sranc046, line 14: Minor comment here, I checked on some other GC sites how they used the FAQ term in french and from what I can tell it is most commonly referred to as la "foire aux questions" without the use of the acronym, might be something to consider? As a french reader, just using the acronym feels a bit awkward and out of place. I would suggest using the full name in french and putting the acronym in parentheses afterwards. @Stephen-ONeil can probably comment further. If we do end up just using the FAQ acronym, it would at least have to be changed to "la FAQ".

sranc046 (changes requested):

client/src/FormFrontend/FormFrontend.yaml
sranc046: There is also the other acronym QSP (questions souvent posées) that is sometimes used

client/src/FormFrontend/FormFrontend.js
```diff
@@ -167,8 +167,13 @@ class FormFrontend extends React.Component {
 
     const user_fields = _.omitBy(
       template,
-      ({ form_type }, key) => key === "meta" || !form_type
+      ({ form_type }, key) => key === "meta" || key === "faq" || !form_type
+      /* TODO remove key === "faq" once faq keys are completed
+      replace with (!_.isUndefined(templates) && !_.includes(templates, template_name)) 
+      Replace ({ form_type }, key) with ({ form_type, templates }, key)
+      Also get template_name from this.state */
```
Stephen-ONeil, line 174: Reminder

Stephen-ONeil (changes requested):
Use of the json type is hacky. How about a special `value_type` called "true" to explicitly cover this sort of use case? Validation should be fairly trivial to write and test for that, it fails if the returned value is anything other than an actual boolean value of true.

Alternatively, could make a type like "non_null" to effectively cover the same use case, but I'm not a fan of how permissive that would be. No need to make it so broad if we don't actually want arbitrary (non-null) content submitted to it.

Otherwise almost there!

client/src/FormFrontend/FormFrontend.yaml
Stephen-ONeil: I'll defer to you! If you could, please check the other instance of FAQ in french text, maybe open a PR with your suggested edits and let Christina know what convention you settle on. My only thought is to not bother with the acronym in parenthesis if it's always spelled out in full for french.

client/src/FormFrontend/FormFrontend.js
```diff
@@ -297,6 +297,39 @@ class FormFrontend extends React.Component {
...
+          return (
+            <>
+              <div style={{ fontWeight: 700 }}>
+                {text_maker("form_frontend_faq_note")}
+                {required_asterisk}
+              </div>
+              <FancyUL>
+                {_.map(field_info.faq_content, ({ q, a }) => {
+                  return (
+                    <div key={_.uniqueId("form-faq-qa-pair_")}>
```
Stephen-ONeil, line 310: `key`s should be stable across renders, see the docs here for [more](https://reactjs.org/docs/lists-and-keys.html#keys) on that.

In cases like this where the components in the returned array will not be changing across renders and will always be in the same order, using the index argument from the `map` for the `key` is totally good enough :+1:

client/src/FormFrontend/FormFrontend.js
```diff
@@ -412,12 +445,23 @@ class FormFrontend extends React.Component {
                             )
                             .value();
 
+                          const keys_staying_completed = _.chain(template)
+                            .pickBy(
+                              (field_info) => !!field_info["stay_completed"]
+                            )
+                            .keys()
+                            .value();
```
Stephen-ONeil, line 453: Interesting, maybe consider for a follow up issue that "stay completed" forms could use local storage to have memory between sessions. Don't try to do that in this PR though, it's going to require some thought (should storage key off a hash of the field's content so it busts as needed? should forms include a semantic versioning field that could, along with being generally useful for logging, be used for the local storage busting?)

client/src/FormFrontend/FormFrontend.js
```diff
@@ -167,8 +167,10 @@ class FormFrontend extends React.Component {
 
     const user_fields = _.omitBy(
       template,
-      ({ form_type }, key) => key === "meta" || !form_type
+      // keeping key === "faq" until faq content is made
+      ({ form_type }, key) => key === "meta" || key === "faq" || !form_type
```
Stephen-ONeil, line 171: How about a `hidden` property in the template (I'd put it at the top, above required), not showing any fields with `hidden: true`? That way the templates control this and only the form backend needs to redeploy to change the fields.

Bonus, it makes "meta" less of a special case field. The client needs to know less about it, it's just another `hidden` field when it comes to field rendering.

Stephen-ONeil (changes requested):
Minor changes, should be the last!

See Stephane's earlier comment too, might as well make the `la FAQ` change on this PR.

csong202:
> Interesting, maybe consider for a follow up issue that "stay completed" forms could use local storage to have memory between sessions. Don't try to do that in this PR though, it's going to require some thought (should storage key off a hash of the field's content so it busts as needed? should forms include a semantic versioning field that could, along with being generally useful for logging, be used for the local storage busting?)

would this also cover privacy policy acknowledgement?

Stephen-ONeil:
> > ...
> 
> would this also cover privacy policy acknowledgement?

Hm, good question. To the extent that acknowledging the privacy policy is a sort of quasi-legal thing, it probably shouldn't be stored between sessions (shared computers, etc) but using session storage for it would be a small UX improvement! It's a frontend only thing anyway, so it doesn't necessarily need to (although could, in part) use the same memory utilities as the other forms.

Tangent, but might be worth writing the API on the forms end as `store_response: "local" | "session"`, so some fields can choose to persist either within or between sessions. Not sure there's a use case for that (but maybe identity fields in the survey should persist between sessions and some other fields only within?). Should be pretty easy enough to implement either way.

PR #1362, opened 2022-01-25
unit tests for CheckBox

csong202:
Updated checkbox unit tests.

Mocking `injected_build_constants` is not optimal since we need separate files to mock different values of same constant. Might convert the constants to functions in later issue to allow different mockings within same unit test file.


client/src/components/CheckBox/CheckBox.unit-test.tsx
```diff
@@ -0,0 +1,78 @@
...
+      ).toBeTruthy();
+    }
+  });
+
+  it("Displays the given label when an onclick is provided", () => {
+    render(solid_checkbox_with_onclick);
+    const checkbox_span = document.querySelector(
+      is_a11y_mode ? "div.checkbox label input" : "div span.link-styled"
+    );
+    expect(checkbox_span && checkbox_span.textContent).toEqual("sample label");
```
Stephen-ONeil, line 51: Best to try and avoid this sort of reliance on detailed implementation details, makes the tests less useful when refactoring. Keep React Testing Library in mind, it's a bit indirect some times but it focus on selecting by things that would be user facing, which is a good approach to keep the tests implementation-detail-agnostic for the most part.

The React Testing Library way would probably be to import `screen` and use `screen.getByText("sample label")`. E.g. `expect(screen.getByText("sample label")).toBeTruthy()` (or maybe something stricter like `.toBeVisible()`, still fresh enough to their approach that I've not built up go-to patterns for this my self).

Stephen-ONeil (changes requested):
Not suggesting changes on every test in this round, start with the `is_a11y_mode` problem and considering what I suggested RE: using React Testing Library and avoiding implementation details. Take a stab at applying any new ideas from that to the other tests too :+1:

client/src/components/CheckBox/CheckBox.unit-test.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  it("Has checkmark svg", () => {
+    const onClick = jest.fn();
+    const { container } = render(
+      <CheckBox {...{ ...testing_default, onClick }} />
+    );
+
+    const icon_svg = container.getElementsByClassName("icon-svg")[0];
+    expect(icon_svg).not.toBeUndefined();
+  });
```
Stephen-ONeil, line 96: This one's testing an implementation detail, can be dropped. Maybe some test to just assert that there's a visual difference between checked and unchecked boxes could be used instead? Testing for _visual_ differences specifically is likely a very hard problem though, so just researching if there's existing solutions to that is enough.

client/src/components/CheckBox/CheckBox.unit-test.tsx
```diff
@@ -0,0 +1,97 @@
...
+    expect(onClick).toHaveBeenCalledTimes(1);
+    fireEvent.keyDown(checkbox, { key: "Enter", code: "Enter", charCode: 13 });
+    expect(onClick).toHaveBeenCalledTimes(2);
+    fireEvent.keyDown(checkbox, { key: " ", code: "(space)", charCode: 32 });
+    expect(onClick).toHaveBeenCalledTimes(3);
+  });
+
+  it("Is checked when checkbox is active", () => {
+    let active = false;
+    const onClick = jest.fn(() => (active = !active));
```
Stephen-ONeil, line 63: Interesting, doing this does make the test work more like a real use case _but_ I think it's actually not the best approach. It makes this test dependent on the `onClick` behaviour funcitoning. That behaviour is already covered by it's own test, and unit tests should avoid overlapping as much as possible. Keeps the feedback from failing tests clearer at a glance.

client/src/components/CheckBox/CheckBox_a11y.unit-test.tsx
```diff
@@ -0,0 +1,69 @@
...
+      fireEvent.click(checkbox);
+      expect(onClick).toHaveBeenCalledTimes(1);
+    } else {
+      throw Error;
+    }
+  });
+
+  it("Is checked when active", () => {
+    let active = false;
+    const onClick = jest.fn(() => (active = !active));
```
Stephen-ONeil, line 39: See comment in other test file

Stephen-ONeil (changes requested):
Comments on the test code itself, I'll keep looking in to the constant mocking problem but I'm leaning towards merging this and opening a follow up issue for working out a better pattern. For now, files per `injected_build_constants` configuration is ok.

If we go that way for now, I suggest some naming convention like `CheckBox--a11y.unit-test.tsx`, and briefly documenting both the convention and the approach for handling `injected_build_constants` in `/client/docs/testing`. You can hold off on that work if you want, just in case one of us works out a better solution quickly.

Stephen-ONeil:
Oh, and you can edit Christina's original post on this PR with updates :+1:

Stephen-ONeil (changes requested):

PR #1367, opened 2022-01-28
Avoid errors on non-ISO 8859-1 characters in search

Stephen-ONeil:
We'd been seeing this occasionally in the logs since the service inventory searching was deployed. I believe it was previously misread as a failure to parse the _response_'s header, but it was actually an error thrown inside the apollo client while parsing the request before it even sent.

E.g. searching this crashes the client "Ḃ, ḃ, Ċ, ċ, Ḋ, ḋ, Ḟ, ḟ, Ġ, ġ, Ṁ, ṁ, Ṗ, ṗ, Ṡ, ṡ, Ṫ, ṫ"

As far as I can tell, everything involved _should_ be able to handle UTF-8 characters, worth looking in to this a bit more (potentially open an issue with apollo-client if I can recreate in a test project, without any of the monkey business we're up to here).

Patching around this by URI encoding the hacky query header on the client side, decoding it on the server side.

### DIFF
--- a/client/src/graphql_utils/graphql_utils.js
+++ b/client/src/graphql_utils/graphql_utils.js
@@ -72,7 +72,7 @@ const query_as_get_with_query_header = async (uri, options) => {
     body: undefined,
     headers: {
       ...options.headers,
-      "gql-query": query,
+      "uri-encoded-gql-query": encodeURIComponent(query),
     },
   };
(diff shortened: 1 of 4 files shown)


Stephen-ONeil:
Self approving for the hotfix.

PR #1368, opened 2022-01-31
Codecov

### LINKED ISSUE #1271: Coverage report merging and CI badges
As of #1269, I've tossed out my old node scripts for test coverage report merging and dropped our old coverage badges which depended on those.

The local dependency pattern used for it was always causing issues with npm audit, and was broken entirely after upgrading from npm 6. On top of that, it was written using the deprecated `istanbul-api` package, so it's likely worth re-writing it largely from scratch (or finding an alternative) when we bring it back.

For reference, [here's](https://github.com/TBS-EACPD/infobase/tree/cc050884b49de710ac49e226fae992e6c477ebac) a commit from before the scripts were dropped. The coverage report merging is in `scripts/common_node_scripts`, and the coverage badge system can be found by looking through the CI config (where the coverage reports are collected and the badges updated) and the READMEs (where the badges are loaded in the markdown).

Stephen-ONeil:
Closes #1271

Nice, codecov looks like it solves our old headaches with merging multiple coverage files together. Lots of other benefits too, but that's the big win of it to me, ha.

A number of features don't work quite right currently, but I think that's just because it has no base branch data to compare to. Should resolve itself post-merging and for future branches.

TODO:
  - [x] checksuming after `codecov` download in CI
  - [x] badges in README.md's
    - [x] reminder: update URLs to reference master before merging
  - [x] look through configuration options, see if there's anything else we want to set right now
    - good enough for now, will need to ease in to setting thresholds etc as testing generally ramps in (exception: form_backend is in a good state and doesn't often need to change, enabling auto to enforce only-increasing coverage)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,12 +72,11 @@ commands:
       - run:
           command: |
             curl -fLso codecov https://codecov.io/bash;
-            # TODO either add shasum to the docker image or do checksums some other way
-            #VERSION=$(grep -o 'VERSION=\"[0-9\.]*\"' codecov | cut -d'"' -f2);
-            #for i in 1 256 512
-            #do
-            #  shasum -a $i -c --ignore-missing <(curl -s "https://raw.githubusercontent.com/codecov/codecov-bash/${VERSION}/SHA${i}SUM")
-            #done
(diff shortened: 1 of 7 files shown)


Stephen-ONeil:
Self approving, discussions of different configurations to be had once we've all used it in practice.

PR #1378, opened 2022-02-07
Reduce parallel type checking in build_all.sh

Stephen-ONeil:
Webpack configuration _always_ ran the `ForkTsCheckerWebpackPlugin` plugin, creating an async process eating ~500 MB of RAM per build. Fully redundant when running multiple builds, and gets costly when running builds in parallel. Added flag to skip the checking inside webpack, use it in `build_all.sh`. Add separate, full, static analysis pre-job to prod deploys. 
 
While nice to do in general, this was specifically because CI builds were back to running out of memory occasionally, post Node 16 upgrade.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,17 +101,11 @@ jobs:
             - repo-wide-lint-cache-{{ .Branch }}
             - repo-wide-lint-cache-master
       - run:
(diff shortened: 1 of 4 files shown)


Stephen-ONeil:
Discussed and thumbs'ed up over slack

PR #1379, opened 2022-02-08
TableClass/tables cleanup

Stephen-ONeil:
Isolating and stripping down the table code, lots of legacy/vestigial junk in there. Considering this a pre-requisite to decoupling the RPB from the old table API. All steps towards dropping the table API altogether. 

Update: don't think it's worth trying to untangle this spaghetti code further. We'll nuke everything that's left once the data is all coming from GraphQL.


sranc046 (approved):
Seems good!

PR #1385, opened 2022-02-09
Drop IE11 support

Stephen-ONeil:
To be merged and released on or after June 15th.

https://death-to-ie11.com/

TODO:
  - [x] new babel target that doesn't include IE11
    - [x] switched to package.json `browserslist` field, so that babel, webpack, and anything else that checks this will be in sync
  - [x] stop intentionally transpiling node modules in prod builds (builds should be much faster, output potentially smaller!)
  - [x] drop misc in-app polyfills
    - [x] pre-spinner polyfill for promises and iterators
    - NOTE: specifically leaving `dynamic_polyfills` for now, these basically don't add weight unless they're actually used, which may include some browsers still technically inside the target. `core_pollyfills.side-effects.js` also should stay, it's part the babel/corejs configuration and we still want to transpile proposal syntax now and then, etc
    - [x] misc polyfills inside other modules, like clipboard, and resize and intersection observers
  - [x] drop IE11 feature detection and related code branches
  - [x] how much faster are prod builds, in CI in particular? Better memory usage too, I hope
    - need more data points, but build may have been about a minute faster and used about 200MB less peak memory per build... I wonder if we can get it low enough to fit four parallel builds in to CI again? 
    - browser test step wasn't significantly faster, too bad, was secretly hoping we'd get some sneaky performance benefits for cutting weight and using less transpiled features
  - [x] what does bundle stats say, do we trim a lot of fat?
    - not massive overall (2.37% decrease), but initial JS is down by 62.41%!
  - [x] We primarily used react virtualized for IE's benefit I believe. Is it worth keeping for mobile devices or are all modern browsers rendering optimized enough for it to not be necessary? It would be nice to drop, adds lots of book keeping and complexity where used, and has been dropped by the original maintainer
    - [x] drop from Typeahead
    - [x] drop from DisplayTable filter dropdown
    - [x] test worst cases of both with modern browsers and low-power devices, compare with and without react virtualized
  - [x] **Important!** Test thoroughly on every non-IE11 browser we can get our hands on
    - [x] copy to clipboard broken everywhere
  - [x] look for hacky style hotfixes or comments mentioning things we can't do yet because of IE11 support requirements. Any related issues
  - [x] hm, client unit test job has started timing out, apparently while trying to collect coverage information. Didn't happen on this branch until after a recent rebase + the commit fixing the clipboard API usage; neither contain a clear culprit for this behaviour. 
    - works fine locally (tested with and without jest caches and previous run coverage output cleared)
    - stills fails in CI when skipping the jest cache
    - solved on #1441, jest need to manually set a sane `maxWorkers` value in CI, as the default detects far more "available" threads than the container actually has access to. Individual workers were ending up totally starved for resources 
  - [x] copy to clipboard broken in edge, at least on work tablets. `window.navigator.clipboard` is undefined, suspect IT has it disabled through some security mechanism
    - not fixed, a "fix" would be an ugly hack at best, but the `WriteToClipboard` component now handles the errors better

Follow up:
  - [ ] our `Details` component could finally use a `details` tag
  - [ ] CSS features we've been having to sleep on, like grid etc.


Stephen-ONeil:
Paused, non-urgent and might ultimately wait for the June deprecation date.

PR #1401, opened 2022-03-04
Review AXE violation allow list

Stephen-ONeil:
At a minimum, some of the false positives might still be worth "fixing", just to make the violation snapshot more manageable. Might also find some legitimate issues, who knows.

TODO:
  - [x] review all current allowed violations. Fix any that are simple, add comments the the snapshot otherwise, and document with next steps here as applicable
  - Report Builder - Table Picker
    - [x] colour contrast issue on focused/hover
    - [x] check the aria-dialog-name warning again
      - fixed some other general `StatelessModal` violations in the process
    - [x] tabindex greater than 0, from react-focus-lock package ~can't fix, _unless_ we want to use patch-package for it~
      - Dropped `FocusLockedModal`, it was redundant to `StatelessModal`, which does it's own focus management
  - [x] nested interactive issue with glossary items inside other interactive elements
    - ah, these were all going to be problematic with #1270 anyway
    - [x]  inside COVID table checkboxes
      - slightly worse design, but I couldn't think of a better alternative to just sticking the relevant glossary entries to the right in parenthesis. Could be in the panel feature for listing "Additional terms", but that buries it past all the content where it'd be relevant
    - [x] nested interactive issue with glossary items inside the results "VisiblityControl" component
      - ugh, don't like this thing, it's a real mash up of different features. Probably shouldn't have allowed it to be a standalone component/spread outside the results panel, ha
  
TODO, but likely as follow up issues:
  - Infographic - Gov - COVID-19 Response
    - focus allowed on aria-hidden glossary icons, to be fixed by PR #1315
  - aha, I think I've finally figured out the weird, flaky, colour contrast problems on a number of infographics. A height clipper over some tall content seems to cause a lot of false positives for content below it that "overlap" with where the overflow: hidden content of the height clipper is technically positioned. Documented in issue #1404
    - no general fix probably? Will have to be careful to sort them out from legitimate colour contrast issues going forward. Some could be "fixed" by using less `HeightClipper` (which is itself a good thing, I think)
      - worst offenders are all caused by very long height clipped tables inside infographics. These should all be paginated anyway. Resolving those will make this a more manageable headache
      - [some of the other cases in the service inventory are fairly short pie charts that are height clipped for no clear reason. Barely reduces their screen real estate (double check, maybe it's done to cover for cases where they could have a very long legend? If those exist, it should just be scrollable instead of putting the whole thing under a height clipper)


sranc046 (approved):

sranc046:
Looks good to me!

PR #1408, opened 2022-03-14
Upgrade client to TS 4.6 (trivial)

Stephen-ONeil:
[TS 4.6](https://devblogs.microsoft.com/typescript/announcing-typescript-4-6/) is out. Not effected by the breaking changes, no action to take on the improvements/new features. Don't review this PR itself, it's trivial, instead I'd like extra eyes on the release blog post to double check that I didn't miss something non-trivial that we could/should do along side this update.

I feel like the [improved indexed access inference](https://devblogs.microsoft.com/typescript/announcing-typescript-4-6/#indexed-access-inference-improvements) is something that we've had to work around in the past, but not remembering concrete examples and probably not worth trying to find and clean them up right now. Nice that it's there though!

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -110,7 +110,6 @@
         "@typescript-eslint/parser": "^5.8",
         "axe-core": "^4.3.1",
         "babel-loader": "^8.1.0",
-        "babel-plugin-istanbul": "^6.1.1",
         "bundle-stats-webpack-plugin": "^3.2.0",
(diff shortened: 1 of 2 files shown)


Stephen-ONeil:
Ignore the CI failure, the branch name just contains a character (the .) that can't be used in the DB name for the dev link.

csong202 (approved):
looks good to me, didn't see anything in the release notes that would break anything I can think of

Issue #1252, opened 2021-09-23
CRSO by program media query


ktw1016:
In `crso_by_prog`, gap year dashline position is hard-coded with media query because there's no way to get width of the bar from Nivo, and dashline is positioned in the middle of a theoretical bar (where width of that bar changes on screen resize)

We should keep in mind especially with Nivo updates that these hard coded media query values are brittle. If Nivo makes internal changes that impact this, this will likely be broken

Stephen-ONeil:
Look in to Nivo's "custom layer" functionality, can overlay an svg with positioning based on the chart's axis, should handle this safely.

PR #1409, opened 2022-03-14
Nivo upgrade, 0.72.0 to 0.79.1

Stephen-ONeil:
Note, core is still at 0.79.0, all other packages up to 0.79.1. That's just how Nivo's versioning :shrug: 

Doing this now because a) it's been a while and b) React 18 is coming up relatively soon (stable release candidate is out). Want to keep Nivo up to date on React 17 before hand so it's not a headache during that upgrade.

There's probably some new Nivo features in here and maybe we could benefit from them, but I'd still prefer not to invest in to Nivo any further than we have till it's semantic versioning and migration docs mature...

TODO:
  - [x] go through the limited patch notes, see if any breaking changes are actually documented
    - only one explicit breaking change, not applicable
    - a lot of the other changes have been minor features and TS migrations, so looks like we shouldn't expect a lot of API changes from this set of upgrades! Phew
  - [x] check site thoroughly for other breaking changes/new behaviour
    - not finding any my self! Will need extra eyes on the dev link though, to be sure
  - [x] see if the problem with @nivo/treemap's labels can be fixed without using patch-package, that's less maintainable than ever now that the Nivo dist modules are minified
    - oof, keeping the patch approach. The only fix via the API is by providing our own `nodeComponent`, which means either ignoring or re-implementing the consumption of a huge chunk of the overall treemap API! That's even worse for maintenance than a hacky little patch!
  - [x] double check that our D3 versions are still compatible. No heads up from npm, but worth confirming
    - Yup, [duplicate D3 packages](https://dev.ea-ad.ca/nivo-upgrade/bundle-stats.html#/packages?bp=%7B%22filters%22%3A%7B%22duplicate%22%3Atrue%7D%7D), **but** it's actually better on this branch than on master. Haven't dug in deeper than the bundle stats. It does **not** appear to be between us and nivo either, most of these seem to be D3 packages duplicating dependencies against each other... (well, worth a closer look, that _could_ have a root cause of us and nivo wanting slightly incompatible versions or something)
    - nope, spotted checked `d3-color` and us and nivo both want a 2.x, it really is our `d3-transition` itself that specifies a 1.x... Not in scope, that's something to work through on a D3 upgrade branch if/when necessary
  - [x] ~minor, but my code review prediction that Taewan backburnered in #1252 has come true, the [hacky dashed line](https://dev.ea-ad.ca/nivo-upgrade/index-eng.html#infographic/crso/AGR-BWN00/financial/.-.-(panel_key.-.-'crso_by_prog_fte))'s slightly out of place, ha~
    - ~may or may not be in scope for this PR, since I'd prefer taking time to do the proper fix rather than just massaging the current hardcoded junk. Depends whether it's relatively trivial do "right"~
    - oops, thought to check master and the line was already off-center, so not this PRs problem at all! Must have happened a while ago, been too minor for anyone to notice/bring up


sranc046 (approved):
Didn't notice any glaring issue, seems good!

PR #1418, opened 2022-03-21
Metadata models

Stephen-ONeil:
Sub-unit of #1416, centralizing metadata models.

Data sources/dataset information is a little scattered right now. Data sources belong to the metadata route but are referred to from `Tables` and panels. Dataset information is mixed; for legacy `Tables` the concept is 1-to-1 and the dataset metadata lives on the table definition. For data not in a table (igoc, program structure, etc) or fetched from GQL, the dataset linkages don't exist very strongly; the info lives directly in the metadata route and the linkage only exists via manual tagging from some panels.

Datasets are an _external_ concept, they're something that lives on open data, etc! Important to remember that, the InfoBase often combines multiple datasets for a given presentation. With GQL data, the relationship is even less innate as the merging of datasets happens on the back end and the client accesses data via the abstraction of the schema (vs the relatively 1-to-1 legacy `Table` API). To support the GQL based future of the InfoBase, a central and standalone model for datasets, with manually managed linkages, is likely necessary.

This PR _will_ decrease cohesion, expect things to feel a bit more manual (e.g. panels having to declare datasets instead of deriving from table dependencies). That's just a flip side of decreasing the coupling between tables and everything else, which is the big driver of this. Some risk that going forward it will be harder to keep the appropriate linkages in place on all panels, etc, but that's going to be a general thing to figure out as data moves to GraphQL, where there's more of a gap between concrete dataset and any given query. Maybe we can eventually make data source/dataset tags part of the query factory interface, or some how even surface it from the GQL side itself, but that'd make the transitional period clumsier. Something for future consideration.

TODO:
  - [x] `models/metadata`
    - [x] `models/metadata/DataSources`, the source information from the metadata route. Datasets, panels, and reports will all declare their own source id(s) as applicable (or maybe panels and reports will reference datasets directly and datasets will own the linkage to a source)
      - [x] all cap source keys are a legacy thing, consider cleaning up
        - [x] added `legacy_key` to `DataSources`, old datasets links will still go to the intended source 
        - [x] use plainer language/more descriptive keys for data sources 
          - [x] make sure the handful of places directly accessing `DataSources.<something>` are also updated! 
        - [x] maybe bother cleaning up old instances of the legacy keys in the code base. There's a bunch scattered around for hardcoded usage with `get_source_link(s)`, mostly in panels
      - [x] some used to pull descriptions from the glossary. Always been problematic
          - [x] Doing away with that, so just live with the duplication for now
          - [x] Open a follow up issue to try and edit the text on the datasets page to be plainer language/more descriptive; leave the technical portion for the glossary? 
            - #1437
    - [x] `models/metadata/DataSets`, limited metadata for individual datasets (a name, a source, and maybe some standalone external links)
      - [x] ... make up my mind whether it's one or two words, maybe should be `Datasets` with a lowercase `s`
        - going with one word, noticed that we consistently say "dataset(s)" on the site itself, so sticking to that
      - ~hm, I forgot datasets could have multiple sources, but understandable. Consider a required primary data source, and optional secondary sources? Does that make sense in practice?~
        - not really. More likely, most or all of these multi-source datasets should be split in to separate datasets some time (e.g. actual and planned program resources don't need to be treated as single datasets)
      - [x] should datasets hold the tag information that currently belongs to the tables?
        - I think so, gives non-table datasets (e.g. covid expenditures) somewhere to centrally declare tags. Panels can get their (default) dataset links _and_ footnote topic tags from the same source
  - [x] remove the duplicate data from tables, tables will probably just need to know their corresponding dataset id
    - for now, eventually the panels and reports will know their corresponding datasets rather than their corresponding tables, and tables won't hold any of this info
  - [x] the metadata route (maybe rename internally to datasets, to match the site) will consume these models rather than the current mix of owning some/accessing some from tables
  - [x] tests
    - [x] `DatasetsRoute/utils.unit-tests.ts`
    - [x] `types/type_utils.unit-tests.ts`
    - ~datasets route would be fairly testable now~
      - Nope, we'll need to finish up the integration test environment with all the appropriate mocks/special loaders for yaml/necessary initialization of systems like `text_maker`, ouch. Tried to work around that with a lot of one off mocks for now but there's too much unrelated cruft to address buried in (usually unused) imports of imports, etc 
    - ~tests on models, hm, it's more configuration than behaviour, but I guess the small bit of mapping on the definitions could be tested? Maybe assert some things about the data (but is there any assumptions not covered by the types?). Test the types? E.g. assert that the key types work out to be a union of strings rather than a generic `string` type or etc~
      - nothing to type on either in isolation, but an integration test would be good to assert that every source has at least one dataset (no code relies on this assumption, but it seems to be a sensible business rule and we can't cover it exactly with the type system). Same problem, too much fiddling and brittleness trying to get all the text maker mess mocked out between the two files
      
Next step followup:
   - [ ] panels should also declare relevant dataset and data source keys directly, rather than sometimes getting them directly and sometimes deriving them from tables. After that, the tables shouldn't have any reason to know about datasets or data sources any more


Stephen-ONeil:
Some handy TypeScript patterns I picked up during this PR:
  - `LiteralKeyedRecordHelper` and the mixed inferred/explicit generic types (hacky, requires a run-time no-op curried function to split inferred and explicit types, but also pretty powerful). Documented thoroughly in `client/src/types/type_utils.ts`
  - writing type assertions as 
```
  type AssertExpectedType<ActualType> =
      ActualType extends SomeExpectedType
        ? true
        : never;

    const type_assertion: AssertExpectedType<TypeToCheck> = true;
```
    - the type system will error on assigning `true` to `never` if the assertion check fails
    - might need to nest the assertions, checking `extends` in both direction if you want exact equality
    - I used this to test the typing behaviour of `LiteralKeyedRecordHelper` in  `client/src/types/type_utils.unit-test.ts`, a bit weird because the jest test I stuck it in will _always_ pass since an issue will only be raised by type checking. For now, I prefer that to having a separate `<module>.type-test.ts` convention

client/src/about/about.yaml
```diff
@@ -104,12 +104,12 @@ our_data_sub_title_1:
...
+    We rely on several governmental reports such as the {{gl_tt "Public Accounts" "PA"}}, the {{gl_tt "Estimates" "EST_PROC"}}, {{gl_tt "Departmental Plans" "DP"}}, and {{gl_tt "Departmental Results Reports" "DRR"}}. A full listing can be found on our [Datasets page](#datasets).
 
     GC InfoBase rolls up available datasets to provide users with a clear picture of government-wide data with a focus on finances and results. Users can easily drill down to get a similar view for individual departments and even for specific programs.
   fr: |
     Nous comptons sur plusieurs rapports gouvernementaux, tels que les {{gl_tt "Comptes publics" "PA"}}, les {{gl_tt "budgets des dépenses" "EST_PROC"}}, les {{gl_tt "plans ministériels" "DP"}} et les {{gl_tt "rapports financiers trimestriels" "DRR"}}. 
-    Vous trouverez la liste complète dans la page [Données](#metadata).
+    Vous trouverez la liste complète dans la page [Données](#datasets).
```
sranc046, line 112: I believe it's "sur la page" rather than "dans la page"

client/src/common_text/faq/people_questions.yaml
```diff
@@ -54,7 +54,7 @@ where_is_data_q:
...
+    Each people management dataset in GC InfoBase has a corresponding dataset available through the Open Data portal. These corresponding Open Data datasets contain quarterly snapshots from March 2008 onward. Open Data links are available from GC InfoBase {{datasets_source_link "Datasets" "RPS"}} page.
   fr: |
-    Chaque ensemble de données sur la gestion des personnes dans InfoBase du GC correspond à un ensemble de données disponibles sur le Portail des données ouvertes. Ces ensembles de données ouvertes inclus également des aperçus trimestriels débutant en mars 2008*. Vous trouverez des liens vers les sites de données ouvertes dans la page « {{metadata_source_link "Données" "RPS"}} » d'InfoBase du GC.
+    Chaque ensemble de données sur la gestion des personnes dans InfoBase du GC correspond à un ensemble de données disponibles sur le Portail des données ouvertes. Ces ensembles de données ouvertes inclus également des aperçus trimestriels débutant en mars 2008*. Vous trouverez des liens vers les sites de données ouvertes dans la page « {{datasets_source_link "Données" "RPS"}} » d'InfoBase du GC.
```
sranc046, line 59: Same thing here for the use of "sur" over "dans"

PR #1423, opened 2022-03-23
Fix Tombstone spilling out of profile panels

Stephen-ONeil:
Very subtle, but see the small grey lines extending out of the panel to the left
image

Reminder (possibly to my past self, ha): bootstrap rows should only appear as children of bootstrap columns! The rows use negative margins, without the column to offset for them you get misalignment like this.


Stephen-ONeil:
Discussed at team meeting so the message is out there, merging since it's trivial.

PR #1425, opened 2022-03-25
use new external package for province chart

AlexCLeduc:
Extracted the chart code to a [re-usable package](https://github.com/TBS-EACPD/canada-province-chart) and replaced the old code with the package

before:
Screen Shot 2022-03-25 at 09 21 12

after:
Screen Shot 2022-03-25 at 09 20 52

A lot of small things may have changed, here's what I can recall of the top of my head:

- The color scale is no longer a single color with an opacity scale, but a monochromatic scale from d3 [interpolateBlues](https://github.com/d3/d3-scale-chromatic#interpolateBlues)
- The province focus behaves almost the same, but now the labels are focused instead of the SVG. This is probably a minor a11y improvement 🤷 
- The shift from long province names to abbreviated province names might be triggered at a slightly different width, but it looks fine to me
- The chart may have extra height sometimes? We can probably play around with this


client/src/charts/canada/canada.js
```diff
@@ -66,7 +69,9 @@ class CanadaGraphBarLegend extends React.Component {
               {`${d.data.year}: ${formatter(d.formattedValue)}`}
             </tspan>
           )}
-          colors={(_d) => get_graph_color(0.5)}
+          colors={(_d) =>
+            color(interpolateBlues(0.5)).copy({ opacity: 0.8 }).toString()
+          }
```
AlexCLeduc, line 74: The province fill-color take the outputted color and apply a 0.8 opacity. It should probably be documented, but if you want to match the color in legends it's also a good idea to apply the same opacity.

client/src/charts/canada/canada.js
```diff
@@ -96,62 +101,86 @@ class CanadaGraphBarLegend extends React.Component {
...
+      return `<div>
+      <div>${formatter(datum.value)}</div>
+      <div>${getName(provCode)}</div>
+    </div>`;
+    },
+    [formatter, containerWidth]
+  );
+
+  return (
+    <div ref={containerRef} style={{ height: "700px" }}>
```
AlexCLeduc, line 173: This height is probably tweakable based on the container width

client/src/charts/canada/canada.js
Stephen-ONeil: Leaving this as a comment on the line's good enough for me :+1:

client/src/charts/canada/canada.js
Stephen-ONeil: Haha, yeah
image

Is an explicit height actually necessary? I think it'll just scale out till it runs out of width, and height will be whatever that needs. I think that's fine for the panel use-case at least. Otherwise `max-height: "700px"` would also be responsive.

client/src/charts/canada/canada.js
```diff
@@ -187,7 +216,9 @@ export class Canada extends React.Component {
             : `${formatter(tick)}+`,
         active: true,
         id: tick,
-        color: get_graph_color(color_scale(tick)),
+        color: color(interpolateBlues(color_scale(tick)))
+          .copy({ opacity: 0.8 })
+          .toString(),
```
Stephen-ONeil, line 221: Same deal as [here](https://github.com/TBS-EACPD/infobase/pull/1425/files#r835270271), maybe define a function for `(scaled_value) => color(interpolateBlues(scaled_value)).copy({ opacity: 0.8 }).toString()` somewhere so we only need to comment on the special `0.8` value once (or not at all at that point)?

Expected values are between 0 and 1, right? Too bad we'd need [this](https://github.com/microsoft/TypeScript/issues/43505) to properly type that.

client/src/charts/canada/canada.js
```diff
@@ -96,62 +101,86 @@ class CanadaGraphBarLegend extends React.Component {
...
+      setTimeout(() => {
         if (!active_prov) {
           prov_select_callback(null);
         }
       }, 200);
       active_prov = false;
-    });
+    }
+    return [onMouseEnter, onMouseLeave];
+  }, [prov_select_callback]);
```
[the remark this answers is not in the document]
AlexCLeduc: I'm just copying what you guys had here, there probably is a cleaner way to do it. I can look into relatedEvent or you can build a wrapper layer that handles this kind of stuff, but the current API is really flexible.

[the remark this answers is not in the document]
Stephen-ONeil: Yeah, doesn't have to be fixed in this PR, didn't know if you left that TODO for us or yourself though. All good to leave it a TODO :+1:

Stephen-ONeil (approved):
:+1: Thanks!

PR #1430, opened 2022-03-29
Transfer payments by region, drop per capita display

Stephen-ONeil:
TODO:
  - ~new data~
    - on hold
  - [x] drop per capita features
    - [x] needs wider discussion within the team, I think
    - [x] remove calculations and text from panel
    - [x] coordinate with pipeline to remove relevant footnotes and population csv up-stream
  - [x] other improvements
    - ~more footnotes (mostly an upstream thing), there's a lot of quirks to the data (e.g. a significant jump in apparent Quebec transfers from ESDC, just because that's where a lot of COVID benefits were processed, when the end recipients were really nation-wide)~
      - on hold
    - [x] panel text is off in cases where the largest "region" is N/A (e.g. ESDC this year, due to a lot of COVID benefits under 100K)
      - Multiple options, decision will likely depend on what can be done with the data upstream
        - [x] alternate text if N/A is the largest, point it out but also re-clarify what N/A means rather than treat it like a regular "region"
        - ~Have a separate "region" code for transfers that were individually <100K, any identified unknown regions >100K use the current N/A code. Add a intro paragraph to the panel explaining quirks including that recipient regions for transfers under 100K aren't tracked. Include a pie chart (or equivalent) showing a comparison of >100K to <100K portions of the total transfer payments. Exclude the amounts <100K altogether from the second half of the panel (otherwise unchanged?)~
          - on hold
    - ~lots of potential special cases to muddle through, e.g. repayments are in the dataset as a type of transfer payment, but are always coded to region N/A, etc~
      - on hold


Stephen-ONeil:
Paused, work to be done on the data upstream, where it's not the highest priority at the moment.

nhamalainen-tbs:
I gave the idea of using the # of recipients for transfers under $100K a bit more thought and realised that it's not a good idea. for one thing, we can't add the number of recipients across different transfers (some will overlap) and highlighting this for users may encourage them to use the information as official recipient counts, which is not the case.

PR #1431, opened 2022-03-29
Fixed search options appearing behind search results

csong202:
Checked ReportBuilder for any breaking changes and didn't see any

### DIFF
--- a/client/src/components/DropdownMenu/DropdownMenu.scss
+++ b/client/src/components/DropdownMenu/DropdownMenu.scss
@@ -32,7 +32,7 @@
   box-shadow: 0px 8px 16px 0px rgba(0, 0, 0, 0.4);
   padding: 12px 16px;
   border-radius: $standardBorderRadius;
-  z-index: 99;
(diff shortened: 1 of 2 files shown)


csong202:
Noticed that on the dev link the filter gets cut off
image

Stephen-ONeil (changes requested):
Our current z-index management is... well, totally unmanaged, haha. I'll open an issue to do some sort of broad overhaul of that, but for this branch can you double check what z-index the pinned content (e.g. FAQ) uses and pick something under that?

image

That's the only floating element I could think of, right now, that can overlap with a display table's controls. Look out though, this `DropdownMenu` component could be used elsewhere. Need to double check other cases, possibly make z-index a prop it accepts, with whatever the best-fit default is, so that different use cases can pick what's appropriate.

Stephen-ONeil:
Oh, right, hahaha, I probably left this review too late, work term's over now right? Thanks again for your time here! One of us will take this one over :+1:

PR #1452, opened 2022-05-13
graphql-codegen

AlexCLeduc:
1. [x] generate `.gql.ts` files based on `.graphql` files
2. [x] Get linting working
3. [x] Update docs for local development and prod deployments
4. [x] Add a couple examples, also showcase suspense and react-query
    - Sorry to batch these, it felt kind of lame to add a non-suspensed example
    -  suspense is worth an extra small library, and when apollo eventually supports it natively you can ditch react-query
6. [x] Figure out how CI is gonna work 🤷 

## CI Difficulty
CI is where I'm not too sure what to do. From now on, a new process will process all `*.graphql` files and generate `.gql.ts` files. I've seen some repos version-control these files, but _usually_ they are git-ignored. 
If we don't version them, CI gets trickier. The process that generates typescript modules uses an introspection query from the server. In CI, the server isn’t yet running, so either we need the server to run to store a schema snapshot (graphql-codegen can also operate "offline", using a schema file) or force CI to generate a schema snapshot before attempting to build or even test client modules. I've seen the schema file work in practice, so I would go that route.
At this point, though, I would recommend just version-controlling the `.gql.ts` files in the repo. If the redundancy get annoying, look into just generating a schema file to be used by graphql-codegen, and add a check in CI to ensure the schema file is sync'd with the server.  

@Stephen-ONeil Maybe you can update the tmux script? I've got my own delicate tmux setup so I don't want to test that it works


server/src/models/schema_root.js
```diff
@@ -10,7 +10,7 @@ export default {
     }
 
     type Query {
-      root(lang:String!): Root
+      root(lang:String!): Root!
```
AlexCLeduc, line 13: Making root non-nullable is going to avoid extra optional chaining once types are generated. 
```ts
// (Before) with a nullable root field 
const services = data?.root?.org?.services
// (After) with a  non-null root
const services = data?.root.org?.services
```

client/src/models/services/queries/new_service_queries.ts
```diff
@@ -0,0 +1,29 @@
...
+  >(ServicesForOrgDocument, { lang, id });
+  return data?.root.org?.services;
+};
+export const useServicesForProgram = (id: string) => {
+  const { data } = useSuspensedQuery<
+    ServicesForProgramQuery,
+    ServicesForProgramQueryVariables
+  >(ServicesForProgramDocument, { lang, id });
+  return data?.root.program?.services;
+};
```
AlexCLeduc, line 29: This is the new recipe for data-fetching. Each query involves importing the query document, as well as the query type and its variables' type. 

There's an extra layer here, because both of these queries were used twice, but you'll often use `suspenseQuery` directly in components if you'd like. I extracted these into their own hooks. 

What's really neat is that you can split up or combine these hooks, making them their own "layer". I haven't combined queries yet, but you'll likely want to create a new hook [that uses `useQueries`](https://react-query.tanstack.com/guides/parallel-queries#dynamic-parallel-queries-with-usequeries), because in suspense mode two `useQuery` hooks side-by-side will not be executed in parallel.

client/src/graphql_utils/useSuspensedQuery.ts
```diff
@@ -0,0 +1,23 @@
...
+  const apolloClient = get_client();
+  const key = query?.loc?.source?.body + JSON.stringify(variables);
+
+  return useQuery(
+    key,
+    async () => {
+      const resp = await apolloClient.query({ query, variables });
+      return resp.data;
+    },
+    { retry: false }
```
AlexCLeduc, line 21: apollo has its own internal retry layer, you don't want that repeated by react-query

Stephen-ONeil:
> At this point, though, I would recommend just version-controlling the `.gql.ts` files in the repo. If the redundancy get annoying, look into just generating a schema file to be used by graphql-codegen, and add a check in CI to ensure the schema file is sync'd with the server.

I'm on board with that. Precedent for the repo. I have that custom loader for importing in to JS/TS from my `.interop.scss` files; it generates/updates type definitions for the Sass files and I check those generated type definitions in.

I think what I'd want is a new parallel job in CI to test that the committed `.gql.ts` files are correct and up to date, probably by regenerating them and asserting no changes (or maybe `graphql-codegen` has a dry-run option for verifying existing files, haven't checked). That'd give me plenty confidence in the committed `.gql.ts` files, on top of the benefits of those files just being committed (e.g. switching branches and having the correct `.gql.ts` immediately, rather than having to wait for a bunch of watch scripts to trip over each-other and regenerate them).

> @Stephen-ONeil Maybe you can update the tmux script? I've got my own delicate tmux setup so I don't want to test that it works

Sure thing, once I'm caught up on the new libraries and have reviewed, I can polish that kind of thing off.

AlexCLeduc:
UPDATE: Replaced react-query with a lighter weight alternative, [suspend-react](https://github.com/pmndrs/suspend-react). react-query + apollo was kind of overkill. 

Also, I realized suspense doesn't need to be a hook, so I stopped using the hook naming convention. `useSuspensedQuery` -> `getSuspensedQuery`

Also, the data will be returned directly from the suspense call rather than nesting it under a `{data}` property. Beforehand I thought it might be useful to add an error property but it looks like when an exception occurs, the consumer component's render never occurs anyway.

Stephen-ONeil:
TODO:
  - [ ] CI job that validates generated .gql.ts files
    - didn't see validation/dry run as a an option with the library (at a quick glance), might need to do something silly like running it and then using `git status | grep "\.gql\.ts"` to look for changes
  - [ ] reconcile with existing query factory pattern
  
I can maybe be convinced to turn these in to follow up issues. Can't manage to be too much of a code tyrant when you're just doing it for fun, haha.

AlexCLeduc:
I went ahead and extracted the more generic getQuery described above and used that within the suspense function.

Stephen-ONeil:
For follow up:
  - [ ] port the rest of the service inventory queries to standalone `.graphql` files and typescript, after the service inventory release (could do it now, probably easy to reconcile, but don't want to get in the way of the impending service inventory release if I don't have to)

PR #1459, opened 2022-06-17
TS 4.7

Stephen-ONeil:
https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/

Mostly improvements you won't have to think about, you'll just benefit from better inference when you encounter it, etc.

Only specific feature that I'd like to highlight is `extends` Constraints on `infer` Type Variables ([here](extends Constraints on infer Type Variables)). Spent a while digging around, but didn't actually find anywhere I could make use of this immediately, `infer` is already uncommon enough, but it's at least a feature we could potentially, explicitly, use in the future.


Stephen-ONeil:
@sranc046 trivial, no need to review this PR, but do see the notes above and at least check out the new feature I highlighted :+1:

PR #1462, opened 2022-06-23
React 18

Stephen-ONeil:
Upgrading to React 18, along with all peer dependency packages caught up in the sweep.

Hold-up dependencies:
  - [x] nivo doesn't have official React 18 support yet, and something in `nivo-tooltip.js` is throwing an invalid hook call error
    - https://github.com/plouc/nivo/issues/1964
      - Fixed in [0.80.0](https://github.com/plouc/nivo/releases/tag/v0.80.0). As always, take care with the nivo updates, no breaking changes indicated but their pre-1.0/bad semantic versioning means no guarantee something won't have broken on us

Likely follow up migrations:
  - [ ] react router v6 upgrade isn't strictly necessary right now, and is a huge pain (even if they have a compatibility features that let you mix some of v5 and v6, that's just not a good idea IMO)

### DIFF
--- a/client/src/glossary/GlossarySidebar/GlossarySidebarDefinition.tsx
+++ b/client/src/glossary/GlossarySidebar/GlossarySidebarDefinition.tsx
@@ -15,27 +15,38 @@ import { IconArrow } from "src/icons/icons";
 
 const { text_maker } = create_text_maker_component(glossary_text);
 
+/*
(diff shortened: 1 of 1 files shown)


Stephen-ONeil:
Paused pending https://github.com/plouc/nivo/issues/1964

Stephen-ONeil:
Still on leave, but chiming in to unpause this. Nivo's fixed it's React 18 constraints.

PR #1468, opened 2022-07-19
Drop transfer payments by region dataset

Stephen-ONeil:
TODO, grab the rational and drop it in this PR for future reference/transparency


Stephen-ONeil:
Didn't forget about the TODO on leaving more context here, will circle back with it next week.

PR #1488, opened 2022-10-17
TS conversion pinned faq

HSenarath:


client/src/components/PinnedFAQ/PinnedFAQ.tsx
```diff
@@ -18,7 +18,20 @@ import { secondaryColor } from "src/style_constants/colors.interop.scss";
...
 
-export class PinnedFAQ extends React.Component {
+const PinnedFAQDefaultProps = {
+  background_color: secondaryColor,
+  is_initially_expanded: false,
+  subject: Gov.instance,
+};
+
+type PinnedFAQProps = typeof PinnedFAQDefaultProps & {
+  q_a_key_pairs: Array<Array<string>>;
```
Steph-Rancourt, line 28: In this case, I would just recommend to use `string[][]` to keep our coding style consistent

client/src/components/KeyConceptList/KeyConceptList.tsx
```diff
@@ -6,7 +6,7 @@ import "./KeyConceptList.scss";
 
 interface KeyConceptListProps {
   question_answer_pairs: [React.ReactNode, React.ReactNode][];
-  compact: boolean;
+  compact?: boolean;
```
Steph-Rancourt, line 9: Hmm, looking at this code I'm not sure of the real use of having `compact` but that can be dealt with later when working on the PinnedFAQ visual bug you mentioned

Steph-Rancourt (changes requested):

client/src/components/PinnedFAQ/PinnedFAQ.tsx
HSenarath: Sounds good, I updated the file and made the change. I'll also keep it in mind for the future!

Steph-Rancourt (approved):

PR #1497, opened 2022-12-02
DRR 2021-22

Steph-Rancourt:


Stephen-ONeil:
**Important:** to anyone using this PR as a reference in the future, there was a couple small configuration changes that were initially overlooked. See #1504 for the fixes and make sure the equivalent are done in future updates :+1: (or maybe we'll replace the hardcoded results doc configuration items with something data driven by then :crossed_fingers:)

PR #1515, opened 2023-02-09
Footnote typescript conversion eg

Egerm010:
NOTE: This branch is unfortunately no longer functional, please refer to https://github.com/TBS-EACPD/infobase/pull/1544

Latest update: Seemingly fixed the issues mentioned below, however the solution I came up with is quite abstract at points (I left comments at some of those parts). Just let me know if anything needs to change.


Egerm010:
Reverted to draft after understanding the circleCI issues my code was producing, will be attempting to fix.

Egerm010:
Fixed more issues such as meta_items typing and imports, but I'm currently stumped on the "footnotes" variables. 

I searched through other files for examples of what types footnote lists have been assigned in the past. Tried FootnoteDef[] (non-array as well) and String[] (non-array as well). Both FootnoteDef[] and String[] caused issues in line 124 that used ._groupBy(subject)- as the property "subject" does not exist on them. Potentially fixed through grouping by subject_type and subject_id (and extracting the subject name from those)- but another 2 issues remained.

Line 177 has a similar issue with calling "subject" on footnotes[0] when it does not exist as a property on FootnoteDef. The other issue is on line 151 which has an issue with the grouped_footnotes variable potentially being a String, which does not comply with the sort_footnote function that calls the variable (and takes FootnoteDef[] as an argument). I tried having the sort_footnote function take strings as input but that did not seem to work either.

Any potential tips would be appreciated.

PR #1518, opened 2023-02-14
Mains 2023 preparation

Egerm010:
Beginning preparation work- still in progress.


Egerm010:
I suspect that my changes in utils.js and common_lang.yaml may be unnecessary. My reasoning for adding them just in case however, was due to the fact that both SEC and MAINS will be updated this time- and so I assigned the current_sups_letter as "C" through hardcoding, while also utilizing mains. As for common_lang- I modified it since Stephen did in a similar fashion for Mains 2022, and made it so that SEC and Mains were set to the current year.

If there's anything I should change or add aside from that, please let me know.

PR #1519, opened 2023-02-16
Pinned content ts conversion

Egerm010:
Feel free to give any feedback or potential recommended changes on this- I tried to leave comments on the changes I felt were the biggest in the code.


client/src/components/PinnedContent/PinnedContent.tsx
```diff
@@ -11,32 +11,54 @@ import { is_a11y_mode } from "src/core/injected_build_constants";
...
+  local_storage_name: string;
+  default_pin_state: boolean;
+}
+
+//Needed for PinnedFAQ to function correctly.
+interface PinnedContentProps {
+  local_storage_name: string;
+}
+
+interface isPinnedMirrorHelper {
```
Steph-Rancourt, line 30: Nitpicking but have the name of the interface match the format of what we use for our other state interface names in other files

Steph-Rancourt (changes requested):

client/src/components/PinnedContent/PinnedContent.tsx
Egerm010: Makes sense, will keep in mind for the future.

client/src/components/PinnedContent/PinnedContent.tsx
```diff
@@ -59,23 +88,26 @@ class NonA11yPinnedContent extends React.Component {
...
-    return get_is_pinned(
-      local_storage_name,
-      default_pin_state,
-      is_pinned_local_storage_mirror
-    );
+    //const { is_pinned_local_storage_mirror } = this.state;
+
+    //Swapped is_local_storage_mirror with 'null' here, as the get_is_planned
+    //const always sets it to null. This is not compatible with it's typing.
+    return get_is_pinned(local_storage_name, default_pin_state, null);
```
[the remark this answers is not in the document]
Egerm010: Potentially fixed, not entirely sure if that last sentence is referring to just the get_is_pinned function call or the get_is_pinned function itself so I changed both, sorry if that isn't right either though- will fix again if not.

client/src/components/PinnedContent/PinnedContent.tsx
```diff
@@ -45,20 +67,32 @@ const get_is_pinned = (
...
+    state: NonA11yPinnedContentState
+  ) {
     super(props);
 
     this.state = {
       is_pinned_local_storage_mirror: get_is_pinned(
         props.local_storage_name,
-        props.default_pin_state
+        props.default_pin_state,
+        state.is_pinned_local_storage_mirror
```
Steph-Rancourt, line 88: In this case, I think it's safe to assume that we can use `null` for the third parameter since it used to default to null in the function (it now requires a boolean or null parameter)

client/src/components/PinnedContent/PinnedContent.tsx
```diff
@@ -45,20 +67,32 @@ const get_is_pinned = (
...
+  NonA11yPinnedContentProps,
+  NonA11yPinnedContentState
+> {
+  placeHolderRef = React.createRef<HTMLDivElement>();
+  contentRef = React.createRef<HTMLDivElement>();
+  static defaultProps: { default_pin_state: boolean };
+
+  constructor(
+    props: NonA11yPinnedContentProps,
+    state: NonA11yPinnedContentState
```
Steph-Rancourt, line 80: Adding state here won't necessary

PR #1520, opened 2023-02-16
Add aria current page to breadcrumb

Egerm010:


Egerm010:
Still working on potentially getting the current_page to read as "GC_Infobase" when on the homepage, not sure if it is needed however.

Egerm010:
Researched other websites utilizing breadcrumbs and it seems as though the homepage link (i.e. in this case the GC Infobase link in the top left), is not of importance if the user is already on the homepage. I did not change the style either as the span did not seem to be effected visually by converting it to a link- and it is not effected upon hovering either (besides the link being clickable but it does nothing as intended, in a similar fashion to clicking the homepage link when you are already on the homepage).

Steph-Rancourt (approved):

PR #1621, opened 2024-06-18
Fix RPB sort function for 'Not Applicable' cases

HSenarath:
Follow up to #1592

To Do:

- [ ]  ~~Crown corps were given 'Not Applicable' value in RPB, but this is breaking the sort function because the columns now have both a string and number to sort. Fix the sort function to be able to handle 'Not Applicable' cases.~~
    - [ ]  Decided to give crown corps null value instead so don’t need to modify sort function
- [x]  Stop null values from being given a value of '0'
    - Currently RPB displays '0' when the csv has an empty value
    - in `TableClass` we gave any `“.”, “”, “-”` a value of zero so it doesn’t break other functionalities such as summing the column values of a csv
        - Changing this to give a value of null instead (This doesn’t cause any issues with the summing function, but need to check if it’ll break anything else on the site)
- [ ]  When populating programSpending remove any pipeline generated planned spending calculations
    - it’s using the old Table API instead of graphQL?
- [ ]  On ‘Authorities and Expenditures’ panel ensure consistency between the items selected in the legend and the ‘Table of Data’ view
- [ ]  Write unit tests

In the future, instead of retroactively removing the pipeline generated planned spending values for crown corps maybe the pipeline shouldn’t generate these values in the first place.

### DIFF
--- a/client/src/components/DisplayTable/DisplayTable.tsx
+++ b/client/src/components/DisplayTable/DisplayTable.tsx
@@ -333,6 +333,7 @@ export class _DisplayTable extends React.Component<
     );
 
     const determine_text_align = (row: DisplayTableData, col: string) => {
+      console.log(row, col);
       const current_col_formatter = col_configs_with_defaults[col].formatter;
       const current_col_plain_formatter =
         col_configs_with_defaults[col].plain_formatter;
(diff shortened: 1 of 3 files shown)


HSenarath:
- [ ] Give planned spending `type: "dollar"`, currently it's set as `"big_int` so it's missing the '$' in the table which is inconsistent with the rest of the spending rows
- [ ] Crown corp planned spending given `-` value instead of `'Not Applicable'`
- [ ] Set text-align to center for any `'-'` values in the rpb
- [ ] Separate ticket in titan repo to address the negative planned spending values found

PR #1648, opened 2024-11-05
Updating Services to use OpenGov Data

HSenarath:
Service inventory opengov data: https://open.canada.ca/data/en/dataset/3ac0d080-6149-499a-8b06-7ce5f00ec56c

List of org variants and programs matched to service id: https://github.com/gcperformance/utilities


HSenarath:
Example of an updated service for 2023-2024: https://dev.ea-ad.ca/service-opengov/index-eng.html#infographic/service/879/intro

PR #1653, opened 2024-11-26
Bar Chart for Gross Expenditure by SOBJ Panel

HSenarath:
Old PRs: 

- https://github.com/TBS-EACPD/infobase/pull/1643
-  https://github.com/TBS-EACPD/infobase/pull/1649

*I created a new PR because it was getting hard to keep track of all the chart changes and the new sobj csv files we were testing out

Background: There were some programs that had some standard objects with negative values, which would cause issues with the gross spending by standard object pie chart. For example, employment insurance: https://www.tbs-sct.canada.ca/ems-sgd/edb-bdd/index-eng.html#infographic/program/HRSD-BGO01/financial.

Final solution is to convert the pie chart to horizontal bar chart that displays each standard object a program has: 
image

### DIFF
--- a/client/src/panels/panel_declarations/finances/sobj/top_spending_areas.js
+++ b/client/src/panels/panel_declarations/finances/sobj/top_spending_areas.js
@@ -2,14 +2,19 @@ import { sum } from "d3-array";
 import _ from "lodash";
 import React from "react";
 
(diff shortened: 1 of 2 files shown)


HSenarath:
@Steph-Rancourt - Could you update the french sections and review the changes to see if I missed anything before I send a dev link? The employment insurance (HRSD-BGO01) is a good example to look at it.

